PR #438, opened 2023-07-17
fix(rules): fix a problem where "all available episodes seen by" not returned correctly when episode 1 of season 1 is not present

appiekap653:
There was a problem where if you didn't have episode 1 of season 1 on your Plex server the expression would never be met.
If you had for example: S03E1-8 and all had been watched, you would suspect that "all available episodes seen by" would return the one that has seen those 8 episodes. but this was not the case.

I have changed it to not look for `season.index === 1 && episode.index === 1` to populate the `allViewers` array.
Instead I populated the array with all users on the server.
from there it does the same thing as before: removing everyone that doesn't exists in the `viewers` array.

With this change it doesn't matter which season you have and how many episode you have.
It will give those that have seen all episodes that are present on the server.

PR #441, opened 2023-07-31
refactor(rules): minor refactor to PR #438

jorenn92:


server/src/modules/rules/getter/plex-getter.service.ts
```diff
@@ -119,17 +119,14 @@ export class PlexGetterService {
                   return null;
                 });
 
-              var i = allViewers.length;
-              while (i--) {
+              allViewers.forEach((el, idx) => {
```
appiekap653, line 122: When I used `forEach` like this and tested it, the result was that it did not worked for seasons if there was only 1 or 2 episodes and about 5 users on the Plex server. Not all users were examined because the `length` of elements and with that the `index` changed with the `allViewers.splice()` function and therefor not all users are examined on the first episode in the for loop. it took 3-4 episodes to have all users examined. The only solution I found was looping the index counting backwards if you want to remove elements while looping.

appiekap653 (commented):
Please see my comments on the forEach refactor. That will break it again when there are less episodes then users on the server.

server/src/modules/rules/getter/plex-getter.service.ts
appiekap653: https://stackoverflow.com/a/9882349

jorenn92: Thanks for this! I wasn't aware of this behavior with forEach & splice.
Did you already try something like this? If I'm not mistaken, it also mitigates the issue.

    const arrLength = allViewers.length - 1;
    allViewers
      .slice()
      .reverse()
      .forEach((el, idx) => {
        if (
          !viewers ||
          !viewers.find((viewEl) => el.plexId === viewEl.accountID)
        ) {
          allViewers.splice(arrLength - idx, 1);
        }
      });

jorenn92: Let's continue this conversation in https://github.com/jorenn92/Maintainerr/pull/449. If you notice an issue with this approach, please let me know.

PR #479, opened 2023-09-18
fix(rules): Fix  a problem where Sonarr's date rules don't work as ex…

jorenn92:
…pected


jorenn92:
This wasn't an issue for Radarr.

Issue #509, opened 2023-10-06
[Feature Request] Add more season/episode fields to the filteroptions


appiekap653:
**Is your feature request related to a problem? Please describe.**
When using the Season or Episode filter-type it will be handy to have some filter fields to make sure Maintainerr doesn't delete the whole season while some episode are still unaired.

When you have for example:
Plex - Users that saw all available episodes contains Mark, and you have set to delete after 5 days.

When there are 3 episodes available for a season of 10 episodes because the rest is still unaired, and you watch those 3 episodes.
Maintainerr will delete the whole after 5 days. If a new episode is released within the 5 days it will not delete, but when the next episode is released within 7 days, it will be deleted by Maintainerr before that date.

**Describe the solution you'd like**
Add some filter fields like:

Sonarr - Total amount of episodes for this season
Plex - Amount of available episodes for this season
Sonarr - Episode-Number of last episode of this season
Plex - Episode-Number of last seen episode

This way you can make sure that when only dealing with whole seasons you can make the rule:
Plex - Amount of available episodes for this season == Sonarr - Total amount of episodes for this season

Or when dealing with partial seasons (example: S01E03 till S01E10 is available on Plex where E10 is the last):
Plex - Episode-Number of last seen episode == Sonarr - Episode-Number of last episode of this season

jorenn92:
Hey, sorry for my very late reply...

I added/clarified a few extra Sonarr rule values that might help with what you're trying to accomplish. 

- Sonarr - Has unaired episodes (Boolean whether a show / season has unaired (monitored) seasons / episodes)
- Sonarr - Season has unaired episodes (On episodes only, boolean whether the parent season has unaired episodes)

You could use these new rules to filter seasons, shows & episodes with unaired episodes out of your collections. I believe this is what you'd like to do?

Unrelated to this issue, but perhaps worth mentioning, these values are also added:
- Sonarr - Number of monitored seasons / episodes (Amount of seasons in the show, or episodes in the season, that are monitored)
- Sonarr - Number of seasons / episodes (also unmonitored) (Same as above, but also counting the unmonitored seasons / episodes)

These changes will be available in the next release. (Sooner on the 'develop' docker tag)

PR #553, opened 2023-10-27
fix(rules): Added Sonarr rules: 'Has unaired episodes', 'Number of mo…

jorenn92:
…nitored seasons / episodes' & 'Season has unaired episodes'


appiekap653:
This does not work as intended 🙁

```
season.statistics.nextAiring !== undefined
``` 

I have a couple of series that I have in Maintainerr as type Season.

Those series still have unaired episodes for the current season but are being added to the collection for removal by Maintainerr.

Rule in Maintainerr:

Sonarr - Has Unaired Episodes - EQUALS - FALSE

Response from my Sonarr API:

```
  "seasons": [
      {
        "seasonNumber": 1,
        "monitored": true,
        "statistics": {
          "episodeFileCount": 0,
          "episodeCount": 0,
          "totalEpisodeCount": 18,
          "sizeOnDisk": 0,
          "releaseGroups": [],
          "percentOfEpisodes": 0.0
        }
      },
      {
        "seasonNumber": 2,
        "monitored": true,
        "statistics": {
          "nextAiring": "2023-11-16T01:00:00Z",
          "previousAiring": "2023-11-02T00:00:00Z",
          "episodeFileCount": 5,
          "episodeCount": 5,
          "totalEpisodeCount": 13,
          "sizeOnDisk": 14454217913,
          "releaseGroups": [
            "FLUX",
            "NTb"
          ],
          "percentOfEpisodes": 100.0
        }
      }
    ],
```

As Sonarr shows I have 5 episodes of the total of 13 episodes for the second season.
The next air date is also specified at: "2023-11-16T01:00:00Z" for episode 6.

Still the series is flagged for removal.
This happens with all my series that has still unaired episodes. So this isn't something like it works with some and works not for some.

I also have some series that have unaired episodes but no air date is specified yet for those episodes.

How are those series handled?

jorenn92:
Hey, it looks like the error is located in this part:

```
if (dataType === EPlexDataType.SEASONS) {
  data.push(...season);
}
```
The spread operator can't be used, because the season object isn't iterable when the dataType is 'season', so this errors out and returns false for everything..

I thought I tested this, but apparently not good enough, thanks for letting me know! I've checked in a fix in [#575](https://github.com/jorenn92/Maintainerr/pull/575)

As for your last question, seasons with unaired episodes, but without a next air date, will return false. I don't think this is something I can avoid, since I don't see another way to find out if a series is still airing. I can't use the episodeFileCount params, since it's possible people didn't download each episode of a season. I'm all open for suggestions though :)

appiekap653:
Maybe we can use the data points I suggested in the first place.
Those data points also fixes the possibility of not downloaded each episode.

#509 

Sonarr - Total amount of episodes for this season
Plex - Amount of available episodes for this season
Sonarr - Episode-Number of last episode of this season
Plex - Episode-Number of last seen episode

jorenn92:
Can't you use these already existing options for this purpose? 
- Sonarr- Number of seasons / episodes (also unavailable)
- Plex - Amount of available episodes

The thing is, I can't keep up with feature requests. For now, I'm using my limited time to prioritize bug reports.

I'd like to look into the rules you requested, they are certainly possible. But they'll have to wait until I have the time for them (as do all feature requests for now).

If you really want them fast, you're free to add them and submit a PR.

PR #662, opened 2023-12-13
feat: Run application as non-root (`node`) user

### LINKED ISSUE #462: [Enhancement] Change Docker user
Currently, the Maintainerr service runs as 'root' inside the container. 
This is not a best practice. Use a dedicated user (or the Node user) inside the application.

bjw-s:
This PR changes the `USER` directive for the Dockerfile to run as `node` instead of the default `root`.
Fixes #462

I am wondering if the default UI port (currently `80`) will become an issue for running rootless, since that is a privileged port number. Perhaps it would be better to change that to `8082` or something, but that could be considered a breaking change.

Aside from pushing to Docker Hub, this PR will also push the image to Github Container Registry, which is a bit friendlier to the Open Source community :)


.github/workflows/build_release.yml
```diff
@@ -0,0 +1,54 @@
...
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ghcr.io/jorenn92/maintainerr
```
onedr0p, line 37:
```suggestion
            ghcr.io/${{ github.repository }}
```

bjw-s: Would that work in this case? Since afaik images need to be lowercased

onedr0p: Ah yeah good point. I'm not sure. Probably best to leave it the way it is then.

jorenn92:
Hey! Foremost, thank you for your contribution! 

I'm not sure either what impact running rootless has on the usage of port 80. These issues will probably be exposed with more testing. I'll build an image using this branch so we can try it.

Regarding the build_release workflow. I've been planning to explore GitHub Actions, but didn't get around to it yet. Is this pipeline functional? As in, will it build and push a docker image? I'd like to fully migrate (and later extend) the current build/release Jenkins pipeline, so this could be a perfect starting point.

jorenn92:
So, I tried the nonroot image. Maintainerr starts as expected (running the UI on port 80 doesn't seem like an issue).

However, since my database file was exposed under a different user, Maintainerr wasn't able to write to it. So a manual permission update was required. 

Ideally, we'd be able to perform this change without requiring manual actions. But I'm not entirely sure what options there are to mitigate this.

onedr0p:
You can use the `user` option in a docker compose file, see Recyclarr docs on this here

https://recyclarr.dev/wiki/installation/docker/#docker-compose-example

bjw-s:
> Regarding the build_release workflow. I've been planning to explore GitHub Actions, but didn't get around to it yet. Is this pipeline functional? As in, will it build and push a docker image? I'd like to fully migrate (and later extend) the current build/release Jenkins pipeline, so this could be a perfect starting point.

This is a fully working pipeline, and you should be able to deploy it as-is. It will build and push an image for each version that you tag (starting with `v`). For any pull requests against the `main` branch it will build the image, but will not actually push it.

> Ideally, we'd be able to perform this change without requiring manual actions. But I'm not entirely sure what options there are to mitigate this.

As @onedr0p already mentioned, that's not really possible I think :( Setting the ownership on the existing file(s) can only be done by the current owner (which is `root`), and would therefore cause the image to still need to run as `root`.

buroa:
Looking forward to this; I removed it out of my deployment until I can restrict it's root privileges.

jorenn92:
> This is a fully working pipeline, and you should be able to deploy it as-is. It will build and push an image for each version that you tag (starting with `v`). For any pull requests against the `main` branch it will build the image, but will not actually push it. I used the workflow that I contributed to another project as a base, except I removed the push to Docker Hub from it: https://github.com/simojenki/bonob/blob/master/.github/workflows/ci.yml.
> 

Awesome, thank again! 

> The only option would be to implement gosu or s6-overlay which are not ideal because they usually require root commands on start to do things like chown/chmod the config dir.
I would rather see docs around this like Recyclarr does. You can use the user option in a docker compose file, see Recyclarr docs on this here

I agree; it's better to do it right and not add extra overhead. I'd like to execute one more 'non-breaking' release first, incorporating a few necessary bug fixes and a small feature.

Once that release is complete, I'll merge this PR, and the next version will either be labeled as 1.7 or 2.0—unsure of what semantic release will choose, but likely 2.0.

In the interim, we can leverage this PR to enhance the documentation, as suggested by @onedr0p

jorenn92:
> > Once that release is complete, I'll merge this PR, and the next version will either be labeled as 1.7 or 2.0—unsure of what semantic release will choose, but likely 2.0.
> 
> This is one of the reasons when I build software I stay below `v1.0.0` that way anything can be a breaking change and you stay semver compliant. I release v1 only after there been a couple years of stability.

I initially intended to follow that approach, but due to a mistake during the implementation of semantic release, it ended up pushing version 1.x, and I decided to just roll with it.

buroa:
@jorenn92 Any update on this?

jorenn92:
> @jorenn92 Any update on this?

It's ready for merging; I just need to allocate some time to enhance the documentation. I plan to include that in this pull request and merge everything together. That's the reason for the delay.

PR #777, opened 2024-01-22
fix: do not create a supervisord.log file

buroa:
This allows us to mount the root filesystem as read-only. 

[See here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/](https://kubesec.io/basics/containers-securitycontext-readonlyrootfilesystem-true/)

### DIFF
--- a/supervisord.conf
+++ b/supervisord.conf
@@ -1,7 +1,7 @@
 [supervisord]
 nodaemon=true
-logfile=/dev/null
-logfile_maxbytes=0
+logfile=/opt/data
+logfile_maxbytes=20971520
 user=node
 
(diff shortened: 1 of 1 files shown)


jorenn92:
Thank you for your contribution!

I suggest redirecting the log file to the /opt/data volume rather than removing it entirely.

jorenn92:
> Thank you for your contribution!
> 
> I suggest redirecting the log file to the /opt/data volume rather than removing it entirely.

I've pushed this change to your branch. Once the checks are completed, I will proceed with merging it.

jorenn92 (approved):

Issue #688, opened 2023-12-29
Exclude movies in collections created in Plex


stixx:
**Describe the bug**
I don't know if this is actually a bug but what I'm trying to accomplish is somewhat the same as the following feature request: #493 

I want to create a rule to exclude all movies that are in collections made in Plex. I could be using the Radarr tag to exclude movies but that is rather tedious with large libraries.

Maybe I'm doing something wrong, if so please help me out :).

**To Reproduce**
Create a rule with the following criteria (see screenshot):
- First value: Present in amount of other collections
- Action: Equals
- Second value: Number
- Custom value: 0

_If this rule is not supposed to work this way than this bug report should be a feature request._

**Expected behavior**
Created collection from rule should not contain movies which are in collections created within Plex.

**Screenshots**
![10 0 1 5_8154_rules](https://github.com/jorenn92/Maintainerr/assets/10729672/a1da6738-7faf-4944-a1ca-a122fe924e34)

**Device (please complete the following information):**
 - Docker: latest image

```
        "Id": "sha256:b02867c7205dcf01990d20fc4c68fc9157d5841f6e98bdb216b0456dd5834465",
        "RepoTags": [
            "jorenn92/maintainerr:latest"
        ],
        "RepoDigests": [
            "jorenn92/maintainerr@sha256:a53573d2366c4908e8cbd5df327a7062a27e767124e68a47714071d6979c94c8"
        ],
```

jorenn92:
Hey! Sorry for the late reply.

There's an issue with the 'Plex - Present in amount of other collections' rule in release 1.7.0, which is why it doesn't work as expected. It's currently fixed in the develop branch and will be available in the next release.

Once the rule is functioning, it should achieve what you desire.

xbglowx:
@jorenn92 will that also fix my problem? https://github.com/jorenn92/Maintainerr/discussions/691

bolmstedt:
This bug seems to still be unresolved in 1.7.1. When I run a rule for movies present in 0 other collections, it adds a movie that matches the rest of the rules to my custom collection, but on the next run it removes that movie from the collection, and then it just keeps cycling like that for consecutive runs, adding it and removing it from the collection.

It does keep a movie that does not match the rules at all which I added manually to the collection in Plex, if that is helpful.

I've had a look at the source code, and it _seems_ to be correct, but it does not seem to work for some reason.

jorenn92:
[...]

PR #804, opened 2024-01-27
fix(rules): Fixed an issue where the 'Plex - present in amount of oth…

jorenn92:
…er collections' rule wouldn't work with custom collections


jorenn92:
Solution for issue in #688

PR #830, opened 2024-02-04
fix(docker): create .cache directory in expected location

wplinge:
Node appears to want to access /opt/app/.cache, but when running as non-default user the /opt/app isn't writable. We attempted to create a more permissive directory during build to accommodate this but the location was slightly wrong.

I can reproduce the issue by running `podman run -u 1001:1001 docker.io/jorenn92/maintainerr:latest`, which gives me a constantly respawning daemon and logs like this:

```
2024-02-04 12:42:18,185 INFO supervisord started with pid 1
2024-02-04 12:42:19,188 INFO spawned: 'docs' with pid 2
2024-02-04 12:42:19,189 INFO spawned: 'server' with pid 3
2024-02-04 12:42:19,191 INFO spawned: 'ui' with pid 4
Internal Error: EACCES: permission denied, mkdir '/opt/app/.cache'
Error: EACCES: permission denied, mkdir '/opt/app/.cache'
2024-02-04 12:42:19,595 WARN exited: docs (exit status 1; not expected)
2024-02-04 12:42:19,596 WARN exited: server (exit status 1; not expected)
```

It looked like a typo to me, but I've not used Node before so may be missing some subtleties on how it works.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -103,6 +103,9 @@ RUN corepack install && \
 
 USER node
 
+# Picked up for Node's .cache directory.
+ENV HOME=/
+
 EXPOSE 6246
 
 VOLUME [ "/opt/data" ]


wplinge:
Well this is annoying, it's a difference in how Docker & Podman behave. Docker really is using `/.cache`.

It seems like Docker sets the environment variable `HOME=/` (even without an entry in `/etc/passwd`), which gets picked up by Node. Podman doesn't so that aspect at least falls back to `$PWD` which is `/opt/app`.

I'm inclined to abandon this PR since I'm really unconvinced every container should have to work round differences like this. I've added a manual `HOME=/` to my config.

mefbgil:
Same here...

 ⨯ Failed to write image to cache 7paYWZ9Ac2VstfuuIEMt5HW8o9nCbISol7IzPhYxgMs= [Error: EACCES: permission denied, mkdir '/opt/app/ui/.next/cache'] {
  errno: -13,
  code: 'EACCES',
  syscall: 'mkdir',
  path: '/opt/app/ui/.next/cache'

wplinge:
That sounds like a good suggestion. I've tested it on `{Docker, Podman} x {default, -u 1002:1002}` and it seems to work in all cases. So I've updated the PR.

jorenn92:
> That sounds like a good suggestion. I've tested it on `{Docker, Podman} x {default, -u 1002:1002}` and it seems to work in all cases. So I've updated the PR.

Great! I'll merge it when all checks have passed.

PR #853, opened 2024-02-08
fix: Automatically prepend https for Overseerr, Sonarr and Radarr when using port 443

Thijn:
This took me way too long to figure out. Hence this PR to fix this for future me (And everyone else).

Because the fields are labeled as Hostname / IP it wasn't clear you could enter https://hostname in that field (Which technically is a URL, not a hostname).
So I put my hostname and port 443 in the settings, but every time I clicked test it failed.

It wasn't untill I refreshed the page it became clear the hostname was prepended with http:// by default.

This PR fixes this by prepending the hostname with https when using port 443, or http if not.

The includes checks could be combined into one (https always contains http, so no need to check twice), and I've added a condition for port 443.

(First PR here, please let me know if anything is missing)

### DIFF
--- a/ui/src/components/Settings/Overseerr/index.tsx
+++ b/ui/src/components/Settings/Overseerr/index.tsx
@@ -79,7 +79,10 @@ const OverseerrSettings = () => {
         ? hostnameRef.current.value
         : hostnameRef.current.value.includes('https://')
           ? hostnameRef.current.value
-          : 'http://' + hostnameRef.current.value
+        : portRef.current.value == '443'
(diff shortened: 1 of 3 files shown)


jorenn92:
Hey, thanks for submitting this PR!

I'll merge it once I've validated that it still works as expected with the latest changes. There were some other bug fixes & refactors with the Plex settings.

jorenn92:
@Thijn, I rebased your branch so it's up to date again. I wasn't able to merge it in it's current state. Once the checks are done, i'll merge it.

jorenn92 (approved):

PR #1117, opened 2024-05-31
Add "Last Aired At" Option for Show

benfugate:
https://features.maintainerr.info/posts/17/latest-episode-air-date-value

Adds a Plex rule value for shows that provides the latest episodes originallyAvailableAt date from the Plex API


ydkmlt84:
@benfugate The docs server inside Maintainerr, the one you updated with the PR, will be going away soon. Please add your doc changes as a PR to this repo. https://github.com/ydkmlt84/Maintainerr_docs

benfugate:
Sure thing, I will get a PR up on that docs repo shortly.

> Also, is this for the latest "aired" episode of the show or the LAST episode that was added?
This would be the original air date for the most recent episode of a show currently available on Plex.
Example: If a 3 season show only has season 1 available on the server, this would provide the original air date of the last episode of season 1.

ydkmlt84 (commented):
So the most recently added to Plex episode's air date. Regardless of how many episodes there are, of the show. I was confused between the latest episode air date and last air date.

benfugate:
Yep, exactly. If you have a suggestion for a better way to word it in the Glossery, of course that's fine with me. Whatever is the clearest/most consistant with the other rules.

ydkmlt84 (commented):
Are you on Discord? I want to ask you a few questions in a DM. Or some other method?

benfugate:
Sure. My discord is @benfugate

ydkmlt84:
> Sure. My discord is @benfugate

I can't find you.

benfugate:
Sure thing, I sent you a message and joined the discord.

ydkmlt84 (changes requested):
Can you move the new parameter to `id:17`, which will of course require the other items to have their IDs updated as well, in the rules.constants.ts file. This will put it below the `last episode added at` rule parameter and keep the similar items grouped together.
 I have some changes for the docs portion too but I will make those when I merge the PR on that repo.

benfugate:
@ydkmlt84 done

ydkmlt84 (approved):
Tested locally. Works good on my end. No issues.

jorenn92:
> Can you move the new parameter to `id:17`, which will of course require the other items to have their IDs updated as well, in the rules.constants.ts file. This will put it below the `last episode added at` rule parameter and keep the similar items grouped together. I have some changes for the docs portion too but I will make those when I merge the PR on that repo.

@benfugate, Firstly, thank you for this contribution! 

@ydkmlt84, Did you test this on an instance that already contained rules before this change? We can only add to the list, not modify it. These IDs are used in rules saved in the database. If this list changes, your old rules will also change (because, for example, ID 17 will now refer to a different rule).

I'm afraid we'll have to revert that portion and add the new rule to the end of the list again.

ydkmlt84:
So...I had a good idea but apparently I got ahead of myself. 

I tested this with the rule at the end. Then had the thought of moving it up higher to be with the other rule, that was similar to it. What I did to test was make a one liner rule, then use test media to make sure it was returning data. I did not test other rules with the new ID.

jorenn92:
> So...I had a good idea but apparently I got ahead of myself.
> 
> I tested this with the rule at the end. Then had the thought of moving it up higher to be with the other rule, that was similar to it. What I did to test was make a one liner rule, then use test media to make sure it was returning data. I did not test other rules with the new ID.

No problem! I'd like to have a better system where this is possible. Introducing an 'order' or 'position' property to properly arrange them in the UI might do the trick. However, we always need a consistent ID for recognizing rules across different versions. Without this, upgrades could break rules at any time. Also, 'import,' 'export,' and 'community rules' won't function correctly.

ydkmlt84:
> > I just had an idea. What if we kept the id as 27, but put the item in the spot for 17? Would it render it that way in the drop down? Or does it go by ID?
> 
> 
> 
> I'm not sure, actually, that might work. But that could become messy after a while as well. I think a new 'order' field would be a better approach.

Agreed, I was just thinking out loud. I tend to do that a lot.

jorenn92 (approved):

PR #1119, opened 2024-06-02
update to docs button and docs URL changes.

ydkmlt84:
I am still learning Git and branches and forks and stuff like that. I know the initial commit on this branch was a while ago, and I am not sure if I made the changes correctly. As far as rebasing and all that goes. I'm trying to use VSC as much as I can, for everything.


jorenn92:
I'm concerned that this branch has diverged too far from the main branch, which would make merging harder. It might be better to rebranch from main and replace the links again?

ydkmlt84:
Yeah probably best. I'll do that.

jorenn92:
> Yeah probably best. I'll do that.

I'm reconsidering this now, perhaps I should just try to rebase the branch.

PR #1213, opened 2024-08-13
added the locally hosted JSON bin instance url and api-key

ydkmlt84:
I believe we may need to start over from scratch with this JSONbin as storage. Unless the old service comes back up. I haven't done a pull of the JSONbin.org rules in a long time so what I currently have is old and outdated.


jorenn92:
I've pushed the latest jsonbin version to jsonbin.maintainerr.info, so it should be up to date now.
Did jsonbin disappear again? I didn't notice it this time.

jorenn92 (approved):

PR #1290, opened 2024-09-28
fix: Tautulli Times viewed & Total views

benscobie:
get_item_watch_time_stats returns all views regardless of the watched percentage which was not the intention.


jorenn92 (approved):

benscobie:
> @benscobie so far this fix is looking better now.. Still doing testing, but I am seeing it not count anything that doesn't meet the Tautulli view % for completion.

Good to hear, thanks for testing 🚀

PR #1291, opened 2024-09-28
fix: Plex cache reset handling if not all apps configured

benscobie:
`constant.applications` was being accessed by index but this array only contains configured applications. There was a chance you'd check the wrong applications props, or get an exception if one of the indexes didn't exist.


jorenn92 (approved):

jorenn92:
Good catch! This probably flew under the radar for a while.

PR #1315, opened 2024-10-09
feat: Add base URL to Tautulli settings

### LINKED ISSUE #1305: Tautullli Integration Fails with HTTP Root Configured
**Describe the bug**

- Tautulli integration does not work when Tautulli has HTTP Root configured. It returns a 404 error.
- It doesn't matter if you set the base URL or not. The error happens when Tautulli has the HTTP Root setting configured. 
- Sonarr and Radarr both support base URL and Tautulli would need the same. 
- Confirmed it works without base URL configured in Tautulli.

**To Reproduce**
1. Go to Settings > Tautulli
2. Configure Tautulli URL, port and API. 
3. Save
4. Test Saved

**Expected behavior**
It's expected that the connection would be successful.

**Screenshots**
```[maintainerr]  |  07/10/2024 12:49:01  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
[maintainerr]  |  07/10/2024 12:49:07  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
[maintainerr]  |  07/10/2024 12:49:16  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
[maintainerr]  |  07/10/2024 12:53:22  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
[maintainerr]  |  07/10/2024 12:53:29  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
```

- Docker
- Maintainerr dc16a13

**Additional context**
- I'm using Docker name resolution and I confirmed that's working fine by running curl from Maintainerr's container shell.

benscobie:
Resolves #1305

I've added a migration to trim any trailing slash of the tautulli_url as the hostname logic is now the same as Sonarr & Radarr.

I've also removed the ref usage in favour of state. Was there a particular reason for using both as this seems to be common throughout the code base? I've only ever seen state being used for form inputs, unless using a library such as React Hook Form which uses refs.

### DIFF
--- a/server/src/database/migrations/1728503476668-Update_tautulli_URL.ts
+++ b/server/src/database/migrations/1728503476668-Update_tautulli_URL.ts
@@ -1,0 +1,13 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UpdateTautulliURL1728503476668 implements MigrationInterface {
+  name = 'UpdateTautulliURL1728503476668';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `UPDATE settings SET tautulli_url = rtrim(tautulli_url, '/')`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {}
+}
--- a/server/src/modules/api/tautulli-api/tautulli-api.service.ts
+++ b/server/src/modules/api/tautulli-api/tautulli-api.service.ts
@@ -102,7 +102,7 @@ export class TautulliApiService {
 
   public async init() {
     this.api = new TautulliApi({
-      url: `${this.settings.tautulli_url}api/v2`,
+      url: `${this.settings.tautulli_url}/api/v2`,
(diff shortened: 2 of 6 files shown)


jorenn92:
> Resolves #1305
> 
> I've added a migration to trim any trailing slash of the tautulli_url as the hostname logic is now the same as Sonarr & Radarr.
> 
> I've also removed the ref usage in favour of state. Was there a particular reason for using both as this seems to be common throughout the code base? I've only ever seen state being used for form inputs, unless using a library such as React Hook Form which uses refs.

I'm not entirely sure anymore. I remember having issues in the past with the state not updating quickly enough, which caused some odd problems. That's why I used refs in those cases. But it's been a while, so I don't recall the exact behavior.

I initially started the project without much experience in NestJS and React (I'm a Java guy), so I'm sure there's some quirks in the code

jorenn92 (approved):

Issue #1302, opened 2024-10-05
Removed Plex Library while rules were attached to that library


ydkmlt84:
I removed a library from inside of Plex that was attached to multiple rules. When I added the library back, with a new Library section ID (but the same media), Maintainerr was not working properly during Rule Execution. The only indication in the log was below:
```
[maintainerr] | 05/10/2024 11:14:20  [WARN] [PlexApiService] Plex api communication failure.. Is the application running?
[maintainerr] | 05/10/2024 11:14:20  [DEBUG] [PlexApiService] Cannot read properties of undefined (reading 'MediaContainer')
```
It was not immediately clear why my rules were giving me that error, until I looked in the Plex logs and found this:
```
Oct 05, 2024 15:09:57.668  DEBUG - Request: GET /library/sections/16/all?includeGuids=1&type=1 (12 live) #159a TLS Page 0-49 GZIP Signed-in Token (ydkmlt84) (129.0.0.0)
Oct 05, 2024 15:09:57.668 WARN - [Req#159a] Invalid library section ID 16 passed.
```
Which led me to think about my "new" library and then realize that I needed to update the rules' library. Then I went to my rules and found that each Rule that had that old library attached, was now showing "undefined".

A better log indication would be more helpful since there wasn't an actual communication issue with the Plex API, which is where I initially looked.

jorenn92:
Good point, Maintainerr won't notice the ID change. I agree a better log message would help a lot in this case.

enoch85:
@ydkmlt84 Is this still valid?

ydkmlt84:
I am not sure about the logs and I would have to recreate the scenario. The thing that happened though, I'm sure is still a thing and I don't see a way around it.

I am not sure when I'll be able to try and recreate but I'll try.

PR #1327, opened 2024-10-19
feat: Support mutiple *arr servers

benscobie:
Adds support for multiple *arr servers as requested here: https://features.maintainerr.info/posts/1/multipe-arr-support. Existing settings are migrated to the new format.

Some other commits alongside this to fix a few things raised recently.

**Improve Plex API failure logging**
- Plex-api throws its own exceptions for API errors so lets log these. Previously all calls presumed success and you'd get a 'Cannot read properties of undefined' error. Slight improvement for #1302.

**Include endpoint and api name in axios error messages**
Only a few calls were logging the endpoint. The logger is also overridden so that the correct context gets logged. Resolves: https://features.maintainerr.info/posts/35/logging-more-details-when-using-debug-true

We now get this:
`[DEBUG] [OverseerrApi] POST request to http://overseerr.home.arpa//api/v1/settings/jobs/availability-sync/run failed: AxiosError: Request failed with status code 404`

Instead of this:
`[DEBUG] [undefined] POST request failed: AxiosError: Request failed with status code 404`

**Improve manual collection and rule execution**
The API had hung when I was testing server failure modes, there were a few places where exceptions could be thrown that wouldn't be caught which might lead to a crash:
- Add missing awaits
- Improve setTimeout usages
- Catch and log any exceptions in the controller
- Only debounce subsequent actions

**Trim trailing slash in Overseerr URL**
Not sure if this is just a Firefox thing, but my Overseerr URL gets a slash appended to the end which breaks API requests to it.


benscobie:
Just had a thought that I should reset the caches when a setting has been update.

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -57,7 +57,8 @@ export class CollectionWorkerService extends TaskBase {
     await super.execute();
 
     // wait 5 seconds to make sure we're not executing together with the rule handler
-    setTimeout(() => {}, 5000);
```
benscobie, line 60: This isn't actually doing anything. If there's no reported issues I can remove my new wait below.

server/src/modules/rules/rules.service.ts
```diff
@@ -299,14 +308,16 @@ export class RulesService {
...
+        // if datatype, manual collection settings or *arr server changed then remove the collection media and specific exclusions. The Plex collection will be removed later by updateCollection()
         if (
           group.dataType !== params.dataType ||
           params.collection.manualCollection !==
             dbCollection.manualCollection ||
           params.collection.manualCollectionName !==
             dbCollection.manualCollectionName ||
-          params.libraryId !== dbCollection.libraryId
+          params.libraryId !== dbCollection.libraryId ||
+          params.radarrSettingsId !== dbCollection.radarrSettingsId ||
```
benscobie, line 319: I've added a warning dialog when switching between arr server to alert the user that this will happen

jorenn92:
Hey, thanks for the great work again! This is a big one, I’ll need some time to process everything.

benscobie:
> I’ll need some time to process everything.

Of course 😄 Feel free to hit me up on Discord if you any questions

jorenn92 (approved):
I’ve finally gone through the code and didn’t notice anything breaking, though it still needs thorough testing. What’s your opinion on merging it into the dev branch? Do you think it’s not quite ready yet, or are you confident we can move forward?

Personally, I believe we could merge it to dev so others can start testing it as well.

server/src/modules/collections/collection-worker.service.ts
benscobie: I understand why it exists, it just wouldn't work like it was as setTimeout doesn't block. It runs the callback function after the timeout asynchronously

benscobie:
> I’ve finally gone through the code and didn’t notice anything breaking, though it still needs thorough testing. What’s your opinion on merging it into the dev branch? Do you think it’s not quite ready yet, or are you confident we can move forward?
> 
> Personally, I believe we could merge it to dev so others can start testing it as well.

I am fairly confident it works as expected 😄 I'll rebase it tonight and give everything a once over again.

PR #1340, opened 2024-10-29
chore: migrate to eslint flat package to support eslint V9

Nowadays:
dependbot recently migrated to eslint V9, however, they changed the format of the eslint file, but more importantly, the name of the config file that gets autoloaded.

This PR migrates to the flat package format + use the new file name for config auto-loading.

- fixed server:lint script
- fixed linting errors across project


benscobie:
Nice one. I think the dependabot auto merge should be updated to not merge major updates without being tested first.

package.json
benscobie, line 1: Would you be able to fix the lint:server script in here too?

`"lint:server": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",`
to 
`"lint:server": "cd server && eslint \"{src,apps,libs,test}/**/*.ts\" --fix",`

Nowadays: Hello @benscobie ;

I will, 

However, there are roughly 46 lint errors in 20ish files, I'll fix that and commit the changes, should be ready this weekend

cheers

Nowadays: Hello again @benscobie I had time ahead of me,

I fixed all linting errors + some import paths that could be shortened

Let me know.

benscobie: Thank you. I had to re-add the file filter to lint:server as it was picking up the dist folder that gets created on run.

benscobie (approved):

PR #1352, opened 2024-11-06
fix: Saving rules when Tautulli not configured

benscobie:


ydkmlt84 (approved):
Approved fix for previously known issue. Tested fix in multi *arr branch.

PR #1358, opened 2024-11-08
docs: update license reference

emmanuel-ferdman:
# PR Summary
Small PR - updates the `server/README.md` to point to the correct license file.


benscobie:
The licence file referenced in that README, though the same licence right now, isn't necessarily the same licence as this project. I would assume the original LICENCE file in server/ was removed, the README can likely just be deleted to be honest.

PR #1360, opened 2024-11-10
chore: Refactor Dockerfile

benscobie:
- Remove yarn requirement in final image
- Commit the yarn release folder as recommended by them
- Simplify copys and permission setting
- Add extra stage for deps
- Remove sharp install as Next.js 15 now includes it by default


benscobie:
Next I want to separate the dependencies for the two projects to reduce the image size. We use Next.js standalone output so that already reduces the dependencies on build. The problem is Nest.js currently requires the whole node_modules folder which includes Next.js dependencies it doesn't need.

benscobie:
I'm going to go ahead and do the above in this branch given that the Dockerfile changes a lot again anyway.

PR #1369, opened 2024-11-12
chore: Use yarn workspaces & add Turborepo

benscobie:
- Use [yarn workspaces](https://yarnpkg.com/features/workspaces) to split the ui and server apps
- Add [Turborepo](https://turbo.build/repo/docs) to simplify commands and pruning of dependencies for release
- Fix eslint in UI project and fix identified issues
- Update prettier configs to new format
- Remove prettier eslint plugin as VSCode is configured to format on save
- Add GitHub workflow to check formatting
- Remove unused dependencies
- Update data folder referencing as the container structure has changed


supervisord.conf
```diff
@@ -4,7 +4,7 @@ logfile=/dev/null
 pidfile=/dev/null
 
 [program:server]
-command=yarn node /opt/app/server/main.js
+command=node /opt/app/server/server/dist/main.js
```
jorenn92, line 7: Could the directory be placed directly at /opt/app/server/dist/main.js instead of in a subdirectory? Is there a specific reason it needs to be nested within a subdir?

supervisord.conf
```diff
@@ -13,7 +13,7 @@ redirect_stderr=true
 
 [program:ui]
 environment=PORT=%(ENV_UI_PORT)s,HOSTNAME=::
-command=yarn node /opt/app/ui/server.js
+command=node /opt/app/ui/ui/server.js
```
jorenn92, line 16: Same as comment ^

supervisord.conf
benscobie: I wish, but I haven't found anything that indicates that it is possible. The directory structure stays intact, [this page](https://turbo.build/repo/docs/reference/prune) has an example of the output folder structure, and here is maintainerrs now:
image

It makes sense why it's like that when you think about how it builds and resolves modules, it's just a bit ugly on the file system.

We've got ui and server at the root and from what I've seen you'd typically have something like this instead:
apps/server
apps/ui
packages/my-library

So it would be `/opt/app/apps/server/dist/main.js` instead. I didn't want to move everything into an apps directory just yet (or at all) in-case of merge conflicts for anything you're working on.

I'm currently working on being able to reference the server package from the ui so that we don't have to duplicate the response interfaces, which was the main driver for this. Though being able to separate the dependencies between the two projects is nice as well.

jorenn92: I’ve been reading up on Turborepo since I wasn’t familiar with it before, and I can see its appeal now. It definitely looks useful. I still don't like the directory structure at all though. Wouldn’t something like this be cleaner?

```
/opt/apps/server  
/opt/apps/ui  
/opt/packages/my-library  
```

Maybe it’s something we could revisit later, once all the features are done but before the next release? Changing it now would indeed mess with my notifications branch. That said, I’ll still have to deal with a tough merge because of all the other changes. I haven’t been able to work on it the past two weeks, so it’s not up to date yet.

benscobie: That might be achievable actually. I'm pruning & building the two apps separately which is why the directory structure is so deep I think. There really isn't a need to do that as we're shipping them together though so I'll have a play around. I think you'd end up with node_modules and a few others sat within /opt/ so we may want another subfolder just to keep it all contained.

jorenn92: Great! Keep me posted when/if you've made some progress.

benscobie: All sorted, the directory structure is now the same. The image size has also gone from 755 MB to 445 MB which is nice.

I had another idea to create an api package (packages/api) to hold all of the DTOs instead of referencing cross project like I did as an example here. I'd also like to locate request validation logic there (not that there is much currently) using Zod, React Hook Form & [this package](https://github.com/BenLorantfy/nestjs-zod). That work might also help with documenting the API better. When @ydkmlt84 and I looked at swagger generation a lot of the auto generation it did wasn't really working with the types we have right now. But I will save that for a separate PR as this one is already big enough!

jorenn92 (approved):

PR #1373, opened 2024-11-13
feat: Add base path support

benscobie:
This feature allows users to serve maintainerr from a subdirectory when reverse proxying. This functionality is surfaced through a new environment variable called `BASE_PATH`.

The solution for this is a bit hacky as Next.js doesn't support base path configuration at runtime. A placeholder is used for the base path (`/__PATH_PREFIX__`) and is added to the next.config.js and an .env file at Docker build time. On startup the placeholder is replaced in the UI code with the `BASE_PATH` environment variable.

I tested this with a simple nginx reverse proxy:

**docker compose**
```
services:
  nginx:
    image: nginx:latest
    container_name: maintainerr-proxy
    ports:
      - 8080:8080
    volumes:
      - ./nginx.conf:/etc/nginx/conf.d/maintainerr.conf

  maintainerr:
    build:
      context: .
    container_name: maintainerr
    user: 1000:1000
    environment:
      - TZ=Europe/Brussels
      - BASE_PATH=/maintainerr
    ports:
      - 6246:6246

```

**nginx.conf**
```
 server {
  listen 0.0.0.0:8080;
  server_name localhost;

  resolver 127.0.0.11;
  
  location ^~ /maintainerr {
    proxy_pass http://localhost:6246;
    proxy_set_header Host $host;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Host $host;
    proxy_set_header X-Forwarded-Proto $scheme;
    proxy_redirect off;
    proxy_http_version 1.1;
    proxy_set_header Upgrade $http_upgrade;
    proxy_set_header Connection $http_connection;
  }
}
```

http://localhost:8080/maintainerr


jorenn92 (approved):

jorenn92:
It's a shame Next.js doesn’t offer a smoother solution for runtime base path support, but this does seem like the best approach for now.

benscobie:
> It's a shame Next.js doesn’t offer a smoother solution for runtime base path support, but this does seem like the best approach for now.

I think as they provide a platform for hosting Next.js apps they try to get as much performance out of the framework as possible, which ends up reducing their costs, but then forget about developers like us! For example the 400 and 500 response pages are statically generated so they can be served from CDNs, having runtime support for the base path would mean they lose that, so you can kinda see why it's the way it is, unfortunately 😆

benscobie:
Ugh I should've tested that PR tag out, I'm getting permission denied on the sh file. I'll raise a new PR.

adrianipopescu:
is this working currently?

ben-propflo:
> is this working currently?

It's currently only available in the development version.

benscobie:
> is this working currently?

It's currently only available in the development release.

PR #1379, opened 2024-11-17
chore: change develop workflow to develop branch

ydkmlt84:
Added a develop branch to the repo so that we only build new "develop" images when a PR is raised on the develop branch. Changed dependabot target branch to the develop branch. 

The idea is to only push something to the main branch when we are making a release. All new PRs should be going to the develop  branch, from a feature branch.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,7 @@ updates:
       interval: weekly
       time: "20:00"
     open-pull-requests-limit: 10
+    target-branch: develop
     ignore:
       - dependency-name: "@heroicons/react"
       - dependency-name: "chalk"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,10 +3,10 @@ name: "Build"
 on:
   push:
     branches:
-      - "main"
+      - "develop"
   pull_request:
     branches:
-      - "main"
+      - "develop"
   workflow_dispatch:
 
 permissions:
--- a/.github/workflows/run_jest_tests.yml
+++ b/.github/workflows/run_jest_tests.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     branches:
       - "main"
+      - "develop"
   workflow_dispatch:
 
 permissions:


jorenn92:
Changes are also required in the semantic release pipeline, i believe merging this will break it

jorenn92:
Do we need to make changes to the release pipeline too, or are you thinking we just merge everything manually right before a release?

ydkmlt84:
> Do we need to make changes to the release pipeline too, or are you thinking we just merge everything manually right before a release?

Yeah, merge before a release. So releases would be a manual action still. That was my thought anyways.

jorenn92:
> > Do we need to make changes to the release pipeline too, or are you thinking we just merge everything manually right before a release?
> 
> Yeah, merge before a release. So releases would be a manual action still. That was my thought anyways.

Releases aren't a manual action right now, aside from initiating the 'release' workflow itself.
Would it make sense to modify the workflow to automatically merge develop into main before starting the release process?

Alternatively, we could automate the release process to trigger right after a merge to main, especially since we'll be locking down the main branch.

Or do you think we'd run into issues?

ydkmlt84:
That's what I meant about it being manual. The way it is now, nothing will create a release without manually running the release action. From what I could tell anyways. 

 I was thinking about adjusting it to merge develop into main, at the beginning of the release action. I think we would just have to be cautious about what is in the develop branch, and that we make sure the develop branch only has "test ready" level items. Basically making sure that the develop branch is 100% ready for a release at that point. 
 So it would be like: feature branch -> develop branch -> testing develop image for a while and fixing reported problems -> push to main and release.

jorenn92:
I think I’d prefer a workflow where merging into the main branch automatically triggers a release. I agree that merges should only occur once we’ve thoroughly tested the features. We could consider using pre-releases (built from the develop branch) during the weeks leading up to an actual release.

jorenn92:
Also, is it okay with you to merge this change after all the other PRs are merged? I think merging it now might cause conflicts with them

ydkmlt84:
> Also, is it okay with you to merge this change after all the other PRs are merged? I think merging it now might cause conflicts with them

I don't mind at all. This was just something that we talked about, and I thought was a good idea to implement. The only thing that I think is important in all of this is to have the "latest" release and the main branch having the same code. Then we always dev on the develop branch and push to main when we are ready (manually or automatically).

I agree we aren't quite ready to go ahead with it yet.

ydkmlt84:
Closing this for now, until we have more discussion and decide how we want to move forward.

PR #1380, opened 2024-11-18
fix: Docker startup

benscobie:


benscobie:
@jorenn92 Sorry! I've tested that image on my setup and it starts up okay now.

ydkmlt84 (approved):
reviewed

PR #1381, opened 2024-11-18
feat: Add notification support

jorenn92:
This PR introduces notification support to Maintainerr. You can now add notification providers via the settings. When creating a rule, you’ll be able to configure which notification provider to use for that rule.

Please note: I’ve only tested this with the Discord and Pushover notification providers. The others have not yet been tested.


jorenn92:
The "About to be Handled" notification type is still a work in progress. I’m still getting the cron job fully set up. Once it’s ready, I’ll update this PR.

jorenn92:
Todo: 
- Remove the 'none' notification type
- Add the cron timer of the 'Notification timer'-task to the settings (maybe)
- Configurable messages (Probably later in a follow up PR)
- Testing and probably fixing untested notification agents

benscobie:
I'll try and get to reviewing this sometime this week.

jorenn92:
TODO: There might be a logic issue in how notifications are being sent. In certain parts of the code, I call a method to send a notification to all agents active for a specific notification type. However, it might be possible that the notification is also being sent out for rules that don't have that notification agent configured.

This is not tested though, it's just an afterthought I just had.

benscobie (commented):
I just had a quick pass through of the code, nice work! I'll try and make some time to test a few of the agents soon as I use a few of them myself.

Most of the comments are around missing awaits. I believe this might have been on purpose, so that you can fire off notifications without slowing down the workers? You also don't really want to explode a worker if the notification code fails for whatever reason as it's not that important. If that's the case I'd suggest using a queue: https://docs.nestjs.com/techniques/queues

handleNotification could simply become `await this.notificationQueue.add()`, and the existing handling code moved to a different function to be called by a consumer? Though the agent is being passed in a few places so might need a little refactor, or you don't queue those? You could also have another queue that takes a constructed message and the agent to send to, that would mean you can retry individual notifications if the service is having an issue.

1. Add item to queue with the required data (type, mediaItems, collectionName, dayAmount)
2. Consumer takes from that queue, runs transformMessageContent and adds to another queue, one queue message per agent.
3. Consumer takes from that queue and calls agent.send().

EDIT:
I've just realized this will require redis which would make it a pain in the ass for current users. We could do something similar with an in-process task runner though. I need to experiment with nestjs error handling, from what I've read they've got a global handler for exceptions but I've seen it crash from unhandledrejections.

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                      </option>
+                    ))}
+                  </select>
+                </div>
+              </div>
+            </div>
+
+            <div className="form-row">
+              <label htmlFor="about-scale" className="text-label">
+                Days for About To Be Handled
```
benscobie, line 211: I was a bit confused at first when I saw this as I hadn't read the types. Perhaps move this after Types and only make it appear when selecting Media About To Be Handled?

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                  <select
+                    id="agent"
+                    name="agent"
+                    onChange={(e) => {
+                      setFormValues({})
+                      setTargetAgent(availableAgents[Number(e.target.value)])
+                    }}
+                    className="rounded-l-only"
+                  >
+                    {availableAgents?.map((agent, index) => (
```
benscobie, line 191: Does the 'none' option need filtering out here, or just be made into an empty string?

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                )
+              })}
+
+              {/* Select types */}
+              <div className="form-row">
+                <label htmlFor="ssl" className="text-label">
+                  Types *
+                </label>
+                <div className="form-input">
+                  {availableTypes.map((n) => (
```
benscobie, line 279: Does the 'none' type need to be filtered out of this?

benscobie: Sorry just saw this one was on your TODO list still

server/src/modules/notifications/agents/discord.ts
```diff
@@ -0,0 +1,172 @@
...
+  embeds: DiscordRichEmbed[];
+  username?: string;
+  avatar_url?: string;
+  tts: boolean;
+  content?: string;
+}
+
+class DiscordAgent implements NotificationAgent {
+  constructor(
+    private readonly settings: NotificationAgentConfig,
```
benscobie, line 87:
```suggestion
    private readonly settings: NotificationAgentDiscord,
```
You can use the specific config types in all of the agents to get type checking on the options.

ui/src/components/Settings/Notifications/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  return (
+    <div className="h-full w-full">
+      <div className="section-settings h-full w-full">
+        <h3 className="heading">Notification Settings</h3>
+        <p className="description">Notification configuration</p>
+      </div>
+      <div className="m-auto mb-5 flex">
+        <div className="ml-2 mr-auto sm:mr-0">
+          <ExecuteButton
```
benscobie, line 74: The first time around I found the test notification thing confusing. I added a notification, smashed Test Notifications a bunch of times and went on a entrypoint spree because nothing was happening. Only when I went back to the modal did I see the 'Test Notification' type and then it clicked in my head.

I think a friendlier version would be having the Test button within the notification modal that just uses the Test Notification type in the background. Looking at the code it might just be a case of passing the non-saved settings up, create a new agent (maybe can extract the factory logic from registerConfiguredAgents) and calling sendNotificationToAgent?

server/src/modules/notifications/agents/gotify.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  public async send(
+    type: NotificationType,
+    payload: NotificationPayload,
+  ): Promise<boolean> {
+    const settings = this.getSettings();
+
+    if (
+      !payload.notifySystem ||
```
benscobie, line 81: What's notifySystem used for? From what I can see it's always set to false so this and a few other agents will always return early.

ui/src/components/Settings/Notifications/index.tsx
benscobie: There's also no feedback after testing currently, and for Gotify (what I've tested so far) the error & debug don't really help much (I'll add a separate comment for that).

Exceptions in the agents send() could bubble up (or throw a new one) with the actual exception message, that could be returned to the user when they hit the /test endpoint so they don't have to dig through logs.

server/src/modules/notifications/agents/gotify.ts
```diff
@@ -0,0 +1,104 @@
...
+    try {
+      const endpoint = `${settings.options.url}/message?token=${settings.options.token}`;
+      const notificationPayload = this.getNotificationPayload(type, payload);
+
+      await axios.post(endpoint, notificationPayload);
+
+      return true;
+    } catch (e) {
+      this.logger.error('Error sending Gotify notification');
+      this.logger.debug(e);
```
benscobie, line 97: The URL could be included here, e.g.:
```
      const url = this.axios.getUri({ ...config, url: endpoint });
      this.logger.debug(`POST request to ${url} failed: ${err}`);
```

ui/src/components/Settings/Notifications/index.tsx
```diff
@@ -0,0 +1,199 @@
...
+                {!config.enabled && (
+                  <div className="bg-amber-600 px-2 py-0.5 rounded text-zinc-200 shadow-md text-xs">
+                    Disabled
+                  </div>
+                )}
+              </div>
+
+              <p className="text-gray-300 space-x-2 mb-4  truncate">
+                <span className="font-semibold">Agent</span>
+                <a href={config.agent} className="hover:underline">
```
benscobie, line 104: I think having a link here was a leftover from where it was copied.

server/src/modules/rules/rules.service.ts
```diff
@@ -383,6 +386,7 @@ export class RulesService {
           params.isActive !== undefined ? params.isActive : true,
           params.dataType !== undefined ? params.dataType : undefined,
           group.id,
+          params.notifications,
```
benscobie, line 389: This also needs adding in setRules when calling createOrUpdateGroup, I can't save new collections at the moment due to the error below, but it also means it won't save any notifications I've selected as well.

The notifications param on createOrUpdateGroup is optional but .map() is called on it later. This is throwing causing groupId to be null in setRules.

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
```diff
@@ -0,0 +1,302 @@
...
+                      (aboutScaleRef.current = +event.target.value)
+                    }
+                  ></input>
+                </div>
+              </div>
+            </div>
+
+            <div>
+              {/* Load fields */}
+              {targetAgent?.options.map((option) => {
```
benscobie, line 231: I think we should have a different input for the Json Payload option. The monaco-editor would be perfect here.

server/src/modules/notifications/agents/agent.ts
```diff
@@ -0,0 +1,24 @@
...
+  NotificationType,
+} from '../notifications-interfaces';
+
+export interface NotificationPayload {
+  event?: string;
+  subject: string;
+  notifySystem: boolean;
+  image?: string;
+  message?: string;
+  extra?: { name: string; value: string }[];
```
benscobie, line 14: I think this is a leftover from overseerr as I can't see it being used.

server/src/modules/notifications/agents/agent.ts
```diff
@@ -0,0 +1,24 @@
...
+import {
+  NotificationAgentConfig,
+  NotificationAgentKey,
+  NotificationType,
+} from '../notifications-interfaces';
+
+export interface NotificationPayload {
+  event?: string;
+  subject: string;
+  notifySystem: boolean;
```
benscobie, line 11: Same with this and the leftover conditions

server/src/modules/notifications/agents/agent.ts
```diff
@@ -0,0 +1,24 @@
+import { Notification } from '../entities/notification.entities';
+import {
+  NotificationAgentConfig,
+  NotificationAgentKey,
+  NotificationType,
+} from '../notifications-interfaces';
+
+export interface NotificationPayload {
```
benscobie, line 8: It would be nice to include some metadata in this payload where relevant, such as the collection name, and/or the media item name. The webhook only has access to the subject and message which isn't super useful.

server/src/modules/notifications/agents/webhook.ts
```diff
@@ -0,0 +1,137 @@
...
+    return finalPayload;
+  }
+
+  private buildPayload(type: NotificationType, payload: NotificationPayload) {
+    const payloadString = Buffer.from(
+      this.getSettings().options.jsonPayload as string,
+      'base64',
+    ).toString('ascii');
+
+    const parsedJSON = JSON.parse(JSON.parse(payloadString));
```
benscobie, line 84: As you're saving the payload as JSON (unlike overseerr I guess), you don't need any of this decoding stuff.

```suggestion
    const payloadString = this.getSettings().options.jsonPayload as string;
    const parsedJSON = JSON.parse(payloadString);
```

server/src/modules/notifications/agents/webhook.ts
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  public async send(
+    type: NotificationType,
+    payload: NotificationPayload,
+  ): Promise<boolean> {
+    const settings = this.getSettings();
+
+    if (
+      !payload.notifySystem ||
```
benscobie, line 106:
```suggestion

```

server/src/modules/notifications/agents/webhook.ts
```diff
@@ -0,0 +1,137 @@
...
+                Authorization: settings.options.authHeader as string,
+              },
+            }
+          : undefined,
+      );
+
+      return true;
+    } catch (e) {
+      this.logger.error('Error sending webhook notification');
+      this.logger.debug(e);
```
benscobie, line 130: Same as comment before around including the URL in the message.

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
jorenn92: I spent too much time considering how to name this and eventually settled on this option, though I wasn't entirely satisfied with it.

You're right that showing this only when a notification provider of that type is selected would be a better approach. But, since the types are configured in the settings, it might still be a bit confusing for some users?

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
jorenn92: I was planning to remove the 'none' option entirely since it's not being used anywhere.

Adding a friendly name mapping sounds like a great idea.

ui/src/components/Settings/Notifications/CreateNotificationModal/index.tsx
jorenn92: Ah yes, good suggestion!

ui/src/components/Settings/Notifications/index.tsx
jorenn92: Good point, I hadn't really considered that. Looking at it now, it does seem odd. I took the easiest route at the time.
Moving the test button to the creation modal seems like the best approach, and we could also remove the 'test' type in that case.

server/src/modules/notifications/agents/gotify.ts
jorenn92: That's a leftover i still need to remove. I copied the notification agents from Overseerr and adapted them to something Nestjs understands. I believe they use it for notifications regarding the Overseerr application itself (like errors).

ui/src/components/Settings/Notifications/index.tsx
jorenn92: Ah, yes, I mindlessly borrowed it from your multi-arr page. I should have caught that.

server/src/modules/notifications/agents/agent.ts
jorenn92: Indeed, also something that still needs cleaning up

benscobie:
@jorenn92 Would you like me to re-review this or are you still making changes?

jorenn92:
> @jorenn92 Would you like me to re-review this or are you still making changes?
Sorry for the late reply and my recent absence. I've been too busy to work on anything over the past few months. I've been keeping this branch updated with the latest changes from 'main,' but I haven't made many other changes yet. I'll keep you posted. I should be more active soon!

server/src/modules/notifications/entities/notification.entities.ts
```diff
@@ -0,0 +1,31 @@
...
+  enabled: boolean;
+
+  @Column('text', { nullable: true })
+  types: string;
+
+  @Column({ type: 'simple-json', nullable: false })
+  options: string;
+
+  @Column({ default: 3, nullable: false })
+  aboutScale: number;
```
benscobie, line 25: I couldn't figure out what this was actually referring to, and looking at the form input I still don't see the relation 😆 Is there a better name here?

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -48,109 +51,143 @@ export class CollectionWorkerService extends TaskBase {
...
+          if (handledMediaForNotification.length > 0) {
+            this.eventEmitter.emit(
+              'agents.notify',
+              NotificationType.MEDIA_HANDLED,
+              handledMediaForNotification,
+              collection.title,
+              undefined,
+              undefined,
+              { type: 'collection', value: collection.id },
+            );
```
benscobie, line 122: This is great! Really nice way of decoupling and also removes the performance impact on needing to wait for these to finish. I do have one suggestion:

Instead of emitting event types specific to the process that is consuming them, I would suggest naming them based on what has been done. In this case something like `collection_handling.complete`. Then create a new class to hold the event information. This means if we want to do something based on collection_handling in the future, we won't have to add another event here or try and work with the existing `agents.notify` event which would be messy. In actual fact, the event added for `NotificationType.MEDIA_ADDED_TO_COLLECTION` would be useful for the frontend to perform live collection updates, that event could be called something like 'collection.media_added'.

My job status PR introduces some structure to events so I will merge that in now (e.g. [the new enum](https://github.com/jorenn92/Maintainerr/pull/1659/files#diff-938a64a1141188a6db9a305fe40e9246322bc49dda1ba7393937497844430a44)). I spent a few hours testing every single pathway and everything worked okay.

This does mean more work in the notification service as you'll need to listen to the different events, and they will each have a different event body to handle, but I think that will also improve the readability.

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -48,109 +51,143 @@ export class CollectionWorkerService extends TaskBase {
...
+          undefined,
+        );
       }
-    } else {
-      this.infoLogger(
-        'Not all applications are reachable.. Skipping collection handling',
+      await this.finish();
+    } catch (e) {
+      this.logger.error('An error occurred where handling collections');
+      this.logger.debug(e);
```
benscobie, line 182: When I added the logs page I [updated](https://github.com/jorenn92/Maintainerr/blob/eef44e281ebe3d5bb776016b89245f2ad321ceca/server/src/modules/logging/logs.module.ts#L46-L82) the logging transports to output any attached stack traces, so you can include them like so:
```suggestion
      this.logger.error('An error occurred where handling collections', e);
```

I would like to refactor all of the areas that pass errors into the non-error log functions, either by changing them to errors because the user should know about them, or in the cases where we do this for API calls, do something a bit different:

Instead of try catch, log, return undefined. I think we should instead:
- Only log in the API functions when the error is not an AxiosError (so our problem, not an API problem) - or don't log at all and include the error in what I describe below.
- Update the API functions to return [result types](https://www.dennisokeeffe.com/blog/2024-07-14-creating-a-result-type-in-typescript) so the caller knows what kind of error it is. These types could include properties like ok, statusCode, error (for non-axios errors?) or if we do not want to leak the HTTP-ness, you could go with: ok, notFound (in the case of getMovie()), error
- Update the callers to handle the result types.
  - One beneficial area is within getters, take `[here](https://github.com/jorenn92/Maintainerr/blob/eef44e281ebe3d5bb776016b89245f2ad321ceca/server/src/modules/rules/getter/radarr-getter.service.ts#L52)` for example. Currently we log a generic warning and then debug out the stack trace in the API call, and then perform another debug log `Couldn't fetch Radarr metadate for media`. The thing is here, we don't really need the two logs from within the API call if the status code is 404. We can just log one message, and this can be very specific about the issue, whether that be by changing the log level, or by having a different message. For example, we could include a link to the docs help page in the areas where we cannot find a record by a TVDB ID (we got a 404), as it's likely to be a Plex metadata issue, any other error can be handled differently. The point is when dealing with APIs, the caller might want to handle a 200, 404 and 500 differently, and the current method of returning undefined for everything doesn't allow that.

We've also got a bit of an issue in the getters here, as we're returning null in cases like media lookup failure (which includes network error and non 2xx!), and inside a rule when say the `movieFile` does not exist, and then [comparing those null values](https://github.com/jorenn92/Maintainerr/blob/eef44e281ebe3d5bb776016b89245f2ad321ceca/server/src/modules/rules/helpers/rule.comparator.service.ts#L212-L213) (though this could just be a bug as that code looks suspect!). I think here, the comparison should not happen, and the media should be skipped over. At first I thought swap these to return undefined, BUT rules like `Plex - Last view date` return null when there is no watch history, so it's valid. This is another area we could use a result type. I would like to add a way to compare null values at some point too: https://features.maintainerr.info/posts/54/add-comparison-with-null-values

I think comparing nulls might also be a contributor to [this issue](https://github.com/jorenn92/Maintainerr/issues/1446), as a temporary error may change how its evaluated. Next time around it might be fine and reverse. I did [recently add](https://github.com/jorenn92/Maintainerr/commit/35c36d9ffd2eab695df6795c0a45aa4b121634fe) retrying of non-Plex API requests (I want to add retries to the Plex API too), as I've seen some users getting random network errors, so this change _may_ have reduced that linked issue occurring.

Sorry this message got really long, brain dump 🤯

Dockerfile
```diff
@@ -39,7 +39,6 @@ COPY --from=builder --chmod=777 --chown=node:node /app/server/node_modules ./ser
 # Copy packages/contracts
 COPY --from=builder --chmod=777 --chown=node:node /app/packages/contracts/dist ./packages/contracts/dist
 COPY --from=builder --chmod=777 --chown=node:node /app/packages/contracts/package.json ./packages/contracts/package.json
-COPY --from=builder --chmod=777 --chown=node:node /app/packages/contracts/node_modules ./packages/contracts/node_modules
```
benscobie, line 42: Ah my bad, the fix for this is sat in this PR here: https://github.com/jorenn92/Maintainerr/pull/1621/

I create a blank folder as it is possible that node_modules returns with package changes:
https://github.com/jorenn92/Maintainerr/pull/1621/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557

jorenn92: Ah! I was racking my brain trying to figure out why it's failing in my build but not in the main build. 

It did just also fail on /app/packages/server/node_modules. Did you also create an empty dir for that one?

server/src/modules/collections/collection-worker.service.ts
jorenn92: In commit [e7a4da66ac753db034fbc9f98209c1be36a4be42](https://github.com/jorenn92/Maintainerr/pull/1381/commits/e7a4da66ac753db034fbc9f98209c1be36a4be42) I've updated the emit actions to use specific events, as you suggested. I also added a dedicated handler for each event in a new service file within the Events module. Each handler then emits a corresponding event for the notification system. This approach allowed me to keep the notification class unchanged, while still decoupling notifications from the core event logic. That way, future changes or additions won't interfere with the notification handling.

I've also integrated everything with the new event modules you introduced.

server/src/modules/notifications/notifications-interfaces.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export enum NotificationType {
+  MEDIA_ADDED_TO_COLLECTION = 2,
+  MEDIA_REMOVED_FROM_COLLECTION = 4,
+  MEDIA_ABOUT_TO_BE_HANDLED = 8,
+  MEDIA_HANDLED = 16,
+  RULE_HANDLING_FAILED = 32,
+  COLLECTION_HANDLING_FAILED = 64,
+  TEST_NOTIFICATION = 128,
+}
```
benscobie, line 103: This can be tackled in a future PR, but this could be moved to shared to save you having to call `/notifications/types`

server/src/modules/notifications/notifications.service.ts
```diff
@@ -0,0 +1,726 @@
...
+      {
+        name: 'gotify',
+        friendlyName: 'Gotify',
+        options: [
+          { field: 'url', type: 'text', required: true, extraInfo: '' },
+          { field: 'token', type: 'text', required: true, extraInfo: '' },
+        ],
+      },
+    ];
+  }
```
benscobie, line 551: This can be tackled in a future PR, but this could be moved to shared to save you having to call `/notifications/agents`

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -17,6 +17,7 @@ export interface IRuleGroup {
   useRules: boolean
   type?: number
   listExclusions?: boolean
+  notifications?: []
```
benscobie, line 20:
```suggestion
  notifications?: AgentConfiguration[]
```

ui/src/components/Settings/Notifications/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+  const [addModalActive, setAddModalActive] = useState(false)
+  const [configurations, setConfigurations] = useState<AgentConfiguration[]>()
+  const [editConfig, setEditConfig] = useState<AgentConfiguration>()
+  const { addToast } = useToasts()
+
+  const basePath = process.env.NEXT_PUBLIC_BASE_PATH ?? ''
+
+  useEffect(() => {
+    document.title = 'Maintainerr - Settings - Notifications'
+    GetApiHandler('/notifications/configurations').then((configs) =>
```
benscobie, line 27:
```suggestion
    GetApiHandler<AgentConfiguration>('/notifications/configurations').then((configs) =>
```

ui/src/components/Settings/Notifications/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+      setConfigurations(configs),
+    )
+  }, [addModalActive])
+
+  const doEdit = (id: number) => {
+    const config = configurations?.find((c) => c.id === id)
+
+    setEditConfig(config)
+    setAddModalActive(!addModalActive)
+  }
```
benscobie, line 43:
```suggestion
  const doEdit = (id: number) => {
    const config = configurations?.find((c) => c.id === id)

    setEditConfig(config)
    setAddModalActive(!addModalActive)
    GetApiHandler('/notifications/configurations').then((configs) =>
      setConfigurations(configs),
    )
  }
```

Ref: https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

packages/contracts/src/events/maintainerrEvent.ts
```diff
@@ -2,7 +2,13 @@ export enum MaintainerrEvent {
   RuleHandler_Started = 'rule_handler.started',
   RuleHandler_Progressed = 'rule_handler.progressed',
   RuleHandler_Finished = 'rule_handler.finished',
+  RuleHandler_Failed = 'rule_handler.failed',
```
benscobie, line 5: In `\src\modules\events\events.controller.ts`

```
  @OnEvent('rule_handler.*')
  @OnEvent('collection_handler.*')
```

needs to become

```
  @OnEvent('rule_handler.started')
  @OnEvent('rule_handler.progressed')
  @OnEvent('rule_handler.finished')
  @OnEvent('collection_handler.started')
  @OnEvent('collection_handler.progressed')
  @OnEvent('collection_handler.finished')
```

as the new events are getting picked up and an error thrown

server/src/modules/rules/rules.service.ts
```diff
@@ -170,6 +170,7 @@ export class RulesService {
     try {
       return await this.ruleGroupRepository.findOne({
         where: { id: ruleGroupId },
+        relations: ['notifications'],
```
benscobie, line 173: Is this needed as `eager: true` is set?

server/src/modules/rules/dtos/rules.dto.ts
```diff
@@ -19,6 +20,7 @@ export class RulesDto {
   manualCollectionName?: string;
   dataType: EPlexDataType;
   tautulliWatchedPercentOverride?: number;
+  notifications?: Notification[];
```
benscobie, line 23: Only the ID is used, passing the entire configuration object seems wrong

```suggestion
  notificationIds?: number[];
```

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -81,6 +82,8 @@ export class CollectionWorkerService extends TaskBase {
         MaintainerrEvent.CollectionHandler_Finished,
         new CollectionHandlerFinishedEventDto('Finished collection handling'),
       );
+
+      this.eventEmitter.emit(MaintainerrEvent.CollectionHandler_Failed);
```
benscobie, line 86: This is in the wrong place now, merge issue? The try/catch is missing too

server/src/modules/events/events.service.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  CollectionMediaAddedDto,
+  CollectionMediaHandledDto,
+  CollectionMediaRemovedDto,
+  RuleHandlerFailedDto,
+} from './events.dto';
+import { NotificationType } from '../notifications/notifications-interfaces';
+
+@Injectable()
+export class EventsService {
```
benscobie, line 13: I would move all of this code to `notifications.service.ts` given you're listening to the events to serve that purpose.

server/src/modules/settings/settings.service.ts
```diff
@@ -64,6 +65,8 @@ export class SettingsService implements SettingDto {
 
   jellyseerr_api_key: string;
 
+  notification_settings: NotificationSettings;
```
benscobie, line 68: This and the referenced interface aren't used

ui/src/components/Common/ExecuteButton/index.tsx
```diff
@@ -1,13 +1,31 @@
...
+      },
+      props.timeout ? props.timeout : 10000,
+    )
+  }, [clicked])
+
+  const onClick = () => {
+    setClicked(true)
+    props.onClick()
+  }
+
```
benscobie, line 28: This shouldn't be here, probably from resolving merge conflicts.

ui/src/components/Common/ExecuteButton/index.tsx
```diff
@@ -1,13 +1,31 @@
 import { PlayIcon } from '@heroicons/react/solid'
+import { useEffect, useState } from 'react'
 import { SmallLoadingSpinner } from '../LoadingSpinner'
 
 interface IExecuteButton {
   text: string
   onClick: () => void
+  timeout?: number
```
benscobie, line 8: This shouldn't be here, probably from resolving merge conflicts.

ui/src/components/Settings/Notifications/index.tsx
```diff
@@ -0,0 +1,192 @@
...
+
+  function confirmedDelete(id: any) {
+    DeleteApiHandler(`/notifications/configuration/${id}`).then(() => {
+      setConfigurations(configurations?.filter((c) => c.id !== id))
+    })
+  }
+
+  return (
+    <div className="h-full w-full">
+      <div className="section-settings h-full w-full">
```
benscobie, line 53: Just use the Tailwind classes directly

```suggestion
      <div className="mb-5 mt-6 text-white h-full w-full">
```

ui/styles/globals.css
```diff
@@ -337,6 +337,9 @@
     @apply mb-10 mt-6 text-white;
   }
 
+  .section-settings {
+    @apply mb-5 mt-6 text-white;
+  }
```
benscobie, line 342:
```suggestion
```

ui/tailwind.config.js
```diff
@@ -16,6 +16,9 @@ module.exports = {
       screens: {
         xs: '440px',
       },
+      colors: {
+        'bg-zinc-910': '#18181b',
+      },
```
benscobie, line 21: I don't see this being used
```suggestion
```

server/src/modules/rules/tasks/rule-executor.service.ts
```diff
@@ -365,9 +392,25 @@ export class RuleExecutorService extends TaskBase {
...
+            value: rulegroup.id,
+          }),
+        );
       }
     } catch (err) {
       this.logger.warn(`Execption occurred whild handling rule: `, err);
+
+      this.eventEmitter.emit(
+        MaintainerrEvent.RuleHandler_Failed,
+        new RuleHandlerFailedDto(rulegroup.collection?.title, {
```
benscobie, line 409:
```suggestion
        new RuleHandlerFailedDto(rulegroup.collection.title, {
```

benscobie (changes requested):
Being able to test right away is 👌 I re-tested Gotify and that worked.

The refactoring/type suggestions I've made can be resolved in a future PR so we can get this out of the door 🚢

ui/src/components/Common/ExecuteButton/index.tsx
jorenn92: Ah yes, that would explain. The execute button is broken since my latest merge. I'll replace this with the version from main.

server/src/modules/events/events.service.ts
jorenn92: I understood the idea behind separating the notification event emitters into something more general was to make future additions, potentially unrelated to notifications, easier to implement. It seemed more logical to me to place them in a more generic event class for that reason.

I agree it looks a bit odd right now, since we're only using it for notifications. But I think future additions would feel out of place if they were kept in the notifications-specific class.

server/src/modules/settings/settings.service.ts
jorenn92: Good catch, that looks like a leftover from an earlier revision.

server/src/modules/rules/rules.service.ts
jorenn92: It shouldn't, but i could not get it working without this set.

server/src/modules/notifications/notifications.service.ts
jorenn92: Agreed, this dates from the first version of this PR, before contracts were implemented. 
As suggested, I'd rather tackle this in a new PR though

server/src/modules/events/events.service.ts
jorenn92: Ah, I see. I had assumed there could only be one @OnEvent decorator per event. I'm not sure why I thought that, especially since I never actually checked the documentation. So that basically invalidates everything I said earlier.

Consider the events.service.ts class gone!

jorenn92:
> Being able to test right away is 👌 I re-tested Gotify and that worked.
> 
> The refactoring/type suggestions I've made can be resolved in a future PR so we can get this out of the door 🚢

Thanks for taking your time for the thorough review! 
I've resolved most of the issues you pointed out.

I'd like to create a follow-up PR to:

- Refactor the NotificationType enum and notification agents into proper contracts.
- Move notification messages to the database and UI, so they're user-configurable.
-  (Likely) fix any broken notification agents.

server/src/modules/notifications/entities/notification.entities.ts
jorenn92: Hm yes, the name does seem confusing. 
I'd rather not change it in this PR though. I'll add that to the list for the follow up PR.

The value refers to this UI value: 
image

server/src/modules/collections/collection-worker.service.ts
jorenn92: I've also been unhappy with how errors are handled. It's inconsistent and leads to even more issues. Some rules return null, others undefined,.. it's all a bit messy.

We really need a unified approach to error handling across the entire application, along with a clear way to surface those errors to users.

Your proposed approach seems like a solid step toward bringing that structure, so you’ve got my full support.

That said, implementing it is definitely going to be a bit of a pain.

server/src/modules/rules/dtos/rules.dto.ts
benscobie: Sure thing. I've removed RulesDto casting as part of [this draft PR](https://github.com/jorenn92/Maintainerr/pull/1698) which may make this refactor easier.

server/src/modules/notifications/notifications.service.ts
```diff
@@ -0,0 +1,786 @@
...
+            enabled: payload.enabled,
+            aboutScale: payload.aboutScale,
+
+            types: JSON.stringify(payload.types),
+            options: JSON.stringify(payload.options),
+          })
+          .execute();
+      }
+      // reset & reload notification agents
+      this.registerConfiguredAgents(true);
```
benscobie, line 142: Just spotted a bug, when updating notification settings the agent isn't updated with these settings due to the length check in `registerConfiguredAgents`.

jorenn92: Good catch! Fixed it by using the 'isEqual' method from Lodash to compare the 2 Notification[]'s

benscobie (approved):
Great work 🚀

PR #1389, opened 2024-11-20
fix: Handling collections & Sonarr media existence check

benscobie:
- Ensure that .sh files have LF line endings. Executing a .sh file with CRLF line breaks will not work.
- I broke handling of media in my multi *arr work as I was attempting to check the related radarr & sonarr setting IDs via the child collection. This would never work as you can only have one or none. I'm now checking the setting ID on the collection itself.
- We were logging "Unmonitored x from Sonarr show with ID x" even when the call failed
- Check that the returned Sonarr media has an ID before performing any actions on it. getSeriesByTvdbId() always returns a result for a valid TVDB ID, it's just that the fallback API call we make doesn't return an ID.
- Check that the unmonitoring seasons call worked before attempting to unmonitor the series. An exception could be thrown here which would break the collection worker job and leave it stuck running.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf
--- a/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
+++ b/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
@@ -370,6 +370,13 @@ export class SonarrApi extends ServarrApi<{
           }
         }
       }
+
+      this.logger.log(
+        `Unmonitored ${
+          typeof type === 'number' ? `season ${type}` : 'seasons'
+        } from Sonarr show with ID ${seriesId}`,
+      );
+
       return data;
     } catch (e) {
       this.logger.log("Couldn't unmonitor/delete. Does it exist in sonarr?", {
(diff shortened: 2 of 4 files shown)


ydkmlt84:
> Released to `jorenn92/maintainerr:pr-1389` 🚀

Testing shows that it works as expected now.

I had two Radarr instances setup.

```
[maintainerr] | 20/11/2024 14:44:35  [INFO] [CollectionWorkerService] Handling collection 'Test'
[maintainerr] | 20/11/2024 14:44:35  [INFO] [CollectionsService] Removing media with id 103464 from collection..
[maintainerr] | 20/11/2024 14:44:36  [INFO] [PlexApiService] Removed collection from Plex
[maintainerr] | 20/11/2024 14:44:36  [INFO] [CollectionWorkerService] Removed movie from filesystem & Radarr
[maintainerr] | 20/11/2024 14:44:36  [INFO] [CollectionWorkerService] Handling collection 'Test' finished
```

ydkmlt84:
> * Check that the unmonitoring seasons call worked before attempting to unmonitor the series. An exception could be thrown here which would break the collection worker job and leave it stuck running.

Are you saying that this change could cause the collection worker job to break, or that without this it could break?

benscobie:
> > * Check that the unmonitoring seasons call worked before attempting to unmonitor the series. An exception could be thrown here which would break the collection worker job and leave it stuck running.
> 
> Are you saying that this change could cause the collection worker job to break, or that without this it could break?

The latter

ydkmlt84 (approved):

Issue #1390, opened 2024-11-20
Tautulli connection fails with leading slash in base url


jgillman:
**Describe the bug**
Tautulli connection fails when using a leading slash in base URL.

**To Reproduce**
*(self explanatory, see screenshots)*

**Expected behavior**
I would expect the connection to succeed with or without the leading slash in the base URL.

I'm not sure what the "standard" is but other similar services either automatically add/remove it or display an error. Overseerr, for example, says "URL base must have a leading slash" if it's not included to avoid confusion.

**Screenshots**

Fail with slash
![Fail with slash](https://github.com/user-attachments/assets/d448948a-9f6e-4e15-8566-e06bd3fe4657)

Success without slash
![Success without slash](https://github.com/user-attachments/assets/d7b5d594-d426-4d58-a9f1-b53ac68aeb54)

Nginx logs
![Nginx logs](https://github.com/user-attachments/assets/8cc0f130-71aa-4dd9-95e7-6cd0fd659bec)

**Device:**
 - OS: Slackware 15.0 x86_64 (docker container)
 - Version 2.2.1

**Additional context**
Possibly related to https://github.com/jorenn92/Maintainerr/issues/1305 ?

My screenshots show that I'm connecting with docker networking hostnames but I've confirmed that the behavior is the same if I use a local IP address or external domain.

benscobie:
You're right it could be documented better. I would prefer a leading slash as well. At the very least Tautulli is consistent with the Radarr & Sonarr base URL settings. For now we could just add some help text alongside the input as we've got other higher priority things to work on. A PR is welcome.

jgillman:
Thanks for the quick reply, @benscobie! Agreed that just adding some help text would be sufficient here. I'll see if I can get a PR together.

ydkmlt84:
This is an easy documentation change for me. I will add it to the docs. Maintainerr UI will need a PR though. :)

mkaltner:
Just to add on, I'm having issues getting Maintainerr to talk to my Tautulli that's using HTTPS with self signed certificates.  Nothing in my docker logs, just "Connection failed!" in the UI and a request to /api/settings/test/tautulli with an NOK status.

benscobie:
> Just to add on, I'm having issues getting Maintainerr to talk to my Tautulli that's using HTTPS with self signed certificates. Nothing in my docker logs, just "Connection failed!" in the UI and a request to /api/settings/test/tautulli with an NOK status.

[...]

PR #1391, opened 2024-11-20
Add label tip for Base URL settings fields

jgillman:
Adds label tips for Base URL fields! Simple fix for https://github.com/jorenn92/Maintainerr/issues/1390

**Screenshot**
CleanShot 2024-11-20 at 11 23 42


jgillman:
lol 🤦 I force-pushed because the label text I added was the opposite of what it should have been

ydkmlt84 (changes requested):
It is still showing as must have, instead of must NOT have. 

Also, I believe it would be better if it simply said `No Leading Slash`. Since it is implied its information for the Base URL setting, due to the position.

jgillman:
@ydkmlt84 Made the requested changes!

Not sure what happened with my first force-push ¯\\\_(ツ)\_/¯

jorenn92:
Wouldn't it be better for the server-side logic to check for a leading slash and remove it if present?

jgillman:
@jorenn92 Definitely would be better for the server to do it! There was some discussion in the [issue I created](https://github.com/jorenn92/Maintainerr/issues/1390) and adding a label seemed like a decent interim solution that I could readily handle myself.

Not sure how much more involved it would be to properly handle on the back end.

jorenn92:
Oh, I see! I missed that issue. I didn’t realize this was just a temporary solution. In that case, it works for me

ydkmlt84 (approved):
Looks good to me.

PR #1394, opened 2024-11-24
fix: __PATH_PREFIX__ not replaced when using user directive

benscobie:
Files were owned by node:node with 755. If you override the user then they won't have permission to replace `__PATH_PREFIX__` on startup. This change makes the UI folder writable by anybody. At first I didn't like this, but as it was 755 before then the node user could already write files. Really this is no worse as we're in a container.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,6 +115,7 @@ RUN corepack install && \
     # This is required for docker user directive to work
     chmod 777 /opt && \
     chmod 777 /opt/app/start.sh && \
+    chmod 777 -R /opt/app/ui && \
     mkdir -m 777 /.cache  && \
     mkdir -pm 777 /opt/app/ui/.next/cache && \
     chown -R node:node /opt/app/ui/.next/cache


jorenn92 (approved):
I’m not a big fan of it either, but you’re right. The node user already had access anyway.

zhazell:
I do not have a BASE_PATH specified and when loading Maintainerr, the home screen shows 404 error. When clicking Overview from the side, it loads, but the URL path includes https://domain-here.com/__PATH_PROXY__/overview. So does each of the pages, it seems that when BASE_PATH is blank, it adds this to the URL.

I have my user setup to match the file system permissions.

benscobie:
> I do not have a BASE_PATH specified and when loading Maintainerr, the home screen shows 404 error. When clicking Overview from the side, it loads, but the URL path includes https://maintainerr.domain-here.com/__PATH_PROXY__/overview. So does each of the pages, it seems that when BASE_PATH is blank, it adds this to the URL.
> 
> I have my user setup to match the file system permissions.

Please fill out a bug report: https://github.com/jorenn92/Maintainerr/issues/new?assignees=&labels=&projects=&template=bug_report.md&title=

PR #1407, opened 2024-11-28
feat: Add "Original Language" rule to Sonarr & Radarr

benscobie:
For feature request: https://features.maintainerr.info/posts/52/allow-to-filter-based-on-the-original-language-value-in-sonarr

I've fixed the Radarr & Sonarr server selection not persisting on initial rule creation. I've also fixed an exception/logic error when checking the cacheReset flag.

### DIFF
--- a/server/src/database/migrations/1733357722729-Remove-default-arr-servers.ts
+++ b/server/src/database/migrations/1733357722729-Remove-default-arr-servers.ts
@@ -1,0 +1,58 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class RemoveDefaultArrServers1733357722729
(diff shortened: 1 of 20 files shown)


benscobie:
@ydkmlt84 Give this latest commit a try. You now have to select a library & server instead of it defaulting to something

benscobie:
> EDIT: I still need to make the 'None' server action work. Or maybe I'll do that in a separate PR and remove the option here for now.

Actually this already works, I missed the existing code that already handles this. When the 'None' option is selected we'll attempt to delete the media via Plex.

ydkmlt84 (approved):
Looks great. I like the changes to library and server selections! Thanks again for helping with the project!!

PR #1408, opened 2024-12-02
feat: add about page

ydkmlt84:
I somehow messed up the yarn.lock file and needed to reset.

- shows releases and a button to view changelog
- shows total number of rules
- shows version
- total media currently in collections
- number of community rules
- useful links section
- donations section

### DIFF
--- a/ui/src/components/Settings/About/index.tsx
+++ b/ui/src/components/Settings/About/index.tsx
@@ -1,0 +1,289 @@
+import { InformationCircleIcon } from '@heroicons/react/solid'
+import React, { useEffect, useState } from 'react'
+import GetApiHandler from '../../../utils/ApiHandler'
+import Releases from './Releases'
+
+interface VersionResponse {
+  status: 1 | 0
+  version: string
+  commitTag: string
+  updateAvailable: boolean
+}
+
+const AboutSettings = () => {
+  useEffect(() => {
+    document.title = 'Maintainerr - Settings - About'
+  }, [])
+
+  // Maintainerr Timezone
+  const [timezone, setTimezone] = useState<string>('')
+  useEffect(() => {
+    GetApiHandler<string>('/app/timezone').then((resp) => {
+      setTimezone(resp)
+    })
+  }, [])
+  // End Maintainerr Timezone
+
+  // Maintainerr Version
+  const [version, setVersion] = useState<string>('')
+  const [commitTag, setCommitTag] = useState<string>('')
+  useEffect(() => {
+    GetApiHandler('/app/status').then((resp: VersionResponse) => {
+      if (resp.status) {
+        setVersion(resp.version)
+        setCommitTag(resp.commitTag)
+      }
+    })
+  }, [])
+  // End Maintainerr Version
+
+  // Maintainerr Rules Count
+  const [ruleCount, setRuleCount] = useState<number>()
+  useEffect(() => {
+    GetApiHandler<number>('/rules/count/').then((resp) => {
+      setRuleCount(resp)
+    })
+  }, [])
+  // End Maintainerr Rules Count
+
+  // Maintainerr Collection Items
+  const [itemCount, setItemCount] = useState<number>()
+  useEffect(() => {
+    GetApiHandler<number>('/collections/media/count').then((resp) => {
+      setItemCount(resp)
+    })
+  }, [])
+  // End Maintainerr Collection Items
+
+  // Maintainerr Community Rules Count
+  const [communityCount, setCommunityCount] = useState<number>()
+  useEffect(() => {
+    GetApiHandler<number>('/rules/community/count').then((resp) => {
+      setCommunityCount(resp)
+    })
+  }, [])
+  // End Maintainerr Community Rules Count
+
+  return (
+    <div className="h-full w-full">
+      <div className="mt-6 rounded-md border border-amber-600 bg-amber-500 bg-opacity-20 p-4 backdrop-blur">
+        <div className="flex">
+          <div className="flex-shrink-0">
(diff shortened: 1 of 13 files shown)


ui/src/components/Settings/About/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+        }
+        setRuleCount(totalCount)
+      }
+    })
+  })
+  // End Maintainerr Rules Count
+  // Maintainerr Collection Items
+  const [itemCount, setItemCount] = useState<number | null>(null) // To store the count
+  useEffect(() => {
+    GetApiHandler('/collections/media').then((resp) => {
```
benscobie, line 76: You would be much better off introducing a new endpoint like `/collections/media/count` which returns only the count instead of bringing back all the data just to count it.

ui/src/components/Settings/About/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+        }
+        setItemCount(totalCount)
+      }
+    })
+  })
+  // End Maintainerr Collection Items
+  // Maintainerr Community Rules Count
+  const [communityCount, setCommunityCount] = useState<number | null>(null) // To store the count
+  useEffect(() => {
+    GetApiHandler('/rules/community').then((resp) => {
```
benscobie, line 90: Same comment as https://github.com/jorenn92/Maintainerr/pull/1408/files#r1885896395

ui/src/components/Settings/About/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+        setVersion(resp.version)
+        setCommitTag(resp.commitTag)
+      }
+    })
+  }, [])
+  // End Maintainerr Version
+  // Maintainerr Rules Count
+  const [ruleCount, setRuleCount] = useState<number | null>(null) // To store the count
+  useEffect(() => {
+    GetApiHandler('/rules').then((resp) => {
```
benscobie, line 62: Same comment as https://github.com/jorenn92/Maintainerr/pull/1408/files#r1885896395

ui/src/components/Settings/About/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+        </div>
+        <hr className="my-2 h-px border-0 bg-gray-200 dark:bg-gray-700"></hr>
+        <div className="form-row my-2">
+          <label htmlFor="name" className="text-label">
+            Time Zone
+          </label>
+          <div className="form-input">
+            <div className="form-input-field">
+              <span className="">
+                <code>{getBrowserTimezone()}</code>
```
benscobie, line 164: I'm not sure that the time zone being here is useful. The server might be using a different timezone to what the browser is using.

ui/src/components/Settings/About/index.tsx
```diff
@@ -0,0 +1,322 @@
...
+        </div>
+        <hr className="my-2 h-px border-0 bg-gray-200 dark:bg-gray-700"></hr>
+        <div className="form-row my-2">
+          <label htmlFor="name" className="text-label">
+            Number Of Rules
+          </label>
+          <div className="form-input">
+            <div className="form-input-field">
+              <span className="">
+                <code>{<>{ruleCount}</>}</code>
```
benscobie, line 177: These can all be simplified to:

```suggestion
                <code>{ruleCount}</code>
```

ui/src/components/Settings/About/index.tsx
ydkmlt84: I agree. I wanted some more data in there when I initially wrote it, but didn't know how to get what TZ the container was set to. Ideally, I would get the information from the container's environment variables. I would like to use that for the host data volume location as well. Do you know how we can get docker environment information into the app?

ui/src/components/Settings/About/index.tsx
ydkmlt84: Wouldn't it end up doing the same thing? Just on the backend? I don't know, but I assume the endpoint would have to do the same thing and then return the count?

ui/src/components/Settings/About/index.tsx
benscobie: For the timezone you can return it from the API. This will return the IANA time zone: `Intl.DateTimeFormat().resolvedOptions().timeZone`

There is no environment variable for the data volume, it's always `/opt/data` so hard coding it like you have is fine.

ydkmlt84: I would like to pull the TZ information from whatever environment variable they have set in their compose. 
It is currently coded like you stated. 
```
  const getBrowserTimezone = (): string => {
    return Intl.DateTimeFormat().resolvedOptions().timeZone
  }
```
As for the /opt/data... I was thinking of showing what host dir they have /opt/data mapped to. Probably a little much. Should just keep it simple.

ui/src/components/Settings/About/index.tsx
benscobie: No, you would use a COUNT SQL query, with typeorm it would look something like this:
```
const count = await this.CollectionMediaRepo.count({
  where: { collectionId: id },
})```

ui/src/components/Settings/About/index.tsx
benscobie: Your code is being executed on the client so it will show the browser's time zone. Run what you've got on the server and it will return the time zone that Node is using, which it will get from the environment.

ydkmlt84: > Your code is being executed on the client so it will show the browser's time zone. Run what you've got on the server and it will return the time zone that Node is using, which it will get from the environment.

Makes sense.

benscobie (approved):

PR #1421, opened 2024-12-07
feat: Add "Show on library recommended" option

Gabri3lZ:
I did not find an existing feature request for this. But since it is a rather straight forward addition, I thought lets just go for it.

This feature adds a new checkbox option to the "Rules" editor, to show the correlating collection on the plex library recommended page (similar to the existing "Show on home" option).

Notably, this involved a database change for the "collection" table. I was not entirely sure how the typeorm database migration generator works. I had to make manual changes to the generated migration file.


benscobie (approved):
Nice one, thank you for your contribution 😄

PR #1448, opened 2024-12-20
fix: Community rules & incorrect out of date

benscobie:
Both bugs were caused by not being able to detect the current version. I've updated the semantic-release config to stamp the workspace package.json's with the release version and have updated how we start the server so that npm_package_version gets populated.


ydkmlt84 (approved):
Looks good on my end, for those two issues.

PR #1450, opened 2024-12-20
chore: moving docs into the repo and generating github page

ydkmlt84:
This PR adds the documentation back into the repo, with a workflow to generate the Github pages. Working on creating the page through the GH CLI commands, since we don't have access to that on the webui. Some setup is required from within the repo to add a page to the repo.

These docs would temporarily be hosted at `jorenn92.github.io/Maintainerr` and once verified, would be switched to the custom `docs.maintainerr.info` domain name.

The structure of the files is the same as in the dedicated docs repo. The only thing I can't test at the moment is if the user-requirements file is going to work with the workflow action. This is really only needed for the swagger-ui plugin, for API documentation display.

See https://docs-dev.maintainerr.info for an example, which is currently hosted using a github page, from the docs repo.


.github/workflows/mkdocs_ci.yml
```diff
@@ -0,0 +1,31 @@
+name: mkdocs
+on:
+  push:
+    branches:
+      - main
```
benscobie, line 5: We could add the docs paths here so we only deploy when necessary.

```suggestion
    branches:
      - main
    paths: ['docs/**','.github/**','mkdocs.yml','user-requirements.txt']
```

ydkmlt84: Does that mean only when changes are made to those files?

benscobie: Yes exactly, so merges with non-docs changes don't trigger a docs deploy.

ydkmlt84: I like it. Except for stuff that has API changes, when we're eventually auto generating that part.  

Actually @jorenn92 seems to respond more when it's on GitHub. 😂

PR #1452, opened 2024-12-20
fix: Resolve various UI & UX issues

benscobie:


ydkmlt84 (approved):
Looks good on my end.

PR #1467, opened 2024-12-23
fix: Do not reset collection on *arr server change

benscobie:
Not sure what I was thinking when I added this in. It doesn't make any sense to reset the collection when changing the *arr server, it's similar to changing rule logic which does not trigger a collection reset so why should this.


ydkmlt84 (approved):
👍🏻

PR #1471, opened 2024-12-28
feat: Add season has season/series finale episode

benscobie:
Adds two new filters for seasons; **Season has season finale episode** and **Season has series finale episode**.

I have also resolved a client side error when changing rule data/media type, as well as improving & resolving some types.

Resolves feature request: https://features.maintainerr.info/posts/18/add-sonarrs-season-finale-label-as-seasons-rule-filter

### DIFF
--- a/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
+++ b/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
@@ -49,7 +49,9 @@ export class SonarrApi extends ServarrApi<{
         }${episodeIds ? `&episodeIds=${episodeIds}` : ''}`,
       );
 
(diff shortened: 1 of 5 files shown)


ydkmlt84:
The wording of the rule in the dropdown list is confusing on first read. Maybe take away the first `Season` because the rule is only available in the seasons media type anyways? So, it would read "Has season finale episode" or " Has series finale episode"?

benscobie:
> The wording of the rule in the dropdown list is confusing on first read. Maybe take away the first `Season` because the rule is only available in the seasons media type anyways? So, it would read "Has season finale episode" or " Has series finale episode"?

Sure, done.

ydkmlt84 (approved):
👍🏻

PR #1475, opened 2024-12-29
fix: *arr server not persisting on initial save

benscobie:
The selected Sonarr/Radarr server was not being saved when creating a new rule.


ydkmlt84 (approved):
👍

PR #1488, opened 2025-01-02
feat: Add rating vote count filters

benscobie:
For example, you may only want to look at a rating when a certain vote count has been reached.


ydkmlt84 (approved):
👍 good idea

PR #1491, opened 2025-01-03
feat: Add "Last episode aired at (season)" for episodes

benscobie:
Similar to the filter **Last episode aired at** that targets shows & seasons, this one will return the last episode aired date for the season the episode belongs to.

This is to support this scenario: https://discord.com/channels/1152219249549512724/1324689589641941003

### DIFF
--- a/server/src/modules/collections/collections.service.ts
+++ b/server/src/modules/collections/collections.service.ts
@@ -1047,7 +1047,7 @@ export class CollectionsService {
     try {
       const result = await this.plexApi.getCollection(id);
 
-      if (result.smart) {
+      if (result?.smart) {
         this.logger.warn(
           `Plex collection ${id} is a smart collection which is not supported.`,
         );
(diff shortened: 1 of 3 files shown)


server/src/modules/collections/collections.service.ts
```diff
@@ -1047,7 +1047,7 @@ export class CollectionsService {
     try {
       const result = await this.plexApi.getCollection(id);
 
-      if (result.smart) {
+      if (result?.smart) {
```
benscobie, line 1050: This fixes an exception being thrown when the collection does not exist. The same value is returned so there is no functional difference, just stops an error log message.

benscobie:
A user has tested this: https://discord.com/channels/1152219249549512724/1324689589641941003/1324776072948092971

ydkmlt84 (approved):
👍

PR #1501, opened 2025-01-08
Rules page UI update

ydkmlt84:
- made to match or be more similar to the collections page
- updated some colors as well
- fixed an error where an `<a>` tag was inside of an `<a>` tag which was unnecessary, for the collection link on the Collections page.


ydkmlt84:
Don't review this one. I have made some changes since and will include them in the modal PR.

PR #1505, opened 2025-01-13
feat: add media modal to overview items

ydkmlt84:
- When clicking on the poster of an item on the Overview page and poster of the collection items, a popup will show with additional information, as well as links to the item in TMBD, Plex, and Tautulli. 
-  The additional information shows depending on the information it receives. If information doesn't exist for that item, it will not show.
- Changed the rules overview page to look like the Collections overview page, for continuity.
- Updated some colors on the collection overview page to match styling of the rules page.


ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,50 +47,53 @@ const RuleGroup = (props: {
...
+            Status
           </div>
-
-          <div className="my-0.5 flex text-sm sm:my-1">
-            <span className="mr-2 w-full max-w-prose overflow-hidden overflow-ellipsis font-semibold">
-              {props.group.description}
-            </span>
+          <div>
+            {props.group.isActive ? (
+              <span className="flex justify-center font-semibold text-green-600">
```
benscobie, line 64: To match the collections page:
```suggestion
              <span className="flex justify-center text-green-600">
```

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,50 +47,53 @@ const RuleGroup = (props: {
...
+                Active
+              </span>
+            ) : (
+              <span className="flex justify-center font-semibold text-red-600">
```
benscobie, line 68:
```suggestion
              <span className="flex justify-center text-red-600">
```

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,50 +47,53 @@ const RuleGroup = (props: {
...
-          ) : (
-            <span className="text-sm font-bold text-red-900">Inactive</span>
-          )}
-          <div className="m-auto mr-2 flex text-sm font-medium">
-            {`Library ${
+        <div>
+          <div className="m-auto mr-2 flex justify-center text-sm font-medium">
+            Library
+          </div>
+          <div className="flex justify-center font-semibold text-amber-500">
```
benscobie, line 78:
```suggestion
          <div className="flex justify-center text-amber-500">
```

ui/src/components/Rules/RuleGroup/index.tsx
benscobie: The font sizes are also different across the two pages

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,50 +47,53 @@ const RuleGroup = (props: {
...
+            {props.group.rules.length}
+          </div>
         </div>
       </div>
-
-      <div className="m-auto w-full sm:w-1/6">
-        <div className="mb-2 flex h-auto">
+      <div className="m-auto grid w-full grid-cols-1 gap-1 xl:grid-cols-2">
```
benscobie, line 95: I would join the two buttons together to align with the Overview & *arr settings pages

ui/src/components/Rules/index.tsx
```diff
@@ -120,17 +120,23 @@ const Rules: React.FC = () => {
...
+        <ul className="xs:collection-cards-vertical">
           {(data as IRuleGroup[]).map((el) => (
-            <RuleGroup
-              onDelete={refreshData}
-              onEdit={editHandler}
+            <li
               key={el.id}
-              group={el as IRuleGroup}
-            />
+              className="collection relative mb-5 flex h-fit transform-gpu flex-col overflow-hidden rounded-xl bg-zinc-800 bg-cover bg-center p-4 text-zinc-400 shadow ring-1 ring-zinc-700 xs:w-full sm:mb-0 sm:mr-5"
```
benscobie, line 130: The collections page cards have slightly more padding inside, this will align the two:

```suggestion
              className="collection relative mb-5 flex h-fit transform-gpu flex-col overflow-hidden rounded-xl bg-zinc-800 bg-cover bg-center p-7 text-zinc-400 shadow ring-1 ring-zinc-700 xs:w-full sm:mb-0 sm:mr-5"
```

ui/src/components/Rules/index.tsx
```diff
@@ -120,17 +120,23 @@ const Rules: React.FC = () => {
             />
           </div>
         </div>
-
-        <div>
+        <h1 className="mb-3 text-lg font-bold text-zinc-200">
+          {'Rules Listing'}
```
benscobie, line 124: I think we can simplify the wording to just Rules
```suggestion
          Rules
```

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,50 +47,53 @@ const RuleGroup = (props: {
...
-              {props.group.name}
-            </div>
+    <div className="flex-auto">
+      <div className="text-base font-bold text-white sm:text-lg">
+        {props.group.name}
+      </div>
+      <div className="mt-2 h-28 overflow-hidden overflow-y-auto overflow-ellipsis font-semibold">
+        {props.group.description}
+      </div>
+      <div className="mb-2 mt-2 grid grid-cols-3 gap-3">
```
benscobie, line 57: Very minor change but I think the buttons look too close to the text above
```suggestion
      <div className="mb-3 mt-2 grid grid-cols-3 gap-3">
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    )
+  },
+)
+
+export default MediaModalContent
```
benscobie, line 283:
```suggestion
MediaModalContent.displayName = 'MediaModalContent'

export default MediaModalContent
```

ui/src/components/Rules/RuleGroup/index.tsx
ydkmlt84: I am not sure I changed that. I mean Github says I did, but I don't remember looking at those things. I will update the PR to make it match the collections page, as I don't like the semibold on the description. I prefer the collection page's style, now that I am looking at that.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+            <div className="mr-0.5 mt-6 flex flex-row items-center justify-between">
+              {metadata?.Guid &&
+                ['movie', 'show'].includes(mediaType) &&
+                metadata.Guid.length > 0 && (
+                  <div className="flex flex-wrap items-center text-xs text-zinc-400">
+                    {metadata.Guid.map((guid, index) => (
+                      <span
+                        key={index}
+                        className="mb-0.5 mr-0.5 flex h-8 items-center justify-center rounded-lg bg-zinc-700 bg-opacity-70 px-2 text-xs text-white shadow-lg"
```
benscobie, line 259: I think it's cleaner to use gap on the parent for margins

```suggestion
                        className="flex h-8 items-center justify-center rounded-lg bg-zinc-700 bg-opacity-70 px-2 text-xs text-white shadow-lg"
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+            <div className="mt-2 text-gray-300">
+              <p>{summary || 'No summary available.'}</p>
+            </div>
+
+            <div className="mr-0.5 mt-6 flex flex-row items-center justify-between">
+              {metadata?.Guid &&
+                ['movie', 'show'].includes(mediaType) &&
+                metadata.Guid.length > 0 && (
+                  <div className="flex flex-wrap items-center text-xs text-zinc-400">
```
benscobie, line 255:
```suggestion
                  <div className="flex flex-wrap items-center text-xs text-zinc-400 gap-1">
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+                  {title} ({year})
+                </h2>
+              </div>
+            </div>
+
+            <div className="mt-2 text-gray-300">
+              <p>{summary || 'No summary available.'}</p>
+            </div>
+
+            <div className="mr-0.5 mt-6 flex flex-row items-center justify-between">
```
benscobie, line 251: Adds some margin to stop the button and list items touching eachother:
image

```suggestion
            <div className="mr-0.5 mt-6 flex flex-row items-center justify-between gap-4">
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+                          audience: '/icons_logos/imdb_icon.svg',
+                        },
+                        rottentomatoes: {
+                          critic: '/icons_logos/rt_critic.svg',
+                          audience: 'icons_logos/rt_audience.svg',
+                        },
+                        themoviedb: {
+                          audience: '/icons_logos/tmdb_icon.svg',
+                        },
+                      }
```
benscobie, line 196: This can be moved to the top of the file as it's static

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+      )
+      GetApiHandler<Metadata>(`/plex/meta/${id}`).then((data) => {
+        setMetadata(data)
+        setLoading(false)
+      })
+    }, [id])
+
+    return (
+      <div
+        className="fixed inset-0 z-50 flex items-start items-center justify-center bg-black bg-opacity-70 px-3"
```
benscobie, line 89:
```suggestion
        className="fixed inset-0 z-50 flex items-center justify-center bg-black bg-opacity-70 px-3"
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+        onClick={onClose} // Close modal when clicking outside
+      >
+        <div
+          className="relative max-h-[90vh] w-full max-w-4xl overflow-auto rounded-xl bg-zinc-800 shadow-lg sm:w-3/4"
+          onClick={(e) => e.stopPropagation()} // Prevent modal close on content click
+        >
+          {/* Top Half with Background Image */}
+          <div className="relative h-0 w-full overflow-hidden p-2 md:h-60 lg:h-60 xl:h-96">
```
benscobie, line 97: I think the top section should show on small screens as well. I can see the way it's currently setup using absolute positioning that things can overlap.

You can get rid of the absolute positioning of those individual items and instead do something like this:
`
            <div className="absolute left-0 top-0 z-10 flex h-full w-full gap-x-4 p-4">
              <div className="flex grow flex-col">
                <div className="grow">Movie / rated</div>
                <div>IMDb ratings etc</div>
              </div>
              <div className="flex flex-col items-end">
                <div className="grow">TMDB & Plex logos</div>
                <div>Genre tags</div>
              </div>
            </div>
`

ui/src/components/Common/MediaCard/index.tsx
```diff
@@ -52,17 +55,28 @@ const MediaCard: React.FC<IMediaCard> = ({
...
+    }
+  }
+  const closeMediaModal = () => setShowMediaModal(false)
 
   useEffect(() => {
     if (tmdbid) {
       GetApiHandler(`/moviedb/image/${mediaType}/${tmdbid}`).then((resp) =>
         setImage(resp),
       )
+      GetApiHandler(`/moviedb/backdrop/${mediaType}/${tmdbid}`).then((resp) =>
```
benscobie, line 77: This creates a lot of extra requests on the overview page for something the user might not ever look at. It would be better to move this into the MediaModalContent component.

ui/src/components/Layout/index.tsx
```diff
@@ -84,7 +84,7 @@ const Layout: React.FC<{ children?: ReactNode }> = (props: {
         </div>
 
         <main
-          className="relative top-16 z-0 mt-2 w-full focus:outline-none"
+          className="relative top-16 z-0 mt-2 w-full focus:outline-none lg:z-40 xl:z-40"
```
benscobie, line 87: What was this for? This doesn't seem like the right way to achieve whatever it was.

benscobie: Ah right I see, the modal isn't overlaying the top nav so you can click on it. We should be able to fix that though, which should remove the need for this. I'll have a look soon.

ui/src/components/Rules/RuleGroup/index.tsx
ydkmlt84: I actually changed it to use the same styles and div layout as from the Collection page. Real difference is that the bottom row here are buttons instead of information like on the collection page. The button row styling and the info row styling here have different styles than the collection page, because one has three columns and the Collection page only has two (as a flex row instead of a grid). This makes the mobile layout different though, as the buttons stack here and on the collection page the info items are made into 4 columns.

ui/src/components/Rules/RuleGroup/index.tsx
ydkmlt84: Should be fixed.

ui/src/components/Rules/RuleGroup/index.tsx
ydkmlt84: I removed the gap but don't know how to remove the rounded style, since they are rounded in their index.tsx. I thought about changing the style of the button file itself, as this is the only place the Edit Button is used, but then that leaves the Delete button still rounded. Maybe you know a better way.

ui/src/components/Rules/index.tsx
ydkmlt84: Replaced with the same CSS from the collections page. This change made the rule card a little too "squished" in my opinion.

ui/src/components/Rules/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Done

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Done. It works and is a good solution. Only issue I saw, with it this way, is when the list of genres gets longer. I used flex-wrap but it leaves a gap depending on how long the genre "word" is. Otherwise this was a great solution.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+                      )
+                    })}
+                  </div>
+                ) : (
+                  ''
+                )}
+                {metadata.Genre && metadata.Genre.length > 0 ? (
+                  <div className="pointer-events-none absolute bottom-4 right-4 z-10 flex flex-wrap">
+                    {metadata.Genre.map((genre, index) => (
+                      <span
```
ydkmlt84, line 223: I actually fixed the before mentioned genre lenght issue by only using the first three genres. Do you think this is a good solution?

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Ok I fixed that by limiting the genres to three. I also changed genres to a column on smaller screens.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,283 @@
...
+        document.body.classList.add('fixed', 'overflow-hidden', 'w-full')
+      } else {
+        document.body.classList.remove('fixed', 'overflow-hidden', 'w-full')
+      }
+      return () => {
+        document.body.classList.remove('fixed', 'overflow-hidden', 'w-full')
+      }
+    }, [show])
+
+    if (!show) return null
```
ydkmlt84, line 59: I changed this to allow background scrolling. It still allows modal scrolling if need be. I didn't like how if you opened a modal, you were put back at the top of the page.

ui/src/components/Rules/RuleGroup/index.tsx
ydkmlt84: It took me too long to figure out that the `<DeleteButton>` used in Radarr and Sonarr wasn't from the delete button in the common components. Since the rulegroup page is the only one using the Edit and Delete button common components, I just changed their style. I also changed it a litle for when they are on top of each other.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
benscobie: Sorry I meant outside of the component itself (MediaModalContent), same level as `interface Metadata`

ui/src/components/Common/MediaCard/MediaModal/index.tsx
benscobie: > Done. It works and is a good solution. Only issue I saw, with it this way, is when the list of genres gets longer. I used flex-wrap but it leaves a gap depending on how long the genre "word" is. Otherwise this was a great solution.

Could you show me an example of what you mean?

I wonder if this fixes your issue: `pointer-events-none flex flex-wrap-reverse items-end justify-end gap-1`

This will wrap in reverse so the overflowing ones go upwards instead of down, so you generally have a longer list at the bottom which looks nicer. I also changed to items-end so the items are always right aligned.

Also with the breakpoints like this `md:flex-row lg:flex-row xl:flex-row` you do not need `lg` and `xl`. md will apply to that and the larger breakpoints, see https://v3.tailwindcss.com/docs/responsive-design#targeting-a-breakpoint-range

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+                          : mediaType === 'season'
+                            ? 'bg-yellow-700'
+                            : 'bg-rose-900'
+                    }`}
+                  >
+                    {mediaType}
+                  </div>
+                  {metadata?.contentRating && (
+                    <div className="pointer-events-none mt-1 rounded-lg bg-black bg-opacity-70 p-2 text-xs font-medium uppercase text-zinc-200">
+                      {`Rated:${metadata.contentRating}`}
```
benscobie, line 125:
```suggestion
                      {`Rated: ${metadata.contentRating}`}
```

ui/src/components/Common/MediaCard/MediaModal/index.tsx
benscobie: This adds a load of warnings for useEffect etc. I use a scroll lock for modals in another app which doesn't have the issue you describe. I'll dig the code out.

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -47,72 +47,76 @@ const RuleGroup = (props: {
...
+            </div>
+            <div>
+              {props.group.isActive ? (
+                <span className="flex justify-center text-green-600">
+                  Active
+                </span>
+              ) : (
+                <span className="flex justify-center text-red-600">
+                  Inactive
+                </span>
```
benscobie, line 73: The collection page uses text-green-500 and text-red-500 for these. Can we align them?

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: image

This is with the CSS you suggested, which looks much better. I think the wrap-reverse and items-end were the winners. It does get hidden underneath the top right links when the screen is super small, but I don't know what device anyone would be using it on that is that small.

image

Also, it gets hidden under the search bar, which I would like to get rid of if not on the overview page anyways, but that is just me.  Also, looks like the bottom left items could use some padding. I will look at that.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: Which part do you mean? Having it with background scrolling, or something else?

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: With the flex-wrap-reverse and items-end change, we can go back to pulling all genres.

ui/src/components/Layout/index.tsx
ydkmlt84: This is one example of what I was trying to fix.

image

If I fix that by making it just z-40, regardless of screen size, it looks like this.

image

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: I made both bottom items match.

image

ui/src/components/Layout/index.tsx
benscobie: Ah yes I was meant to look into this sorry!

ui/src/components/Common/MediaCard/MediaModal/index.tsx
benscobie: Having the scroll disabled but not having it scroll back to the top of the page when it opens

ui/src/components/Common/MediaCard/MediaModal/index.tsx
benscobie: > Also, it gets hidden under the search bar, which I would like to get rid of if not on the overview page anyways, but that is just me. Also, looks like the bottom left items could use some padding. I will look at that.

That's just the text overflowing as it's ran out of space. You could reduce the gap-X on smaller screens to get a bit more space back, though a 216px screen is super small, I typically target 360px as that's mostly the smallest resolution these days.

ydkmlt84: Yeah there were different ways to fix it but I didn't like what it did in other places. I ended up leaving it. So that now they are both displayed like the picture above, with both being a row and flex-wrapping in reverse. To me it looks more uniform that way anyways.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
```diff
@@ -0,0 +1,279 @@
...
+      GetApiHandler<Metadata>(`/plex/meta/${id}`).then((data) => {
+        setMetadata(data)
+        setLoading(false)
+      })
+      GetApiHandler(`/moviedb/backdrop/${mediaType}/${tmdbid}`).then((resp) =>
+        setBackdrop(resp),
+      )
+    }, [id])
+
+    return (
```
ydkmlt84, line 82: This fixes the background scroll, as well as the eslint feedback. It also doesn't jump back up to the top of the page like it did before.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: See below comment. Now that eslint is working, I was able to correct my mistakes and implement the scroll lock.

ui/src/components/Common/MediaCard/MediaModal/index.tsx
ydkmlt84: fixed

ui/src/components/Layout/index.tsx
benscobie: This should be fixed with my latest commit

ydkmlt84 (commented):
For the open modal onclick event...I went back and forth with doing it in that div but ultimately, I left the onclick in the on-hover div because without it you can't see the add or excl buttons. It wouldn't work when you first clicked on the cover, because it showed the add/excl buttons, but then you could click on the image again and open the modal. 

Obviously, we are talking about touch devices only. Non-touch devices those buttons still show up on hover, and then you can click on the image.

This way, it opens the modal right away and doesn't show you those options.

benscobie:
> For the open modal onclick event...I went back and forth with doing it in that div but ultimately, I left the onclick in the on-hover div because without it you can't see the add or excl buttons. It wouldn't work when you first clicked on the cover, because it showed the add/excl buttons, but then you could click on the image again and open the modal.
> 
> Obviously, we are talking about touch devices only. Non-touch devices those buttons still show up on hover, and then you can click on the image.
> 
> This way, it opens the modal right away and doesn't show you those options.

Ah yea I missed that on mobile, I'll have another play and see if I can come up with something that works good for both.

benscobie:
@ydkmlt84 See what you think about that latest commit. I believe this works the same way as you had it before. On touch devices you click the poster to show the buttons and then you can click it again to show the modal.

ydkmlt84:
> @ydkmlt84 See what you think about that latest commit. I believe this works the same way as you had it before. On touch devices you click the poster to show the buttons and then you can click it again to show the modal.

That did the trick. Everything else looks good to me. I put in a lot of checks for when something isn't right, to not display it...I am curious to see what will happen on one of these items that people have with the incorrect metadata.

benscobie (approved):

PR #1506, opened 2025-01-13
feat: Added a boolean check for watchlisted status

ydkmlt84:
I tested it the best that I could.

### DIFF
--- a/server/src/modules/rules/getter/plex-getter.service.ts
+++ b/server/src/modules/rules/getter/plex-getter.service.ts
@@ -496,7 +496,7 @@ export class PlexGetterService {
           const filterResults = await Promise.all(userFilterPromises);
           const result = plexUsers.filter((_, index) => filterResults[index]);
 
-          return result.length > 0 ? 1 : 0;
+          return result.length > 0;
         }
         case 'sw_seasonLastEpisodeAiredAt': {
           const lastEpDate = await this.plexApi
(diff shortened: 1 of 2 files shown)


server/src/modules/rules/getter/plex-getter.service.ts
```diff
@@ -469,6 +469,35 @@ export class PlexGetterService {
...
+                watchlist.find((i) => i.id === media_uuid[1]) !== undefined
+              );
+            } catch (e) {
+              return false;
+            }
+          });
+          const filterResults = await Promise.all(userFilterPromises);
+          const result = plexUsers.filter((_, index) => filterResults[index]);
+
+          return result.length > 0 ? 1 : 0;
```
benscobie, line 499:
```suggestion
          return result.length > 0;
```

ydkmlt84: Do I not need it to return a 1 or a 0 based on the length? Or does it do that with the way you suggested?

benscobie: `result.length > 0` will return a boolean, [see this](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_operators#comparison_operators). I suspect that 0 and 1 worked though because we're doing a [loose comparison](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness) somewhere, but we should stick to booleans types.

ydkmlt84: Makes sense!

benscobie (approved):

PR #1528, opened 2025-01-23
fix: update optional page route param to new syntax

bitnimble:
A fresh clone of the repo + `yarn && yarn dev` results in the following error, as `path-to-regexp` is locked to v8.2.0 in `yarn.lock`:

image

This PR simply fixes the route param to use the new optional syntax.


ben-propflo (approved):
I really need to remember to switch GH account!

benscobie (approved):
Thank you. We really need to add a test to load the app up to catch these kind of things. This broke because there was a nestjs major version bump that shouldn't have been auto merged. I've added a commit to stop that happening.

PR #1530, opened 2025-01-24
feat: allow comparing list sizes against a number

bitnimble:
This PR:
- Creates new rule types, `NUMBER_LIST` and `TEXT_LIST` (I chose "list" to align with the term already used in the UI/copy)
- Adds new rule actions for those rule types - `COUNT_EQUALS`, `COUNT_NOT_EQUALS`, `COUNT_BIGGER`, `COUNT_SMALLER`
- Moves the existing rule possibilities/actions that were only applicable to lists from `NUMBER` and `TEXT` to the new rule types
- Updates all existing rules that are lists to use the new RuleType
- Adds tests, updates the frontend, etc to accommodate for the new rule type
- Also does a small amount of cleanup of types, which was so I could make sure that I wasn't accidentally missing any cases when updating `doRuleAction` to support the new potential value types. I want to do a lot more cleanup here, but I'll do that in its own refactoring PR instead.

I _think_ this requires no migration, since all rule actions that could have been applicable to list values previously are still valid on the new rule types (it is a superset). But I'm not 100% sure if I'm looking at the right place for how the rule actions are serialised and persisted.

image

image

### DIFF
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -84,24 +84,8 @@ export class RuleType {
     [RulePossibility.EQUALS, RulePossibility.NOT_EQUALS],
     'boolean',
   );
-  static readonly NUMBER_LIST = new RuleType(
-    '4',
-    [
-      RulePossibility.EQUALS,
-      RulePossibility.NOT_EQUALS,
-      RulePossibility.CONTAINS,
-      RulePossibility.NOT_CONTAINS,
-      RulePossibility.CONTAINS_PARTIAL,
-      RulePossibility.NOT_CONTAINS_PARTIAL,
-      RulePossibility.COUNT_EQUALS,
-      RulePossibility.COUNT_NOT_EQUALS,
-      RulePossibility.COUNT_BIGGER,
-      RulePossibility.COUNT_SMALLER,
-    ],
-    'number list',
-  );
   static readonly TEXT_LIST = new RuleType(
-    '5',
+    '4',
     [
       RulePossibility.EQUALS,
       RulePossibility.NOT_EQUALS,
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -469,7 +469,7 @@ export class RuleConstants {
           mediaType: MediaType.SHOW,
           showType: [EPlexDataType.SEASONS, EPlexDataType.EPISODES],
           cacheReset: true,
-          type: RuleType.TEXT,
+          type: RuleType.TEXT_LIST,
         },
         {
           id: 42,
@@ -477,7 +477,7 @@ export class RuleConstants {
           humanName:
             '[list] Collections media is present in (titles) (incl. smart collections)',
           mediaType: MediaType.BOTH,
-          type: RuleType.TEXT,
+          type: RuleType.TEXT_LIST,
           cacheReset: true,
         },
       ],
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -563,9 +563,9 @@ export class RuleConstants {
         {
           id: 9,
           name: 'fileQuality',
-          humanName: '[list] File - quality (2160, 1080,..)',
+          humanName: 'File - quality (2160, 1080,..)',
           mediaType: MediaType.MOVIE,
-          type: RuleType.NUMBER_LIST,
+          type: RuleType.NUMBER,
         },
         {
           id: 10,
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -514,7 +514,7 @@ export class RuleConstants {
         {
           id: 2,
           name: 'tags',
-          humanName: '[list] Tags (Text if 1, otherwise list)',
+          humanName: '[list] Tags',
           mediaType: MediaType.MOVIE,
           type: RuleType.TEXT_LIST, // return text[]
         },
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -170,7 +170,7 @@ export class RuleConstants {
           name: 'seenBy',
           humanName: '[list] Viewed by (username)',
           mediaType: MediaType.MOVIE,
-          type: RuleType.TEXT, // returns usernames []
+          type: RuleType.TEXT_LIST, // returns usernames []
         },
         {
           id: 2,
@@ -191,7 +191,7 @@ export class RuleConstants {
           name: 'people',
           humanName: '[list] People involved',
           mediaType: MediaType.BOTH,
-          type: RuleType.TEXT, // return text[]
+          type: RuleType.TEXT_LIST, // return text[]
         },
         {
           id: 5,
@@ -220,7 +220,7 @@ export class RuleConstants {
           name: 'fileVideoResolution',
           humanName: '[list] Media file resolution (4k, 1080,..)',
           mediaType: MediaType.MOVIE,
-          type: RuleType.TEXT,
+          type: RuleType.TEXT_LIST,
         },
         {
           id: 9,
@@ -241,14 +241,14 @@ export class RuleConstants {
           name: 'genre',
           humanName: '[list] List of genres (Action, Adventure,..)',
           mediaType: MediaType.BOTH,
-          type: RuleType.TEXT, // return text[]
+          type: RuleType.TEXT_LIST, // return text[]
         },
         {
           id: 12,
           name: 'sw_allEpisodesSeenBy',
           humanName: '[list] Users that saw all available episodes',
           mediaType: MediaType.SHOW,
-          type: RuleType.TEXT, // return usernames []
+          type: RuleType.TEXT_LIST, // return usernames []
           showType: [EPlexDataType.SHOWS, EPlexDataType.SEASONS],
         },
         {
@@ -295,7 +295,7 @@ export class RuleConstants {
           name: 'sw_watchers',
           humanName: '[list] Users that watch the show/season/episode',
           mediaType: MediaType.SHOW,
-          type: RuleType.TEXT, // return usernames []
+          type: RuleType.TEXT_LIST, // return usernames []
           showType: [
             EPlexDataType.SHOWS,
             EPlexDataType.SEASONS,
@@ -307,7 +307,7 @@ export class RuleConstants {
           name: 'collection_names',
           humanName: '[list] Collections media is present in (titles)',
           mediaType: MediaType.BOTH,
-          type: RuleType.TEXT,
+          type: RuleType.TEXT_LIST,
           cacheReset: true,
         },
         {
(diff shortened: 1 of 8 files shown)


server/src/modules/rules/getter/getter.service.ts
```diff
@@ -24,7 +24,7 @@ export class ValueGetterService {
     libItem: PlexLibraryItem,
     ruleGroup?: RulesDto,
     dataType?: EPlexDataType,
-  ) {
+  ): Promise<RuleValueType> {
```
bitnimble, line 27: Explicitly typed to avoid it virally turning variables into `any`

server/src/modules/rules/helpers/rule.comparator.service.ts
```diff
@@ -373,25 +375,25 @@ export class RuleComparatorService {
     this.workerData = [];
   }
 
-  private doRuleAction<T>(val1: T, val2: T, action: RulePossibility): boolean {
```
bitnimble, line 376: >val1: T, val2: T

`val1` and `val2` were never always the same type (e.g. string[] being compared against string with action CONTAINS), so

server/src/modules/rules/helpers/rule.comparator.service.ts
```diff
@@ -436,11 +438,11 @@ export class RuleComparatorService {
     if (action === RulePossibility.CONTAINS) {
       try {
         if (!Array.isArray(val2)) {
-          return (val1 as unknown as T[])?.includes(val2);
+          return (val1 as unknown[])?.includes(val2);
```
bitnimble, line 441: I have a refactor coming soon which will hopefully fix up all of these remaining casts :)

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
```diff
@@ -4,31 +4,19 @@ import { IRule } from '../'
 import ConstantsContext, {
   IProperty,
   MediaType,
+  RulePossibility,
   RulePossibilityTranslations,
 } from '../../../../../contexts/constants-context'
 import { EPlexDataType } from '../../../../../utils/PlexDataType-enum'
 import _, { first } from 'lodash'
 
-enum RulePossibility {
```
bitnimble, line 12: Found there were 2 duplicate enums; opting to delete this one and go with a single source of truth, given we already import from the other file

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
```diff
@@ -442,7 +430,7 @@ const RuleInput = (props: IRuleInput) => {
               value={secondVal}
             >
               <option value={undefined}> </option>
-              <optgroup label={`Custom value's`}>
+              <optgroup label="Custom values">
```
bitnimble, line 433: Unrelated; minor typo fix

ui/src/contexts/constants-context.tsx
```diff
@@ -58,6 +62,10 @@ export enum RulePossibilityTranslations {
   NOT_CONTAINS = 'Not Contains (Exact list match)',
   CONTAINS_PARTIAL = 'Contains (Partial list match)',
   NOT_CONTAINS_PARTIAL = 'Not Contains (Partial list match)',
+  COUNT_EQUALS = 'Count Equals',
+  COUNT_NOT_EQUALS = 'Count Does Not Equal',
+  COUNT_BIGGER = 'Count Is Bigger Than',
+  COUNT_SMALLER = 'Count Is Smaller Than',
```
bitnimble, line 68: gave the copy my best shot... not sure if I was expected to match the existing Title Casing 😅

server/src/modules/rules/constants/rules.constants.ts
```diff
@@ -49,8 +53,6 @@ export class RuleType {
       RulePossibility.SMALLER,
       RulePossibility.EQUALS,
       RulePossibility.NOT_EQUALS,
-      RulePossibility.CONTAINS,
```
benscobie, line 52: I think I understand this removal but just to check... Looking at doRuleAction it seems if the second value is not an array, which is extremely likely as you cannot provide a custom list for number, then we'll end up getting an exception as we'll try to call includes on it?

bitnimble: Haha, I'll double check this in the morning. This also made me realise that I forgot to add some tests for the `throw`s, and also to make sure there is actually a catch somewhere to stop things blowing up

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
benscobie: Something on my todo list is to add a shared package for these duplicated types as well as API schemas (though there currently isn't any, another todo). I refactored to use Turborepo recently to make this possible. I've referenced @maintainerr/server from the UI in one place as an example, but this breaks pretty badly if that included file imports some other stuff and I think the separate package will fix that.

server/src/modules/rules/helpers/rule.comparator.service.ts
```diff
@@ -500,26 +502,48 @@ export class RuleComparatorService {
...
+      assertIsArray(val1);
+      assertIsNumber(val2);
+      return val1.length > val2;
+    }
+    if (action === RulePossibility.COUNT_SMALLER) {
+      assertIsArray(val1);
+      assertIsNumber(val2);
+      return val1.length < val2;
+    }
   }
```
benscobie, line 547: It's possible to select a [list] for val2 so you'll want something like this `(Array.isArray(val2) ? val2.length : val)`

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
bitnimble: yeah, having a shared schema types package would be really nice. there's a few other places I saw that are just using open strings etc and assuming things will serialize/deserialize correctly, but they totally could just go out of sync very easily

server/src/modules/rules/constants/rules.constants.ts
bitnimble: the CONTAINS flow only calls `.includes` on val1 though, right? val2 could either be an array or a number and it'd still be fine 🤔 

looking here: https://github.com/jorenn92/Maintainerr/blob/0237cd4b43e722aef84849aab49df8aba2fdef02/server/src/modules/rules/helpers/rule.comparator.service.ts#L441

benscobie: Yes what you've done is fine. What would have happened is if val1 was a number and contains was used, it would run 2.includes(val2). Number doesn't have an includes() method so it would've thrown, get caught and return null instead.

I tested whether we'd need a migration and it's fine. Any rules configured with the removed ones would still execute but likely be broken because of the above issue anyway. When the rule UI is loaded the dropdown won't have anything selected so they'd need to reselect which is fine.

server/src/modules/rules/constants/rules.constants.ts
```diff
@@ -414,14 +455,14 @@ export class RuleConstants {
...
-          type: RuleType.NUMBER,
+          type: RuleType.NUMBER_LIST,
         },
         {
           id: 9,
           name: 'fileQuality',
           humanName: '[list] File - quality (2160, 1080,..)',
           mediaType: MediaType.MOVIE,
-          type: RuleType.NUMBER,
+          type: RuleType.NUMBER_LIST,
```
benscobie, line 465: It turns out that the getters for these just return a number so the comparison currently returns false as they aren't arrays. The API response types are set to number, but I'm sure you could have multiple audio channels. I'll have a dig into the Radarr API and see what's going on with that one. `fileQuality` will only ever be a single number as you can't have multiple files so that one needs to be set to RuleType.NUMBER.

I actually think we'll need to re-include contains for RuleType.NUMBER as it's no longer possible to compare to a list. E.g. you may want to target qualities of 720, 1080

benscobie: I can confirm that Radarr will only ever serve up a single number for audio channels, so that one can go back to RuleType.NUMBER now. I'm going to check all of the other [list]'s to see if we've got any more imposters...

server/src/modules/rules/constants/rules.constants.ts
```diff
@@ -83,6 +83,38 @@ export class RuleType {
...
+      RulePossibility.CONTAINS_PARTIAL,
+      RulePossibility.NOT_CONTAINS_PARTIAL,
+      RulePossibility.COUNT_EQUALS,
+      RulePossibility.COUNT_NOT_EQUALS,
+      RulePossibility.COUNT_BIGGER,
+      RulePossibility.COUNT_SMALLER,
+    ],
+    'number list',
+  );
+  static readonly TEXT_LIST = new RuleType(
```
benscobie, line 102: I just found that it's currently not possible to save a custom value for these types. It looks like there's a couple of functions that switch on the types to handle custom values:
See `server\src\modules\rules\constants\constants.service.ts`
`getCustomValueIdentifier`
`getCustomValueFromIdentifier`

server/src/modules/rules/constants/rules.constants.ts
```diff
@@ -372,7 +413,7 @@ export class RuleConstants {
           name: 'tags',
           humanName: '[list] Tags (Text if 1, otherwise list)',
```
benscobie, line 414:
```suggestion
          humanName: '[list] Tags',
```

This seems to always return a list even if there's 1 result, may have been a functionality change in the past.

server/src/modules/rules/constants/rules.constants.ts
bitnimble: aha good catch, thanks, just realised I didn't fully test end-to-end yet - only to the point of selecting it in the UI. I'll also actually run it on some rules and make sure it works...

server/src/modules/rules/constants/rules.constants.ts
benscobie: Forgot to come back to this, all of the other [list]'s are fine.

server/src/modules/rules/constants/rules.constants.ts
bitnimble: Seems working fully:

image
image
image

bitnimble:
There are a few places in the FE that are technically a bug because we now have a situation where the rule type !== value type. e.g. a rule that has a text_list type has a custom value type of `number` (since the operation is a "count").
This doesn't end up actually resulting in anything bad happening though, since one of the React effects causes it to re-parse the firstVal type, resetting the rule type back to the correct value... 

I'd like to fix that in the next PR since it unfortunately requires a big of rethinking around the `RuleType` type and that might end up being a change that touches quite a few places, and I don't believe it causes any visible issue at the moment.

bitnimble:
If we want to be sure that nothing breaks though, I can always start work on that change first, then rebase this PR on that once merged. I don't mind delaying this PR until that happens as I'm not urgently waiting for this feature or anything like that.

benscobie:
> This doesn't end up actually resulting in anything bad happening though, since one of the React effects causes it to re-parse the firstVal type, resetting the rule type back to the correct value...

That's another area I'd like to get cleaned up, there's a ton of inappropriate uses of useEffects that make changing things really hard.

server/src/modules/rules/constants/rules.constants.ts
benscobie: @bitnimble fileAudioChannels is still NUMBER_LIST

server/src/modules/rules/constants/rules.constants.ts
benscobie: The non-count lists do not though, e.g.:
image

benscobie (approved):
🚢 it

PR #1531, opened 2025-01-25
feat: Add the ability to clear Tautulli settings and save empty ones

### LINKED ISSUE #1390: Tautulli connection fails with leading slash in base url

mkaltner:
fix #1390

The purpose of this change is to add a clear button to the Tautulli settings page and remove the error check for saving empty ones.

### DIFF
--- a/ui/src/components/Settings/Tautulli/index.tsx
+++ b/ui/src/components/Settings/Tautulli/index.tsx
@@ -70,33 +70,45 @@ const TautulliSettings = () => {
         ? tautulli_url.slice(0, -1)
         : tautulli_url
 
-      const payload = {
-        tautulli_url: `${tautulli_url}${baseUrl ? `/${baseUrl}` : ''}`,
-        tautulli_api_key: apiKey,
-      }
+      await saveSettings(tautulli_url, apiKey)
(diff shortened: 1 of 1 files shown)


mkaltner:
Doesn't set the values to `null` in the database.  Will try to fix it correctly unless the maintainers implement it first.

PR #1542, opened 2025-01-27
chore: adding repo_dispatch to the release action

ydkmlt84:
Automatic push to docs repo action to trigger docs build and release for versioning.


benscobie (approved):
Is DOCS_DEPLOY set now or do we need @jorenn92 for this?

ydkmlt84:
> Is DOCS_DEPLOY set now or do we need @jorenn92 for this?

I set it with the API. You can see the list of secrets with `gh secret list` from within a local repo directory. You can do this with collaborator access.

PR #1545, opened 2025-01-28
feat: Add log settings, view and download

benscobie:
This PR introduces a new internal library package called contracts. This completes the idea I had when I first added turborepo and workspaces, which was to enable code sharing between the frontend and backend. As well as now sharing the DTOs, we also share validation logic using [Zod](https://zod.dev/). I have added a supporting package, [nestjs-zod](https://www.npmjs.com/package/nestjs-zod), which allows us to create a DTO from a Zod schema and easily use for Zod for validating request objects.

I have also added the Nest Swagger module, with URL `api/swagger`, and enabled the CLI plug-in so we don't have to manually add annotations. I had to add a separate process `generate-metadata.ts` to enable generating Swagger metadata for the new contracts project. This was start in watch mode when using `yarn dev`, and will build when using `yarn build`.

As for the package itself, I went with the compiled package approach as documented [here](https://turbo.build/repo/docs/core-concepts/internal-packages) as it sounded the most ideal, but does require a separate process as well. It runs the same as the Swagger metadata generator process with `yarn dev`and `yarn build`. The project uses [tsup](https://tsup.egoist.dev/) to bundle into ESM and CommonJS, as NestJS uses cjs and Next.js uses ESM. I had avoided ESM/CJS hell up until this point, it was not pleasant but I got there in the end. 

Lastly, I have added [React Hook Form](https://react-hook-form.com/) which simplifies form handling and integrates nicely with Zod to validate input. I have created components for input and select as well as grouped versions that contain the label and error message.

Now onto the actual user functionality:

This PR adds a tab in Settings called Logs. From here the user can:
- Set the log level, max size and max backups
  - A new table `log_settings` holds these settings. The DEBUG env var is no longer used but the log level DB setting will be set to debug if this is set.
- View and filter the live logs
  - The previous 200 lines of the newest log file are also returned. In the future I'd like this to be changeable to skip the need to download logs. I had a look at doing this but it got a bit complicated as old logs are archived.
  - Live logs uses a combination of [events](https://docs.nestjs.com/techniques/events) and [SSE (server-sent events)](https://docs.nestjs.com/techniques/server-sent-events).
- Download log files

Winston is now initialised within the nest module system as I needed to load the log level from the DB. I ended up removing `nest-winston` and creating our own logger as it was causing me issues with DI.

We're also now logging out the entire stack trace when an error is logged. Previously just the message was logged which includes no information around where the error actually occurred in code.


.vscode/settings.json
```diff
@@ -6,8 +6,13 @@
     "typescript",
     "typescriptreact"
   ],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "explicit",
+    "source.organizeImports": "explicit"
+  },
   "typescript.tsdk": "node_modules/typescript/lib",
   "typescript.enablePromptUseWorkspaceTsdk": true,
+  "typescript.tsserver.experimental.enableProjectDiagnostics": true,
```
benscobie, line 15: This enables TS errors for the whole project instead of just open files.

server/prettier.config.mjs
```diff
@@ -4,7 +4,6 @@
  */
 const config = {
   singleQuote: true,
-  trailingComma: 'all',
```
benscobie, line 7: This is the default setting anyway

ui/prettier.config.mjs
```diff
@@ -3,9 +3,7 @@
  * @type {import("prettier").Config}
  */
 const config = {
-  arrowParens: 'always',
   singleQuote: true,
-  tabWidth: 2,
```
benscobie, line 8: These are the default settings too

ui/src/pages/api/logs/stream.ts
```diff
@@ -0,0 +1,20 @@
+const apiPort = process.env.API_PORT || 3001
+
+export const config = {
+  runtime: 'edge',
+}
+
+export default async function handler() {
+  const { body } = await fetch(`http://localhost:${apiPort}/api/logs/stream`, {
```
benscobie, line 8: This page is required to get Next.js to proxy the SSE request correctly.

.vscode/settings.json
```diff
@@ -6,8 +6,13 @@
     "typescript",
     "typescriptreact"
   ],
+  "editor.codeActionsOnSave": {
+    "source.fixAll.eslint": "explicit",
+    "source.organizeImports": "explicit"
+  },
```
benscobie, line 12: More auto formatting options to keep everything consistent

ydkmlt84:
Does merging this PR effect my current auth PR or do we need to merge that one first for any reason?

ui/src/components/Settings/Logs/index.tsx
```diff
@@ -0,0 +1,324 @@
...
+        </thead>
+        <tbody className="divide-y divide-zinc-500 bg-zinc-700">
+          {pagedLogFiles.map((row, index: number) => {
+            return (
+              <tr key={`log-${index}`}>
+                <Table.TD>
+                  <a href={`${API_BASE_PATH}/api/logs/files/${row.name}`}>
+                    {row.name}
+                  </a>
+                </Table.TD>
```
ydkmlt84, line 290: What if you added a download icon button instead of using the name as a link?

                <Table.TD className="flex items-center space-x-2">
                  <span>{row.name}</span>
                  <DownloadIcon
                    className="h-5 w-5 cursor-pointer text-amber-500"
                    onClick={() => {
                      const link = document.createElement('a')
                      link.href = `${API_BASE_PATH}/api/logs/files/${row.name}`
                      link.download = row.name
                      document.body.appendChild(link)
                      link.click()
                      document.body.removeChild(link)
                    }}
                  />
                </Table.TD>

benscobie:
> Does merging this PR effect my current auth PR or do we need to merge that one first for any reason?

No it'll be fine as there's no overlap.

ydkmlt84 (approved):
👍

Issue #1140, opened 2024-06-21
Show does not delete after "Take action after days" = 0


truncsphere:
**Describe the bug**
I have a rule that deletes stale shows that haven't been watched in over 1 year. Currently 2 shows fit this description. I have watched the countdown reach 0 twice and neither time did it delete the show. Instead the logs show it's been removed from the collection and then it re-adds and restarts the timer. 

**Expected behavior**
To delete said shows after 30 days. 

**Screenshots**
![image](https://github.com/jorenn92/Maintainerr/assets/16292618/848c2c95-77f4-4304-98d1-8f4c37cc3750)

**Device (please complete the following information):**
 - Maintainerr version 2.0.4

**Other notes**
I have a rule to delete movies not watched after X years and it removes these with no issues.

adampetrovic:
Also have the same problem, running on both older versions and the most recent version. The logs aren't overly helpful.

jorenn92:
Hey, something else re-adds the media. Mostly this is a due to Overseerr’s auto watchlist download feature. Make sure something like that isn’t enabled

adampetrovic:
Overseerr works off the Plex Watchlist. None of these items are on that list.

![screenshot_2024_06_22_19_07_43](https://github.com/jorenn92/Maintainerr/assets/88030/3df06d82-bd3c-42e6-8911-c2d7288da286)

truncsphere:
I don't use watchlists with Overseer so the media is not being re-added. It never gets deleted.

truncsphere:
Any update on this issue?

jorenn92:
Hey, Sorry for the late reply. I haven't seen any related reports and haven't been able to replicate this myself. So i'm actually unsure what's going on. Do you happen to have logs from Plex, Sonarr, and perhaps Overseerr during the deletion timeframe?

I think it has something to do with a failure to fetch the tvdb id, which is used for removal through Sonarr. Altough it should log a message when this is the case. 

This error is due to a failure while fetching the tvdb ID from tmdb. But there are a few fallback mechanisms, so it's possible one of those did find the ID. 
`[maintainerr] | 22/06/2024 03:50:12  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 500`

truncsphere:
So this just happened to happen again this morning. Shows got removed by Maintainer, but are still on Plex. The shows were re-added and the counter started again at 30 days. 

[...]

Issue #1046, opened 2024-04-18
Handling collections failure causing collection handling to stall


HoustonDad:
**Describe the bug**
Collection handler erroring out 

**To Reproduce**
1. Create a TV Show rule, use the following export: 
```
mediaType: SHOWS
rules:
  - "0":
      - firstValue: Plex.sw_amountOfViews
        action: EQUALS
        customValue:
          type: number
          value: 0
```
Library: TV Shows
Media Type: Shows
Sonarr Action: Unmonitor and delete all seasons / episodes
Take action after days: 2
Keep logs for months: 6
3. Click Save
4. Run Rules
5. Wait a day or two for the "Take action after days" 
6. Collections -> Handle Collections
7. Watch logs

**Expected behavior**
The collection handler will run through the collection, until it errors out with the following, and does not restart / keep going, it seems to crash fully. I have to restart the docker container and click "Handle Collections" to restart it.

```
[maintainerr] | 18/04/2024 17:37:27  [DEBUG] [undefined] GET request failed: AxiosError: Request failed with status code 404
[maintainerr] | 18/04/2024 17:37:27  [ERROR] [ExceptionsHandler] Cannot read properties of undefined (reading 'ratingKey'
```

**Screenshots**
You can see in the screenshot below, the handling failed, did not continue, and I had to restart the container. 
![2Z5aEz499S](https://github.com/jorenn92/Maintainerr/assets/620445/e6787270-723c-4789-969a-a49c868c2760)

**Device (please complete the following information):**
 - OS: Ubuntu 18.04.6 LTS
 - Version:
 ```
ghcr.io/jorenn92/maintainerr    latest        4f5e7f571052   3 weeks ago     660MB
```

**Additional context**
The only workaround I've had is to restart the container.

**Comment by astrodad** [comment]
This is happening to me as well, but I get the same ratingKey error as soon as I click on "Handle Collections".

```
[Nest] 296  - 04/23/2024, 5:00:12 PM     LOG [CollectionWorkerService] Start handling all collections.
[Nest] 296  - 04/23/2024, 5:00:12 PM     LOG [CollectionWorkerService] Handling collection 'TV Shows Leaving Server Soon'
[Nest] 296  - 04/23/2024, 5:00:12 PM     LOG [CollectionsService] Removing media with id 3456 from collection..
[...]

PR #1547, opened 2025-01-31
fix: Deleting Show via Plex fallback

benscobie:
If the TVDB ID can't be found in Sonarr we attempt to delete media via Plex instead. This was not working for Shows as `plexData` is not populated, leading to an exception that crashes the collection worker. This was reported [here](https://github.com/jorenn92/Maintainerr/issues/1046#issuecomment-2571793133) and [here](https://github.com/jorenn92/Maintainerr/issues/1140#issuecomment-2626019502). Instead use `media.plexId` which is always available.

I've re-tested deleting media via Plex, and this works for shows, seasons and episodes.


ydkmlt84 (approved):
Looks good on my end.

PR #1561, opened 2025-02-06
feat: Replace plex-api 

benscobie:
This PR replaces [plex-api](https://www.npmjs.com/package/plex-api?activeTab=dependencies). This used [request](https://www.npmjs.com/package/request) under the hood which was deprecated 5 years ago and has numerous open issues around connection issues. I took the code we were using from plex-api and replaced request with the native fetch implementation as well as adding types.

I have also disabled keep alives as this resolved a socket hang up issue [for a user](https://discord.com/channels/1152219249549512724/1336762004228145172) and was the main motivation for looking into the plex-api library. There are numerous reported issues with node and keep alives resulting in socket hang ups and this was the easiest solution. Given most people will be communicating with Plex on LAN this should not be too much of an issue. Besides, all other APIs we talk to use axios which does not have keep alives enabled out of the box.

Lastly, I've wrapped the main rule executor body with a try-catch to prevent the job getting stuck when an unhandled exception is thrown.

### Testing
Validate that Plex API calls still work, so essentially create a new rule, run rules, change the rule slightly so items will get removed and then run rules again.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -39,7 +39,6 @@
     "lodash": "^4.17.21",
     "nest-winston": "^1.10.2",
     "node-cache": "^5.1.2",
-    "plex-api": "^5.3.2",
     "reflect-metadata": "^0.2.2",
     "rxjs": "^7.8.1",
     "sqlite3": "^5.1.6",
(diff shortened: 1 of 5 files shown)


ydkmlt84 (approved):
Tested! 👍

Issue #1432, opened 2024-12-14
Descriptions truncated in collections list


Rollbacke:
**Describe the bug**
The description of the collections, which should be complete, is truncated in the list of collections.

**To Reproduce**
Just go in collection and watch descriptions.

**Expected behavior**
Complete description.

**Screenshots**
https://github.com/user-attachments/assets/8b7d9616-71b1-4382-abbb-a7241da61f57

**Device (please complete the following information):**
 - Browser: Firefox
 - Maintainerr version: 2.2.1

Rollbacke:
It's purely cosmetic, it helps me remember which collection it is, but otherwise everything works fine.

benscobie:
The truncation is on purpose, but it shouldn't be scrollable like that.

benscobie:
This should be resolved in v2.5.0.

Rollbacke:
Hey @benscobie,
Unfortunately, the issue is not resolved. Still present on 2.9.0.

https://github.com/user-attachments/assets/f83e438d-de4e-41dc-b62e-1becdc0beac8

Rollbacke:
I didn't see that the "Rules" page had changed to look like the "Collections" page.
So I want to clarify that the problem is ONLY in the "Collections" page, not in the "Rules" page.

benscobie:
Previously the text was completely cut off after 2 lines which I fixed, the scrolling is intentional to prevent large cards.

Rollbacke:
I agree with you, and I think that's what should be done, but in the video, you can see the descriptions of the "En cours de suppression" collections (3) and (5) which are truncated, whereas there should be a scroll bar.
I don't explain why these and not the others...

Rollbacke:
Maybe it's a size of sentence related. The "En cours de suppression" collections (3) and (5) are a little bit smaller than the other with the scroll bar.

ydkmlt84:
> I agree with you, and I think that's what should be done, but in the video, you can see the descriptions of the "En cours de suppression" collections (3) and (5) which are truncated, whereas there should be a scroll bar. I don't explain why these and not the others...

Are you saying that the description on 3 and 5 has more content and should have a scrollbar like the others do?

ydkmlt84:
> -Maybe it's size of sentence related. The "En cours de suppression" collections (3) and (5) are a little bit smaller than the other with the scroll bar.- Tested: no.

[...]

PR #1563, opened 2025-02-09
chore: UI fixes/additions

ydkmlt84:
- added library name to media card on hover
- made transparent background on mediacard hover a little darker
- fixed MediaModal popup during removal of a global exclusionn
- fixed truncate/scroll issue with description on collection page
- other various small fixes/changes

### DIFF
--- a/ui/src/components/Collection/CollectionDetail/RemoveFromCollectionBtn/index.tsx
+++ b/ui/src/components/Collection/CollectionDetail/RemoveFromCollectionBtn/index.tsx
@@ -13,11 +13,12 @@ interface IRemoveFromCollectionBtn {
 }
 const RemoveFromCollectionBtn = (props: IRemoveFromCollectionBtn) => {
   const [sure, setSure] = useState<boolean>(false)
-  const [popup, setppopup] = useState<boolean>(false)
+  const [popup, setPopup] = useState<boolean>(false)
 
-  const handlePopup = () => {
+  const handlePopup = (e?: React.MouseEvent) => {
+    e?.stopPropagation()
     if (props.popup) {
-      setppopup(!popup)
+      setPopup(!popup)
     }
   }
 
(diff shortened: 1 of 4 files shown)


ydkmlt84:
Fixes #1432

ui/src/components/Collection/CollectionItem/index.tsx
```diff
@@ -20,7 +20,7 @@ const CollectionItem = (props: ICollectionItem) => {
           : {})}
       >
         {props.collection.media && props.collection.media.length > 1 ? (
-          <div className="z-1 absolute inset-0 flex flex-row overflow-hidden">
+          <div className="z-1 pointer-events-none absolute inset-0 flex flex-row overflow-hidden">
```
benscobie, line 23: This has a side effect of removing the pointer around the outside of  the box, but it's still clickable. You can just set the z-index to a negative value so the images go behind everything else, making it scrollable:
```suggestion
          <div className="z-[-100] absolute inset-0 flex flex-row overflow-hidden">
```

ui/src/components/Collection/CollectionDetail/RemoveFromCollectionBtn/index.tsx
```diff
@@ -71,12 +72,14 @@ const RemoveFromCollectionBtn = (props: IRemoveFromCollectionBtn) => {
       )}
 
       {popup ? (
-        <Modal title="Warning" onOk={handle} onCancel={handlePopup}>
-          <p>
-            This item is excluded <b>globally</b>. Removing this exclusion will
-            apply the change to all collections
-          </p>
-        </Modal>
+        <div onClick={(e) => e.stopPropagation()}>
```
benscobie, line 75: This wrapping div is not necessary. Update handle() to call e.stopPropagation() like handlePopup() does.

benscobie (changes requested):

ui/src/components/Collection/CollectionItem/index.tsx
ydkmlt84: Again, I am not sure how you see those things. Thanks!

ydkmlt84:
Will look at later.

PR #1573, opened 2025-02-10
ci: Group nestjs and remove PR limit on dependabot

benscobie:
The `@nestjs` packages have been slowly moving to v11 over the last few weeks. The open PR limit stopped them from all being updated at the same time, so I've removed that and introduced a group for them.


ydkmlt84 (approved):
👍

PR #1587, opened 2025-02-18
chore: move from self hosted community rules

ydkmlt84:
This PR removes the need for me to self host the community rules, uses a free service (for now) provided by Cloudflare KV, and removes the need to include an API key into the committed code. Adds some automated community rules cleanup as well. There is also a rate limit of 5 per minute to POST or PATCH (new rules or karma) requests to the worker, which will time out for 60 seconds, to avoid spamming.

- Moved the community rules into a Cloudflare Worker/KV flow to remove self-hosted requirement
- Fixed the poster image scroll issue in collection description
- Added an optional "Uploaded by" box to the rule upload
- Added an uploaded by column to the community rule display
- Added uploaded by as search criteria to community rules
- Added a hasRule tag to uploaded rules
- Added a rule scrubber to CF worker that removes uploaded rules with no actual rules attached. (using a hasRules: false tag when empty)
- Added a rule scrubber that removes community rules with less than -100 karma. The automation runs at midnight UTC every day.

### DIFF
--- a/ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
+++ b/ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
@@ -14,6 +14,7 @@ interface ICommunityRuleUpload {
 const CommunityRuleUpload = (props: ICommunityRuleUpload) => {
   const nameRef = useRef<any>()
   const descriptionRef = useRef<any>()
+  const uploadedByRef = useRef<any>()
   const [thanksModal, setThanksModal] = useState<boolean>(false)
   const [failed, setFailed] = useState<boolean>(false)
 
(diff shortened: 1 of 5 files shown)


ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
```diff
@@ -24,6 +25,7 @@ const CommunityRuleUpload = (props: ICommunityRuleUpload) => {
         type: props.type,
         description: descriptionRef.current.value,
         JsonRules: props.rules,
+        uploadedBy: uploadedByRef.current.value || undefined,
       })
         .then((resp) => {
           if (resp.code === 1) {
```
benscobie, line 31: You could save `resp.result` here and use that instead of the generic message when an upload fails; _Something went wrong uploading your rules. Please try again later_

benscobie (approved):

benscobie:
Remember to squash your commits/use squash & merge

Issue #1590, opened 2025-02-20
Add buttons have an invalid state after deleting a rule


Flaminel:
**Describe the bug**
`Add` and `New Section` buttons appear even though the first rule is not yet configured. Also, the same buttons never appear again if a rule is deleted while the latest added one is not configured yet (see steps to reproduce).

**To Reproduce**
1. Go to 'Rules'
2. Click on 'New rule'
3. Scroll down to 'Rules'
4. Configure Rule#1.
5. Press the 'Add' button.
6. Configure Rule#2.
7. Press the 'Add' button.
8. Press the 'Delete' button on Rule#2.
9. Configure Rule#2 (which was Rule#3).

**Expected behavior**
The `Add` and `New Section` buttons do not have an invalid state and they appear after a rule has been configured.

**Screenshots**
![Image](https://github.com/user-attachments/assets/c4a77b34-b59f-496f-8eb2-082a0ce31ce4)

**Device (please complete the following information):**
 - OS: Docker on Debian.
 - Version: 2.9.0

Flaminel:
Fixed by #1591.

PR #1591, opened 2025-02-20
fix: prevented invalid state for add buttons

Flaminel:
Relates to #1590

### DIFF
--- a/ui/src/components/Rules/Rule/RuleCreator/index.tsx
+++ b/ui/src/components/Rules/Rule/RuleCreator/index.tsx
@@ -129,7 +129,11 @@ const RuleCreator = (props: iRuleCreator) => {
       props.onUpdate(rulesCreated.current.map((el) => el.rule))
     }
 
-    added.current = added.current.filter((e) => e !== id)
+    added.current = added.current
+      .filter((e) => e !== id)
+      .map((e) => {
+        return e = e > id ? e - 1 : e
+      })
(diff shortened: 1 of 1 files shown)


ui/src/components/Rules/Rule/RuleCreator/index.tsx
```diff
@@ -96,7 +96,7 @@ const RuleCreator = (props: iRuleCreator) => {
   )
   const rulesCreated = useRef<IRulesToCreate[]>([])
   const deleted = useRef<number>(0)
-  const added = useRef<number[]>([])
+  const added = useRef<number[]>([1])
```
benscobie, line 99: This breaks the first rule when editing an existing rule. I removed it and couldn't see any breakages, what was this required for?

benscobie (changes requested):

ui/src/components/Rules/Rule/RuleCreator/index.tsx
Flaminel: Without this part, the user could add a new rule or section before configuring the first rule.
image

> This breaks the first rule when editing an existing rule.

Yeah, that makes sense. I should probably set it to 1 only if it's a new rule.

Flaminel: I just pushed a fix for that.

benscobie (approved):
Good work, thank you again!

Issue #1470, opened 2024-12-27
Cannot delete Section #1 Rule #1


chrgraham:
**Describe the bug**
The UI lacks a Delete button for `Section #1` `Rule #1`

**To Reproduce**
1. Open any Rule
2. Observe the first `Delete` button is for `Rule #2`

**Expected behavior**
A `Delete` button is available for every rule, and the app successfully modifies the YAML when the first rule is deleted.

**Screenshots**
<img width="1214" alt="image" src="https://github.com/user-attachments/assets/e4c5e29b-40bd-478b-a273-23c874df637d" />

**Device (please complete the following information):**
 - OS: Docker
 - Version: 2.4.0

**Additional context**
A viable workaround is to `Export`, then manually modify the YAML, then `Import` and past the modified YAML. It is important to  make sure that:
- Rules array ordering starts at 0 and sequentially numbered
- `Rule 0` does not have an `operator`
- You did not otherwise bungle the YAML formatting

benscobie:
This is in [v2.11.0](https://github.com/jorenn92/Maintainerr/releases/tag/v2.11.0)

PR #1592, opened 2025-02-20
feat: allow Section 1 Rule 1 deletion if other rules are present

Flaminel:
Depends on #1591 
Relates to #1470

### DIFF
--- a/ui/src/components/Rules/Rule/RuleCreator/index.tsx
+++ b/ui/src/components/Rules/Rule/RuleCreator/index.tsx
@@ -121,16 +121,13 @@ const RuleCreator = (props: iRuleCreator) => {
   const ruleDeleted = (section = 0, id: number) => {
     if (rulesCreated.current.length > 0) {
       let rules = rulesCreated.current?.filter((el) => el.id !== id)
+      const section1IsEmpty = !rules.some((r) => r.rule.section === 0)
+
       rules = rules.map((e) => {
         e.id = e.id > id ? e.id - 1 : e.id
 
-        if (
-          section === 1 &&
-          rules.filter((r) => r.rule.section === 0).length === 0
-        ) {
-          if (e.rule.section && e.rule.section > 0) {
-            e.rule.section -= 1
-          }
+        if (section1IsEmpty && section === 1 && e.rule.section) {
+          e.rule.section -= 1
         }
 
         return e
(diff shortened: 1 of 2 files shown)


ui/src/components/Rules/Rule/RuleCreator/index.tsx
```diff
@@ -123,6 +123,16 @@ const RuleCreator = (props: iRuleCreator) => {
...
         e.id = e.id > id ? e.id - 1 : e.id
+
+        if (
+          section === 1 &&
+          rules.filter((r) => r.rule.section === 0).length === 0
+        ) {
+          if (e.rule.section && e.rule.section > 0) {
+            e.rule.section -= 1
+          }
+        }
```
benscobie, line 134: You could move the filter check outside of the loop as it's always the same. You also don't need the `> 0` check as `0 == false`.

```suggestion
        const section1IsEmpty = !rules.some((r) => r.rule.section === 0)
        // ... rules.map
        if (
          section1IsEmpty &&
          section === 1 &&
          e.rule.section
        ) {
            e.rule.section -= 1
        }
```

ui/src/components/Rules/Rule/RuleCreator/index.tsx
```diff
@@ -132,19 +142,24 @@ const RuleCreator = (props: iRuleCreator) => {
...
+
     const rules = [...ruleAmount[1]]
     rules[section - 1] = rules[section - 1] - 1
 
-    if (rulesCreated.current.length > 0) {
-      setEditData({ rules: rulesCreated.current.map((el) => el.rule) })
+    // If the section becomes empty, remove it from the list
+    if (rules[section - 1] <= 0) {
+      rules.splice(section - 1, 1)
     }
```
benscobie, line 155: I don't think this is needed as `nonEmptySections` is doing the same thing?

ui/src/components/Rules/Rule/RuleCreator/index.tsx
Flaminel: Thank you for the tips! I've fixed them.

ui/src/components/Rules/Rule/RuleCreator/index.tsx
Flaminel: I was under the impression that I have to also update that collection. Nice catch!

benscobie (approved):
Thank you

PR #1593, opened 2025-02-21
fix: removed appVersion filtering of community rules.

ydkmlt84:
This was causing an issue with some community rules not showing because we were filtering rules based on appVersion. Ensuring that the uploaded rules' appVersion was equal to or lower than the current version of the app. Shouldn't that always be the case? I am not sure why we were doing that in the first place.

This really only became an issue with the comparison to version 2.10.0, and the way we were comparing the versions.

### DIFF
--- a/ui/src/components/Common/CommunityRuleModal/index.tsx
+++ b/ui/src/components/Common/CommunityRuleModal/index.tsx
@@ -54,11 +54,7 @@ const CommunityRuleModal = (props: ICommunityRuleModal) => {
     GetApiHandler('/rules/community/').then((resp: ICommunityRule[]) => {
       if (resp) {
         if (!('code' in resp)) {
-          resp = resp.filter(
-            (e) =>
-              e.appVersion!.replaceAll('.', '') <=
-                appVersion.current.replaceAll('.', '') && e.type === props.type,
-          )
+          resp = resp.filter((e) => e.type === props.type)
           resp = resp.sort((a, b) => b.karma! - a.karma!)
           setCommunityRules(resp)
           setoriginalRules(resp)


benscobie:
> Ensuring that the uploaded rules' appVersion was equal to or lower than the current version of the app. Shouldn't that always be the case? I am not sure why we were doing that in the first place.

Not everybody is going to be running the latest version though so it's not always the case. I suspect the check is in place as if the rules format changes or a new rule gets added, then an upload from that newer version likely won't work on an older version.

ydkmlt84:
Correct me if I'm wrong but it's checking that the uploaded rule version is lower than or equal to the current version? Which would always be the case.

benscobie:
> Or are you saying they're running version 2.5, we would want to hide rules higher than that. If you think we should do that, I'll change the filtering type as the current way doesn't work properly.

Yes exactly. What's wrong with the filtering, it looks correct to me?

ydkmlt84:
> > Or are you saying they're running version 2.5, we would want to hide rules higher than that. If you think we should do that, I'll change the filtering type as the current way doesn't work properly.
> 
> Yes exactly. What's wrong with the filtering though as it looks correct and follows that ^ already.

The way it applies the filter, lexically (according to chatgpt and shown in the results of the available community rules), 2.9.0 is larger than 2.10.0. So even on the latest version a lot of rules are being filtered out that shouldn't be. 

image

ui/src/components/Common/CommunityRuleModal/index.tsx
```diff
@@ -54,11 +54,18 @@ const CommunityRuleModal = (props: ICommunityRuleModal) => {
...
+              !e.appVersion ||
+              e.appVersion.localeCompare(
+                appVersion.current || '0.0.0',
+                undefined,
+                { numeric: true },
+              ) <= 0
+            const typeCheck = e.type === props.type
+
+            return versionCheck && typeCheck
+          })
```
benscobie, line 68:
```suggestion
          resp = resp.filter(
            (e) =>
              parseInt(e.appVersion!.replaceAll('.', '')) <=
                parseInt(appVersion.current.replaceAll('.', '')) && e.type === props.type,
          )
```

benscobie: Actually that doesn't work either because the version parts have variable lengths 🤦

ydkmlt84: Yeah, I think that was the issue before, more or less. Breaking down 2.9.0 into 290 and comparing it to 2.10.0 (2100). Mathematically 2100 is a larger number than 290, so I don't see why a 2.9.0 rule would get filtered out. Apparently the lexical version stops when it compares the second integer and 9 is larger than 1...so it returns that 290 is higher than 2100.

ydkmlt84: > Actually that doesn't work either because the version parts have variable lengths 🤦 You would have to split the version into the 3 parts and compare them individually I think.

What is wrong, from your eyes, with using localecompare?

benscobie: It's fine if we're sticking with 3 parts. It won't work with other semvar things like tags if we ever did that in the future, but nor would the existing one so this is good for the meantime.

ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
```diff
@@ -16,7 +16,7 @@ const CommunityRuleUpload = (props: ICommunityRuleUpload) => {
   const descriptionRef = useRef<any>()
   const uploadedByRef = useRef<any>()
   const [thanksModal, setThanksModal] = useState<boolean>(false)
-  const [failed, setFailed] = useState<boolean>(false)
+  const [failed, setFailed] = useState<string | false>(false)
```
benscobie, line 19: You can just set the default to an empty string

```suggestion
  const [failed, setFailed] = useState<string>('')
```

ui/src/components/Common/CommunityRuleModal/index.tsx
ydkmlt84: Yeah, that was the other issue that I couldn't work around with localecompare. If there was say a 2.15.0-LTS or something to that effect, it wouldn't work. Is there no way to just compare if a number is larger or smaller than another number? With our current method wouldn't the number always get larger with each version, even after removing the dots?

I guess we could revisit this if we ever decided to do something different with the version numbers.

ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
ydkmlt84: Done

ui/src/components/Common/CommunityRuleModal/index.tsx
benscobie: We can just use a package that handles semver comparisons:
https://www.npmjs.com/package/compare-versions

ydkmlt84: > We can just use a package that handles semver comparisons: https://www.npmjs.com/package/compare-versions

Let me give that a go. Seems easy enough.

benscobie (approved):

Issue #1599, opened 2025-02-25
Rules with a "take action after days" of 0 still deletes movies despite what docs say


andrew-kennedy:
**Describe the bug**
When "Take action after days" is set to 0 within a rule, movies are deleted every time the collections are processed. Based on 

<img width="782" alt="Image" src="https://github.com/user-attachments/assets/667af60b-31b2-4ba0-a527-aa62fd63fb8d" />

I was under the impression that 0 would lead to "nothing being removed" (which is what I want), and would simply maintain a collection according to the rules that wouldn't be deleted.

Is there no way to have collections "maintained" but not deleted? I want to have collections within plex for all requested movies according to the users who requested them, without deleting movies in those collections. I have other "leaving soon" collections where I intend to keep deletion enabled.

Flaminel:
Honestly a value of `0` is something I often use as sometimes I just need a quick cleanup on something to make room for another movie. I wouldn't actually want this "feature" to go away, but rather maybe update the docs and add a `-1` value for what you're describing?

ydkmlt84:
0 items should not be getting deleted automatically, from my testing. I'll have to look into this again.

andrew-kennedy:
If you look here:

https://github.com/jorenn92/Maintainerr/blob/b49e34fdacf2a511b1dc7db0b184a0bea78abb01/server/src/modules/collections/collection-worker.service.ts#L84-L94

It seems when the value is 0, the dangerDate will always be "now", and the added date will always be before "now", so it'll always be deleted.

andrew-kennedy:
Opened a PR to address this issue as I took a quick glance at the code and it seemed like a super easy fix. Even preserved the 0 behavior for you @Flaminel

benscobie:
> Opened a PR to address this issue as I took a quick glance at the code and it seemed like a super easy fix. Even preserved the 0 behavior for you [@Flaminel](https://github.com/Flaminel)

This is in [v.2.11.0](https://github.com/jorenn92/Maintainerr/releases/tag/v2.11.0)

PR #1600, opened 2025-02-26
Add support for a new Do Nothing *arr (and Plex) action.

andrew-kennedy:
This PR addresses https://github.com/jorenn92/Maintainerr/issues/1599 by introducing a new `ServarrAction`, `ServarrAction.DO_NOTHING` which disables the `Take action after days` form section on the rule creation screen, and also prevents collections with this action from ever taking action on the media in them. 

This is useful for dynamic, rules-managed collections that you may want around for browsing within plex.

<details>
<summary>
Previous PR Description
</summary>

This PR addresses https://github.com/jorenn92/Maintainerr/issues/1599 by introducing a switch statement that handles negative `deleteAfterDays` values in collections separately from zero or positive values, which are already handled "correctly". It allows for collections to be maintained according to whatever rules, without ever having media deleted from sonarr/radarr/plex.

This also modifies the OverviewContent component to pass undefined as the daysLeft to the MediaCard when there is a negative value for a collection's `deleteAfterDays` value, which makes those media cards not display a countdown number in the upper right for collections with deletion disabled.
<details>


benscobie (changes requested):
Hey, thank you for the PR. I think this could be implemented in a more user-friendly way by adding an extra option 'Do nothing' to the action list. It's a bit more work though:
- Add a new value to the `ServarrAction` - DO_NOTHING
- Check `arrAction == ServarrAction.DO_NOTHING`
instead of your new `case -1` and `< 0` checks
-  Update `ui\src\components\Rules\RuleGroup\AddModal\index.tsx`
    - Add the new option to ArrAction component usages
    - Hide the Take action after days input if arrAction == do_nothing
    - I'd also set deleteAfterDays to undefined when selecting the option to keep things tidy. It seems it's already nullable on the backend, but we seem to be assuming it's never null in some places so we just need to double check we won't hit those statements.

andrew-kennedy:
Updated the PR to support a new *arr action. I audited all places deleteAfterDays appears and should handle undefined fine in all of them, though I didn't do extensive testing.

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
```diff
@@ -345,7 +345,7 @@ const AddModal = (props: AddModal) => {
         collection: {
           visibleOnRecommended: showRecommended,
           visibleOnHome: showHome,
-          deleteAfterDays: +deleteAfterRef.current.value,
+          deleteAfterDays: (arrOption === undefined || arrOption === 4) ? undefined : +deleteAfterRef.current.value,
```
ben-propflo, line 348: I changed an existing rule from Delete to Do Nothing and got a JS error on save. The deleteAfterRef.current.value condition can be removed for this block as it will be undefined in some cases now.

ui/src/components/Collection/CollectionItem/index.tsx
```diff
@@ -11,6 +11,13 @@ interface ICollectionItem {
 const CollectionItem = (props: ICollectionItem) => {
   const LibrariesCtx = useContext(LibrariesContext)
 
+    const renderDeleteAfterText = () => {
+      if (props.collection.deleteAfterDays === undefined) {
+        return "Never"
+      }
+      return `After ${props.collection.deleteAfterDays} days`
+    }
+
```
benscobie, line 20: I'd personally just keep all of this inline in the template as it's not very complicated.

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
benscobie, line 348: I changed an existing rule from Delete to Do nothing and got a JS error on save. The deleteAfterRef.current.value condition can be removed for this block as it will be undefined in some cases now.

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -73,25 +73,31 @@ export class CollectionWorkerService extends TaskBase {
...
       for (const collection of collections) {
-        this.infoLogger(`Handling collection '${collection.title}'`);
 
-        const collectionMedia = await this.collectionMediaRepo.find({
-          where: {
-            collectionId: collection.id,
-          },
-        });
+        if (collection.arrAction === ServarrAction.DO_NOTHING) {
+          this.infoLogger(`Skipping collection '${collection.title}' as it's action is 'Do Nothing'`);
```
benscobie, line 78:
```suggestion
          this.infoLogger(`Skipping collection '${collection.title}' as its action is 'Do Nothing'`);
```

server/src/modules/collections/collection-worker.service.ts
```diff
@@ -73,25 +73,31 @@ export class CollectionWorkerService extends TaskBase {
...
+              new Date().getTime() - +collection.deleteAfterDays * 86400000,
+            );
+
+            for (const media of collectionMedia) {
+              // handle media addate <= due date
+              if (new Date(media.addDate) <= dangerDate) {
+                await this.handleMedia(collection, media);
+                handledCollections++;
+              }
+            }
```
benscobie, line 100: This tab formatting is incorrect. You can run `yarn format` to correct it.

benscobie (changes requested):
setRules & updateRules in rules.service.ts need updating as they're currently casting deleteAfterDays to a number 💥

ui/src/components/Collection/CollectionItem/index.tsx
```diff
@@ -89,7 +89,11 @@ const CollectionItem = (props: ICollectionItem) => {
 
             <div className="mr-0 sm:mr-0">
               <p className="font-bold">Delete</p>
-              <p className="text-amber-500">{` After ${props.collection.deleteAfterDays} days`}</p>
+              <p className="text-amber-500">
+                {props.collection.deleteAfterDays === undefined
```
benscobie, line 93:
```suggestion
                {props.collection.deleteAfterDays == null
```

ui/src/components/Overview/Content/index.tsx
```diff
@@ -113,18 +113,21 @@ const OverviewContent = (props: IOverviewContent) => {
         (colEl) => colEl.plexId === +plexId,
       )
       if (collectionData && collectionData.collection) {
+        if (collectionData.collection.deleteAfterDays === undefined) {
```
benscobie, line 116:
```suggestion
        if (collectionData.collection.deleteAfterDays == null) {
```

benscobie (changes requested):
Thank you for the changes 😃 These are the last two as I've now managed to save a rule. The types used in the frontend aren't the best at the moment, a lot are missing ` | null`.

andrew-kennedy:
Thanks for the testing! I don't have a good test setup due to the nature of how my plex server is setup. I'd have to deploy a test build container of maintainerr inside my podman network to get it working properly which is quite a pain for me.

benscobie (approved):
Thank you for your contribution 😄

PR #1604, opened 2025-03-01
feat: Add IMDB, RT and TMDB Plex rating filters

benscobie:
The rating values can be overwritten by tools like Kometa, which will allow users to filter by a different source such as Trakt (though they still retain the IMDB/RT/TMDB images in Plex & our code).

I also spotted a logic error when building the second value list. If the library was a movie type, rules that had a MediaType of BOTH and a showType would not be displayed in the list, this is now fixed. This was fixed solely in first value [by this commit](https://github.com/jorenn92/Maintainerr/commit/5f589e69d0fa3fb266da59a904c7b2cc6f6734f1).


ydkmlt84 (approved):
Tested 👍

PR #1606, opened 2025-03-02
feat: Add Jellyseerr support

benscobie:
This is a complete copy of our existing Overseerr code as the APIs we use are functionally the same at the moment. The only feature I have not brought across is the 'Force reset Overseerr record' option, as availability sync should be used instead. If we have problems with this we can always introduce the option again.

Feature request: https://features.maintainerr.info/posts/69/jellyseerr

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ It's a one-stop-shop for handling those outlying shows and movies that take up p
 
 # Features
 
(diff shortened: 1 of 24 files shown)


ydkmlt84 (approved):
Tested 👍

PR #1607, opened 2025-03-02
feat: added authentication

ydkmlt84:
Lots to discuss but the gist is:

- UI access requires login with a username and password, verified by JWT token cookie
- outside API access requires either sessionToken cookie (logged in) or API key
- turn auth requirement on and off in settings as well as set username and password
- added auth_settings table to DB


ui/src/middleware.ts
```diff
@@ -1,16 +1,87 @@
...
-export function middleware(request: NextRequest) {
-  if (request.nextUrl.pathname.startsWith('/api/')) {
-    const url = request.nextUrl.clone()
-    const destination = new URL(`http://localhost:${apiPort}`)
-    url.host = destination.host
-    url.port = destination.port
-    url.protocol = destination.protocol
-    url.basePath = ''
-    return NextResponse.rewrite(url)
+export async function middleware(request: NextRequest) {
```
benscobie, line 5: There isn't much need for this middleware. You can add an interceptor to axios so when calling any of the ApiHandlers, any 401 response would redirect the user to the login page.

ui/src/middleware.ts
```diff
@@ -1,16 +1,87 @@
...
+            path: '/',
+            maxAge: 60 * 60, // 1 hour in seconds
+          })
+          return response
+        } else {
+          console.warn('Middleware - Token refresh failed.')
+        }
+      } catch (error) {
+        console.error('Middleware - Refresh request failed:', error)
+      }
```
benscobie, line 79: Refreshing the token on every request is unnecessary. The client should be aware of the expiration time, so it can simply be refreshed on a timer. You could do this in the AuthProvider.

ydkmlt84: this would only happen once, if the sessiontoken wasnt present (expired). The next requests would have the sessiontoken, because of the refreshtoken function. That was what I was attempting to do anyways.

ui/src/contexts/auth-context.tsx
```diff
@@ -0,0 +1,55 @@
...
+}) => {
+  const [authEnabled, setAuthEnabled] = useState<boolean | null>(null)
+  const [isAuthenticated, setIsAuthenticated] = useState<boolean | null>(null)
+
+  const fetchAuthStatus = async () => {
+    try {
+      const statusRes = await fetch('/api/authentication/status', {
+        credentials: 'include',
+      })
+      const authEnabledHeader = statusRes.headers.get('X-Auth-Enabled')
```
benscobie, line 26: I can't see a need for this X-Auth-Enabled header stuff. `/api/authentication/status` could simply require authentication, and if a 401 is returned then the provided credentials (if any) are incorrect. The user can then be redirected to the login page. 

You don't even need to do that to be honest if you just do this: https://github.com/jorenn92/Maintainerr/pull/1607/files#r1976728680

The UI will start loading in and we'll attempt to make API calls, and then the user will be promptly redirected to the login page.

server/src/authentication.middleware.ts
```diff
@@ -0,0 +1,70 @@
...
+import { Request, Response, NextFunction } from 'express';
+import { AuthenticationService } from './modules/authentication/authentication.service';
+import * as jwt from 'jsonwebtoken';
+
+interface AuthenticatedRequest extends Request {
+  user?: any;
+}
+
+@Injectable()
+export class AuthenticationMiddleware implements NestMiddleware {
```
benscobie, line 15: As mentioned on Discord, this should be replaceable with guards & strategies in nestjs/passport: https://docs.nestjs.com/recipes/passport#implement-protected-route-and-jwt-strategy-guards

server/src/modules/authentication/entities/authentication_settings.entities.ts
```diff
@@ -0,0 +1,22 @@
...
+  authEnabled: boolean;
+
+  @Column({ type: 'text', nullable: true })
+  username: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  passwordHash: string | null;
+
+  @Column({ type: 'text', nullable: true })
+  jwt_secret: string;
```
benscobie, line 18:
```suggestion
  jwtSecret: string;
```

ui/src/components/Settings/Main/index.tsx
```diff
@@ -1,65 +1,188 @@
...
+  const [hasChanges, setHasChanges] = useState(false)
+  const [apiKey, setApiKey] = useState<string>('')
+  const [originalApiKey, setOriginalApiKey] = useState<string>('')
+  const [showApiKey, setShowApiKey] = useState(false)
+
+  const apiPort = process.env.API_PORT || 3001
 
   useEffect(() => {
     document.title = 'Maintainerr - Settings - General'
+    fetch(`http://localhost:${apiPort}/api/authentication/settings`, {
```
benscobie, line 39: Is there a reason you're calling the API directly instead of going via the UI /api/ proxy route?

server/package.json
```diff
@@ -34,8 +34,11 @@
     "@nestjs/schedule": "^5.0.1",
     "@nestjs/typeorm": "^11.0.0",
     "axios": "^1.7.9",
+    "bcrypt": "^5.1.1",
```
benscobie, line 37: We should use Argon2id instead, see: https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

server/src/database/migrations/1740877643707-add_auth_table.ts
```diff
@@ -0,0 +1,409 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddAuthTable1740877643707 implements MigrationInterface {
+  name = 'AddAuthTable1740877643707';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
```
benscobie, line 6: You'll need to regenerate this when fixing the casing of jwt_secret, but typeorm has generated a lot of extra stuff that isn't required. You only need the create table for auth_settings.

ui/src/components/Layout/NavBar/index.tsx
```diff
@@ -54,12 +56,26 @@ let navBarItems: NavBarLink[] = [
...
+    try {
+      await fetch('/api/authentication/logout', {
+        method: 'POST',
+        credentials: 'include',
+      })
+      window.location.href = '/login'
+    } catch (error) {
+      console.error('Failed to log out:', error)
+    }
+  }
```
benscobie, line 78: Given that logout is just removing 2 cookies, this can all be done client side in JavaScript, I think you've got this somewhere already.

server/src/database/migrations/1740877643707-add_auth_table.ts
ydkmlt84: That was from a yarn migration:generate run. I didn't do anything different after it ran. I saw that too but naively assumed it knew what it was doing.

ui/src/middleware.ts
benscobie: Oh yes I missed the if statement, might be a bit wonky when there are multiple API requests going off, there will be a bunch of cookies being set. The typical refresh token flow is what I described above.

server/src/modules/authentication/authentication.service.ts
```diff
@@ -0,0 +1,196 @@
...
+
+    // Generating AccessToken
+    const accessToken = jwt.sign({ username }, secret, { expiresIn: '1h' });
+
+    // Generating RefreshToken
+    const refreshToken = jwt.sign({ username }, secret, { expiresIn: '30d' });
+
+    res.cookie('sessionToken', accessToken, {
+      httpOnly: true,
+      secure: process.env.NODE_ENV === 'production',
```
benscobie, line 95: This won't work for people that aren't using HTTPS. Arguably somebody hosting over HTTP shouldn't need auth as this is likely a local server, but you never know.

ydkmlt84: https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#:~:text=A%20cookie%20with%20the%20HttpOnly,site%20scripting%20(XSS)%20attacks.

benscobie: I wonder what they mean by `It's never sent with unsecured HTTP **(except on localhost)**,`. If they literally only mean http://localhost then that's not enough.

server/src/modules/authentication/authentication.service.ts
```diff
@@ -0,0 +1,196 @@
...
+      sameSite: 'lax',
+      path: '/',
+      maxAge: 60 * 60, // 1h in seconds
+    });
+
+    res.cookie('refreshToken', refreshToken, {
+      httpOnly: true,
+      secure: process.env.NODE_ENV === 'production',
+      sameSite: 'lax',
+      path: '/',
```
ben-propflo, line 105: The path should be set to the refresh token endpoint, taking into account any subfolder config. Currently you're sending the refresh token along with the access token which mostly defeats the point of the refresh token. This change will break the middleware refresh you do so you'll need to move the refresh to the client as I suggested.

server/src/modules/authentication/entities/authentication_settings.entities.ts
ydkmlt84: Done

ui/src/components/Settings/Main/index.tsx
ydkmlt84: Do you mean instead of like this?

fetch(`/api/authentication/settings`,

benscobie: Yes, though we should use the handlers like GetApiHandler as they automatically handle any base path

server/src/modules/authentication/authentication.service.ts
benscobie, line 105: The path should be set to the refresh token endpoint, taking into account any subfolder config. Currently you're sending the refresh token along with the access token which mostly defeats the point of the refresh token. This change will break the middleware refresh you do so you'll need to move the refresh to the client as I suggested.

server/package.json
ydkmlt84: Done

ydkmlt84:
I am tabling this whole idea for now. I am now leaning more towards auth being something we never implement.

PR #1608, opened 2025-03-02
fix: Sonarr returns monitored=true if non-existent

benscobie:
If the /series?tvdbId=$ API call returns nothing we would use a different endpoint as a fallback. This endpoint incorrectly returns monitored: true for whatever reason. This issue becomes apparent when you have two Sonarr instances feeding a single Plex library and try to use the monitored condition.


ydkmlt84 (approved):

ydkmlt84 (approved):
+1

ydkmlt84 (approved):
Looks good on my end.

PR #1618, opened 2025-03-03
fix: Test media breaking a running rule executor

### LINKED ISSUE #1610: Testing media whilst rules running can cause breakages
Test media & the rule job both use the RuleComparatorService which contains state for the rule execution (executeRulesWithData). It's currently injected as a singleton so it's possible for that state to be corrupted.

Example error log:
`[maintainerr] | 02/03/2025 19:23:31  [DEBUG] [RuleComparatorService] Cannot read properties of undefined (reading 'sectionResults')`

There's a few possible solutions here:
- Change the injection scope to request or transient which would require scope changes to some other services such as RulesService
- Move state into the executeRulesWithData function and pass it around to the private functions
- Create a factory for the RuleComparatorService

To replicate:
1. Open a new tab and get Test media ready to go
2. Run rules - it helps if they take a bit of time to process to give you enough time to get back to test media.
3. Smash test media and you should see errors in the logs

benscobie:
Resolves #1610 by using a new RuleComparatorService in each call site as it is not "thread" safe.


ydkmlt84:
Receiving these errors when trying to run locally with `yarn dev`. I believe it is because of the `/api/` overwrite from the middleware, and adding that extra slash. 
image

benscobie:
> Receiving these errors when trying to run locally with `yarn dev`. I believe it is because of the `/api/` overwrite from the middleware, and adding that extra slash. image

The server has likely crashed, check the logs

ydkmlt84:
> > Receiving these errors when trying to run locally with `yarn dev`. I believe it is because of the `/api/` overwrite from the middleware, and adding that extra slash.
> 
> The server has likely crashed, check the logs.
> 
> EDIT: Or you're accessing the UI straight away and the server hasn't finished booting yet.

I am dumb and I am not going to tell you what the problem was :)

ydkmlt84 (approved):
Tested. 👍

ydkmlt84 (approved):
Tested 👍

PR #1619, opened 2025-03-04
fix: Overseerr/Jellyseerr invalid API key passing test

### LINKED ISSUE #1605: Overseerr test succeeds with invalid API key
The endpoint we use to test the Overseerr connection does not require auth. We should find another endpoint to use that requires auth for a proper test.

benscobie:
We now use the `/settings/about` endpoint to validate that the API key is correct. I've also improved the error messages by indicating what the problem likely is.

Resolves #1605


server/src/modules/api/overseerr-api/overseerr-api.service.ts
```diff
@@ -350,4 +352,46 @@ export class OverseerrApiService {
...
+      this.logger.log("Couldn't validate Overseerr API connectivity!", {
+        label: 'Overseerr API',
+        errorMessage: e.message,
+      });
+      this.logger.debug(e);
+
+      return {
+        status: 'NOK',
+        code: 0,
+        message: 'Unknown error',
```
ydkmlt84, line 393: Tested the incorrect API and received the Alert on screen that the API key was invalid. 👍

However, with an incorrect port, all it showed was `Unknown error`. Could we provide some more details here possibly? The log entry was somewhat helpful, but not the Alert on screen. Even the log entry though could say something about the test being unsuccessful? `Unable to validate connectivity` doesn't really lead to a connection issue, in my opinion. It could also be my "western" English.

benscobie: Fair point, I've included the actual exception message now and updated the log wording and format.

ydkmlt84 (approved):
👍

PR #1621, opened 2025-03-07
feat: Improve Tautulli settings UX

### LINKED ISSUE #1390: Tautulli connection fails with leading slash in base url

benscobie:
This PR improves Tautulli settings configuration by:
- Merging Hostname, Port and Base URL inputs into one URL field. Users can simply copy and paste the URL from their browser.
  - I have added a migration to strip any slashes as the base URL input wasn't being stripped when constructing the URL
  - I have also ensured they're stripped for the older style settings
- Implementing test before save like the *arr settings
- Sharing client & server validation
- Allowing saving of empty inputs (deleting the setting)
- Returning more detailed error messages to the user. We forward any 400 error message we get back from Tautulli such as incorrect API key. We return the status codes in other situations, and lastly we return the exception message, which captures things like connection refused.

I have also fixed eslint not working in VSC, and have upgraded it and fixed the error in #1671 

Fixes #1390


ydkmlt84:
Cannot save after a successful test
image

ydkmlt84:
It is not super clear what needs to be put into the URL field. I know what needs to be put there but maybe a sublabel under the URL label? The placeholder in the field helps a little bit, but not if you are using IP and port.

Copying the URL from the browser is nice except if you copy the whole thing, it has a subfolder at the end of it. `http://192.168.1.91:8181/home` is my URL, and this doesn't work in the URL field. I get a type error of `undefined (reading 'message')`.

All functions work as intended. Can test without saving, clear settings, save changes once a successful test is performed. Just that one nitpick/suggestion.

benscobie:
> It is not super clear what needs to be put into the URL field. I know what needs to be put there but maybe a sublabel under the URL label? The placeholder in the field helps a little bit, but not if you are using IP and port.
> 
> Copying the URL from the browser is nice except if you copy the whole thing, it has a subfolder at the end of it. `http://192.168.1.91:8181/home` is my URL, and this doesn't work in the URL field. I get a type error of `undefined (reading 'message')`.

Ah yes fair point. I will add a few examples underneath the input and have a look at that error.

ydkmlt84 (approved):
Looks good. I am sure someone will still have questions but maybe this will help a little bit.

Issue #1620, opened 2025-03-07
PlexApiService Repeated Failure while fetching watchlist of user with ID:


paytah232:
**Describe the bug**
When running rulesets with options to check watchlist, I see many error for a particular user.
`[maintainerr] | 07/03/2025 15:27:02  [WARN] [PlexApiService] Failure while fetching watchlist of user with ID: d84598c3a5cd7d63`
I don't know why this occurs, or how to check the particular user ID (or how to delete it).

**To Reproduce**
Unknown - I would assume that this error occurs when the api cannot return a value and is a normal response to the issue.

**Expected behavior**
I can't time it, but I can only assume that this query failing increases processing time. I think it would make the system logs the user ID this call fails again. When it fails a large number of times (not sure what that is; 5? 50?), it removes that user from that particular query list.

Ideally, this speeds up the rate at which the rules with run, and reduces unnecessary server load where we know the api is going to fail consistently.

When this user is blacklisted - perhaps it is only for the current day. On the next day, the same thing happens again, until the user is blacklisted again.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d27d704e-2252-4311-b9ca-fc21652d37e9)

**Device (please complete the following information):**
 - OS: Linux (Synology NAS) running on docker
 - Version 2.10

ben-propflo:
The queries are cached here so the running time won't be affected. Could you [enable debug logging ](https://docs.maintainerr.info/latest/Installation/#enabling-debug-logging) and re-run your rules? This should give us a more helpful error message to go on. I agree the log message isn't great either, we should include the username really.

paytah232:
@benscobie Turned on debug, and get the following additional debug lines now:
```
maintainerr  | [maintainerr] | 08/03/2025 14:16:04  [WARN] [PlexApiService] Failure while fetching watchlist of user with ID: d84598c3a5cd7d63
maintainerr  | [maintainerr] | 08/03/2025 14:16:04  [DEBUG] [PlexApiService] Cannot read properties of null (reading 'user')
```

benscobie:
@paytah232 [v2.12.0](https://github.com/jorenn92/Maintainerr/releases/tag/v2.12.0) has improved the log messages here. Give that release a go and check your logs again.

paytah232:
@benscobie 
Updated, and looking good - Seems like my issue is because the of the user's privacy settings
[...]

PR #1622, opened 2025-03-08
chore: Improve watchlist query error message

benscobie:
Include the Plex username in the error message, as well as printing out error messages returned from the API. This is to help users more easily identify which user(s) are problematic. To aid with: https://github.com/jorenn92/Maintainerr/issues/1620


ydkmlt84 (approved):
👍

PR #1627, opened 2025-03-11
feat: Add show rating rules to season & episode types

benscobie:
This PR builds on this [recent PR](https://github.com/jorenn92/Maintainerr/pull/1604). I've added these in response to [this feedback](https://discord.com/channels/1152219249549512724/1345167415142776892/1349057400711155808) on Discord.

PR #1630, opened 2025-03-12
feat: Add smart collection filters

benscobie:
This PR adds 4 new rules:
- Present in amount of other collections (incl. smart collections)
- Present in amount of other collections (incl. parents and smart collections)
- [list] Collections media is present in (titles) (incl. parents and smart collections)
- [list] Collections media is present in (titles) (incl. smart collections)

These use the same logic as the existing non-smart collection rules, with the addition of checking smart collections.

Smart collections are not returned in the metadata Collection array. Instead we must iterate over each smart collection, load its children and then check if the rating key is present.

The rule IDs skip a few to cater for the ones added in PR #1627


ydkmlt84 (approved):
👍🏻

PR #1648, opened 2025-03-23
Change labels for sonarr actions with mediaType=shows for clarity

kaisermann:
Based on [this discussion](https://github.com/jorenn92/Maintainerr/discussions/1643#discussioncomment-12591340) and my own confusion around these Sonarr actions when `mediaType=shows`, I just added a `show` after the `unmonitor` verb so it's clearer what the code is actually doing.


ui/src/components/Rules/RuleGroup/AddModal/index.tsx
```diff
@@ -580,11 +580,11 @@ const AddModal = (props: AddModal) => {
                           },
                           {
                             id: 1,
-                            name: 'Unmonitor and delete all seasons / episodes',
+                            name: 'Unmonitor show and delete all seasons / episodes',
```
benscobie, line 583: Same for this but I need to validate what happens to the episode monitor status. It doesn't look like we explicitly unmonitor each episode, but I am wondering if unmonitoring the season somehow causes the episodes to be unmonitored:

Unmonitors show
Unmonitors seasons
Unmonitors(?) and deletes existing episodes

enoch85:
@kaisermann Is this PR still something you need/want to merge?

Asking since you didn't respond to @benscobie for one year now.

enoch85:
@kaisermann You are welcome to join the development btw. :+1:

PR #1649, opened 2025-03-23
Add sonarr show status check to `mediaType=season`

kaisermann:
Putting this in draft as it wasn't discussed before.
---
Coming from [this discussion](https://github.com/jorenn92/Maintainerr/discussions/1643#discussioncomment-12591340), I would like to be able to differentiate my `show` rules between `ended` and `continuing` shows. Currently, we can only check for the `status` when `mediaType=show`.

This PR makes the show status check possible with `mediaType=season`.

I didn't test this thoroughly, but I ran the application locally, connected it to my Plex, added a sample rule, and it seemed to work as expected. 

In time, is there a reason why this wasn't allowed before?


benscobie:
> In time, is there a reason why this wasn't allowed before?

No reason, just an oversight. I can't see a reason why a season or episode type can't target show level rules. The same goes for episodes types and targeting season level rules.

kaisermann:
> > In time, is there a reason why this wasn't allowed before?
> 
> No reason, just an oversight. I can't see a reason why a season or episode type can't target show level rules. The same goes for episode types and targeting season level rules.

Okay! Will work on updating the rule constants later then. From what I could gather, having `showType = undefined` is the same as defining the three `EPlexDataType` (Shows, Seasons, Episodes). Do you prefer it to be more explicit with the array (some are like these) or the `undefined` approach (others are also like this 😅)?

benscobie:
> Okay! Will work on updating the rule constants later then.

I haven't checked this, but some might not be as straight forward as just updating the constant. If they're solely referencing `showResponse` and the description text makes sense it'll likely be fine. For the episode referencing season level data case I described, it might be a bit more troublesome. For example **sw_lastWatched / Newest episode view date** targets shows & seasons and specifically gets the watch history for the media items rating key. So for a show it would reference the show, and then the season if a season type. You wouldn't want to get the watch history for the episode for **Newest episode view date** though as that's incorrect. You'd need to add a condition to check the type and then do something different like look at the season instead, we do that in quite a few places. The other difficulty with this specific one is, if you saw **Newest episode view date** for an Episode media type, you don't know if that's referring to the season or the show. You could also argue that it would be nice to have both options available, **Newest episode view date (season)** and **Newest episode view date (show)** which would mean needing two rules. The rule descriptions are hard-coded as well so we can't change it based on the media type selection either.

I don't expect you to resolve any of those cases though!

> [From what I could gather, having `showType = undefined` is the same as defining the three `EPlexDataType`](https://github.com/kaisermann/Maintainerr/blob/9516e6b74aacabbf3e319f59c47dd08f5678d3e2/ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx#L487-L488) (Shows, Seasons, Episodes). Do you prefer it to be more explicit with the array (some are like these) or the `undefined` approach (others are also like this 😅)?

Whichever has the most amount of usages probably 😆

kaisermann:
> @benscobie: You could also argue that it would be nice to have both options available, Newest episode view date (season) and Newest episode view date (show) which would mean needing two rules

That makes sense IMO. 

> I don't expect you to resolve any of those cases though!

Ahahah, I appreciate the clarity! Maybe one day! 🙈

---

On this topic of rules changing what they target, I stumbled upon the rule "Number of seasons / episodes (also unavailable)" which can be used for both `show` and `seasons`. Does it actually mean `Number of seasons when mediaType=show` and `Number of episodes when mediaType=season`? If that's the case, couldn't/shouldn't these be separate rules?

---

Would it be ok to merge this with just the `ended` / `status` checks for now and move this to a discussion thread?

benscobie:
> Does it actually mean `Number of seasons when mediaType=show` and `Number of episodes when mediaType=season`? If that's the case, couldn't/shouldn't these be separate rules?

Correct. There are a few other rules like this but I'm not sure if there has ever been a rule when they should be split out or not. Is the naming confusing to you?

> Would it be ok to merge this with just the `ended` / `status` checks for now and move this to a discussion thread?

Absolutely!

benscobie (approved):
Thank you for the PR ❤️

PR #1659, opened 2025-03-26
feat: Add rule & collection job status indicators

benscobie:
This PR adds rule & collection job progress above the version info, similar to *arrs but with actual bars! The run buttons now reflect the task running status and I have added an error toast if trying to run when already running (edge case if the status isn't available as the button is usually disabled).

rules_progress

This uses the event emitter from Nest.js to emit started, progress and finished events for both jobs. The events are sent down to the client using a new SSE endpoint and `@OnEvent` decorators. I've set this up to make adding new events and listening to them on the client super easy. One future use-case would be emitting events when media is added & removed from a collection and having the collection view be live updating.

### DIFF
--- a/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
+++ b/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
@@ -1,8 +1,6 @@
 import { Logger } from '@nestjs/common';
 import { ServarrApi } from '../common/servarr-api.service';
 import {
-  AddSeriesOptions,
-  LanguageProfile,
   SonarrEpisode,
   SonarrEpisodeFile,
   SonarrInfo,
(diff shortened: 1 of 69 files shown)


server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
```diff
@@ -330,32 +330,33 @@ export class SonarrApi extends ServarrApi<{
...
+          );
+
+          if (episodesForSeason.length > 0) {
+            await this.UnmonitorDeleteEpisodes(
+              +seriesId,
+              s.seasonNumber,
+              episodesForSeason.map((x) => x.id),
+              false,
+            );
+          }
```
benscobie, line 352: I've had this change sat around for a while so going to stuff it in here. This just adds a missing await to `UnmonitorDeleteEpisodes` and removes the unnecessary for loop.

benscobie:
> I do have some concerns about the changes to the rule and collection workers (not because of the code, but simply because it’s an area I don't like to touch).

I'll add unit tests for both of the workers, might as well start now! I think I'll split out the *arr deletion logic to new classes at the same time as well. There's a feature request for one-off handling in the UI which will be easy after that refactor.

jorenn92:
> > I do have some concerns about the changes to the rule and collection workers (not because of the code, but simply because it’s an area I don't like to touch).
> 
> I'll add unit tests for both of the workers, might as well start now! I'll split out the media deletion logic to new classes at the same time to make the tests easier to write. There's a feature request for one-off handling in the UI which will be easy to implement after that refactor.

Great idea! Unit tests for these areas would be very valuable.

benscobie:
@jorenn92 https://github.com/jorenn92/Maintainerr/pull/1659/commits/1bf300de391e024ac0c4883ec11718815f29f04f should be good to go.

A summary of the changes:
[new] CollectionHandler - Responsible for handling a single collection media item. Tests added for this.
[new] RadarrActionHandler & SonarrActionHandler - Responsible for handling a single collection media item for a specific service. . Tests added for this.
[new] MediaIdFinder - Contains the TVDBId lookup code. I noticed there's a few other ID lookup type functions elsewhere so I may need to tidy. I haven't added tests for this.
[changed] CollectionWorkerService - This class is now very simple. It looks up collections & media to handle and uses CollectionHandler for the handling, and then triggers availability sync. Tests added for this.

benscobie:
> Would it make sense to do a pre-release first before rolling it out to everyone?

Definitely 😎

danthom1704:
I love this. Now I don't have to go to the logs to see if it finished. Great work.

PR #1661, opened 2025-03-26
feat: allow jellyfin users in jellyseerr integration

### LINKED ISSUE #1658: Jellyseerr user matching not working
**Describe the bug**
Utilizing any rule involving user matching with Jellyseerr does not work. Testing media shows a blank list "[]" for values.

**To Reproduce**
1.  Create a rule utilizing "Jellyseerr - Requested by user (Plex or local username)"
2.  Check that rule against any media inside collections
3. See results

**Expected behavior**
Users who request media via Jellyfin should be matched inside Maintainerr.

**Screenshots**

Image

 - OS: Unraid container
 - Version 2.12.2

**EDIT**

This issue seems to be specific to users logging in via Jellyfin. Local Jellyseerr accounts work as expected.

Sean-Destefano:
Adds the ability for Jellfin users signing into Jellyseerr to be matched in rules.

Resolves #1658

### DIFF
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -910,7 +910,8 @@ export class RuleConstants {
         {
           id: 0,
           name: 'addUser',
-          humanName: 'Requested by user (Jellyfin, Emby, Plex or local username)',
+          humanName:
+            'Requested by user (Jellyfin, Emby, Plex or local username)',
           mediaType: MediaType.BOTH,
           type: RuleType.TEXT,
         }, //  returns username[]
--- a/server/src/modules/rules/constants/rules.constants.ts
+++ b/server/src/modules/rules/constants/rules.constants.ts
@@ -910,7 +910,7 @@ export class RuleConstants {
         {
           id: 0,
           name: 'addUser',
-          humanName: 'Requested by user (Plex or local username)',
+          humanName: 'Requested by user (Jellyfin, or local username)',
           mediaType: MediaType.BOTH,
           type: RuleType.TEXT,
         }, //  returns username[]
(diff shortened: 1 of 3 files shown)


server/src/modules/rules/getter/jellyseerr-getter.service.ts
```diff
@@ -116,6 +116,8 @@ export class JellyseerrGetterService {
                     if (includesSeason) {
                       if (request.requestedBy?.userType === 2) {
                         userNames.push(request.requestedBy?.username);
+                      } else if (request.requestedBy?.userType === 3) {
```
benscobie, line 119: So it looks Emby linked accounts also populate jellyfinUsername, so we can support that too:

```suggestion
                      } else if (request.requestedBy?.userType === 3 || request.requestedBy?.userType === 4) {
```

server/src/modules/rules/getter/jellyseerr-getter.service.ts
```diff
@@ -130,6 +132,8 @@ export class JellyseerrGetterService {
                     // for shows and movies, add every request user
                     if (request.requestedBy?.userType === 2) {
                       userNames.push(request.requestedBy?.username);
+                    } else if (request.requestedBy?.userType === 3) {
```
benscobie, line 135:
```suggestion
                    } else if (request.requestedBy?.userType === 3 || request.requestedBy?.userType === 4) {
```

server/src/modules/rules/constants/rules.constants.ts
```diff
@@ -910,7 +910,7 @@ export class RuleConstants {
         {
           id: 0,
           name: 'addUser',
-          humanName: 'Requested by user (Plex or local username)',
+          humanName: 'Requested by user (Jellyfin, or local username)',
```
benscobie, line 913:
```suggestion
          humanName: 'Requested by user (Jellyfin, Emby, Plex or local username)',
```

benscobie (approved):
Thank you for the PR ❤️

PR #1664, opened 2025-03-30
Feat : partial jellyfin & qBittorrent & docker for local developpement

JolanLar:
Jellyfin:

- Set rules for the last viewed date and time for movies and shows.
- Support rules for all types of shows, seasons, and episodes.
- Do not collect media from Jellyfin; Plex is still required.
- Use the Jellyfin API without the need for any additional plugins.
- Ensure that TMDb data between Jellyfin and Plex is synchronized.

qBittorrent:

- Add torrent deletion settings to rules.
- Get the torrents to delete from Radarr/Sonarr.
- Perform deletion when the rule operation is set to deletion.

Docker:

- Use docker-compose up -d to start the project locally in development without installing Yarn or other dependencies.


JolanLar:
It you want me to do some adjustements you can ask me here :)

benscobie:
@JolanLar Firstly thank you for the PR, there is a lot to digest here!

I'll start with the qBitorrent addition for the moment. I had a look at implementing torrent deletion a little while ago and it is quite a can of worms. There are a lot of scenarios to deal with which is why I didn't start 😆 

I'm not sure this is going to work for Sonarr the way it currently is. Say I have a rule that is set to `Episodes`. The episode was downloaded as part of a season pack. I believe here, Maintainerr would delete the entire season pack if just 1 episode is handled. The same goes if you downloaded the entire show in a single torrent, targeting season or episode would mean the whole show gets deleted. Radarr doesn't really suffer from this issue as it's a lot simpler.

This will also miss torrents in the scenario where an episode was downloaded as a single torrent and then replaced with a season pack later.

Lastly, if you're cross seeding the torrent, this will leave that torrent in an error state as the files have been removed. If you cross seed using hard links then you have a different problem as the space won't have been freed.

Ref the Docker addition, I think [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers) might be better suited here. Windows suffers bad IO performance when mounting a local volume in WSL2. The recommendation there is to clone inside the container, devcontainers in VSCode makes this easy with `Dev Containers: Clone Repository in Container Volume...`

We typically create a PR per feature/bug fix, and certainly in this case with multiple chunky features. Looking at your commit log you should be able to cherry pick them out easily enough.

JolanLar:
Thanks for the feedback. I’ve only used the `shows` rules so far, so I haven’t gone much deeper into the torrent & Sonarr features yet. I’ll take a look in the next few days to properly implement `season` and `episode` management.

For the upgraded files, I think the request made to `/history` returns all the torrents, not just the last one, so they’re going to be deleted too. But I’ll double-check to be sure.

I'm not familiar with cross-seeding, but I think the easiest way would be to tag all the torrents with the Sonarr/Radarr tag, so they’d appear in the /history request and be deleted as well. Otherwise, I can check for other torrents sharing the same files and delete them.

For the Docker addition, I used the same logic as [Overseerr](https://github.com/sct/overseerr), but the VSCode [devcontainers](https://code.visualstudio.com/docs/devcontainers/containers) look interesting. I’ll look into that.

inject3d:
I was wondering if this PR will be merged in the near future? Any chance to get rid of the plex integration and requirements?

JolanLar:
Hi, after upgrading Jellyfin to version 10.11.0 (a major update), my Jellyfin integration no longer works. According to their patch notes, they’ve made several breaking changes to the database: https://jellyfin.org/posts/jellyfin-release-10.11.0/.
I’ll take a look and make the necessary adjustments to get it working again, but I won’t be implementing the additional features requested for this branch to be merged.
If anyone would like to pick this up, feel free to do so.

ydkmlt84 (commented):
Closing this as we will not be supporting any download client removal. There are other, better suited tools out there already.

[qbitmanage](https://github.com/StuffAnThings/qbit_manage)

[Cleanuparr](https://github.com/Cleanuparr/Cleanuparr)

We also have a beta test for the Jellyfin support. You can try it out with the docker tag `jellyfin-dev`

Lastly, this PR is quite old and a lot of changes have happened since then.

PR #1675, opened 2025-04-02
chore: Set retry log level to debug

benscobie:
These aren't particularly helpful in most cases and are spamming user's logs. If all retries fail, the function executing these API requests should be handling any errors and logging appropriately anyway.


ydkmlt84 (approved):
👍

PR #1679, opened 2025-04-02
chore: Improve Overseerr/Jellyseerr response types

benscobie:
### Description

I spotted that some of the Overseerr response types were incorrect. I've separated TV and Movie into their own types as per their OpenAPI spec, and split the `/tv/x/season/xx` request code into its own function.

I have also separated the request types, though this is not per their spec. They combine the types into one object, meaning a movie type will always have an empty seasons array for example. These type changes prevent accidentality accessing seasons on a movie request.

I used Copilot to copy the changes across to Jellyseerr as they're almost identical, and it did a pretty good job after a few attempts!


ydkmlt84 (approved):
👍

PR #1680, opened 2025-04-02
fix: Ensure signals are passed down to the apps

### LINKED ISSUE #1672: Docker stop takes longer than expected
Not sure why, but the docker container always takes a long time to stop `docker stop mainainerr` like its waiting for a timeout to be enacted.

benscobie:
Fixes #1672


ydkmlt84 (approved):
👍

PR #1681, opened 2025-04-02
chore: Split Docker platform builds across workers

benscobie:
### Description

This PR splits the amd64 & arm64 Docker builds up so that they run on their own runners using a matrix. I followed [this](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) to achieve that. I have also removed using QEMU for arm64 support, instead using the [new preview arm64 runners](https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/) which results in a much faster build :rocket:.

This results in builds taking 3-4 minutes instead of 20 minutes.


benscobie:
@jorenn92 As I have split the build steps up here, the branch rule's status checks will need updating to point to those as currently this PR is not mergeable. I can do this if you're happy to give me permissions to do so.

jorenn92:
> @jorenn92 As I have split the build steps up here, the branch rule's status checks will need updating to point to those as currently this PR is not mergeable. I can do this if you're happy to give me permissions to do so, otherwise whenever you have some spare time 😄

Great improvement! I'd be happy to give you permission to change the branch protection rules. But i don't seem to be able to find a way to grant that access. Or i could just be blindly missing the config option.. 

Anyway, i've changed the rules to now require 'Build linux/amd64' & 'Build linux/arm64' instead of the old 'Build'. 

The other PRs are possibly broken now until this one is merged though.

benscobie:
> I'd be happy to give you permission to change the branch protection rules. But i don't seem to be able to find a way to grant that access. Or i could just be blindly missing the config option..

Ah man, it seems personal repository's are very limited: https://docs.github.com/en/get-started/learning-about-github/access-permissions-on-github#personal-accounts

We would have to create an organization but as you've updated it now I think we can leave that alone for the moment 😆

PR #1684, opened 2025-04-03
chore: improve types in RuleInput component

bitnimble:
### Description

This PR improves types of a few variables in the RuleInput component, changing `string` or `number` to the narrower type that it actually is.
Also changes `action` from `string` to `RulePossibility` (a number), and updates reads + writes accordingly. 

### Checklist

### DIFF
--- a/ui/src/contexts/constants-context.tsx
+++ b/ui/src/contexts/constants-context.tsx
@@ -49,23 +49,23 @@ export enum RulePossibility {
   COUNT_SMALLER,
 }
 
-export enum RulePossibilityTranslations {
-  BIGGER = 'Bigger',
-  SMALLER = 'Smaller',
-  EQUALS = 'Equals',
-  NOT_EQUALS = 'Not Equals',
-  CONTAINS = 'Contains (Exact list match)',
-  BEFORE = 'Before',
-  AFTER = 'After',
-  IN_LAST = 'In Last',
-  IN_NEXT = 'In Next',
-  NOT_CONTAINS = 'Not Contains (Exact list match)',
-  CONTAINS_PARTIAL = 'Contains (Partial list match)',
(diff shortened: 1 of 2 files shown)


ui/src/contexts/constants-context.tsx
```diff
@@ -49,23 +49,23 @@ export enum RulePossibility {
...
+  [RulePossibility.IN_LAST]: 'In Last',
+  [RulePossibility.IN_NEXT]: 'In Next',
+  [RulePossibility.NOT_CONTAINS]: 'Not Contains (Exact list match)',
+  [RulePossibility.CONTAINS_PARTIAL]: 'Contains (Partial list match)',
+  [RulePossibility.NOT_CONTAINS_PARTIAL]: 'Not Contains (Partial list match)',
+  [RulePossibility.COUNT_EQUALS]: 'Count Equals',
+  [RulePossibility.COUNT_NOT_EQUALS]: 'Count Does Not Equal',
+  [RulePossibility.COUNT_BIGGER]: 'Count Is Bigger Than',
+  [RulePossibility.COUNT_SMALLER]: 'Count Is Smaller Than',
 }
```
benscobie, line 69: We could move these to `packages/contracts` to remove the duplication between server & UI.

Not sure on the best namespace for them though, constants sounds like a catch all 🤔 What do you think?

bitnimble: I'd leave these ones here (UI facing strings, so belongs in UI imo), but the actual rule possibilities I'd move to contracts for sure. but I've got a few others to move as well, so let's leave that for another PR I reckon

ben-propflo: Ah yep sounds good.

benscobie: Ah yep sounds good.

benscobie (approved):

PR #1693, opened 2025-04-07
fix: Restrict Test Media to the collection's library

### LINKED ISSUE #1455: Test media doesn't limit media search to the selected library
It only appears to limit based on the type, movie or TV. If I have two TV libraries, test media will search across them both which can be confusing.

Introduce a drop-down with the ability to change the library to search in, with an option to search all to maintain existing logic.

benscobie:
### Description

This PR updates the Test Media modal to only search the library that the collection belongs to. The previous behaviour was causing issues for people that had the same media in different libraries (separate 1080p and 4k libraries for example), as duplicate items would appear in the drop-down. Selecting one outside of the library could lead to different results.

### Related issue

Fixes #1455

### How to test

1. Validate Test Media for each media type
2. Create two Plex libraries containing the same media item. Create a rule for one of those libraries. Run Test Media and validate that only 1 of the media items is returned instead of 2. Perform a test and validate the results are as expected.


ydkmlt84 (approved):
👍Tested

PR #1696, opened 2025-04-08
feat: update rule creation UI/UX

ydkmlt84:
### Description

- Updates the rule creation page layout to be cleaner and easier to understand.
- Adds `Made with Version` to the community rules table
- Adds `required services` and `level` (media type) to a rule upload for future use of only showing rules available based on media type and services configured
- Replaces react-toast-notifications with toastify (dark theme)
- Changes the nav bar version indicator URL to `/settings/about`
- changed rule constants applications to remove the const to allow mapping for rule uploads. Added ID numbers to always be consistent going forward

### Related issue

None

### Checklist

- [ x ] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document.
- [ x ] I have performed a self-review of my code.
- [ x ] I have linted and formatted my code.
- [ x ] My changes generate no new warnings.
- [ x ] New and existing unit tests pass locally with my changes.

### How to test

1. Create rules and edit existing rules
2. Click on the version indicator in bottom left to ensure it goes to the right page
3. Upload a rule to the community
4. Initiate an action that shows a toast notification

### Additional context

localhost_3000_rules (2)
image


bzly:
Nice, these changes definitely make the UI clearer. 

I think it would be good if the section operator was moved outside of the first rule though, as it is not specific to it—ideally between the sections as it 'belongs' to the former as much as the latter. 

It might be outside the scope of this pull request, but some UI to better control the logic between the sections would be nice, so one doesn't have to think about the order of operations, e.g. it would be wonderful to be able to specify `((1 OR 2) AND (3 OR 4)) OR 5`.

ydkmlt84:
> Nice, these changes definitely make the UI clearer. 
> 
> 
> 
> I think it would be good if the section operator was moved outside of the first rule though, as it is not specific to it—ideally between the sections as it 'belongs' to the former as much as the latter. 
> 
> 
> 
> It might be outside the scope of this pull request, but some UI to better control the logic between the sections would be nice, so one doesn't have to think about the order of operations, e.g. it would be wonderful to be able to specify `((1 OR 2) AND (3 OR 4)) OR 5`. 

I definitely have some ideas for an even better experience but I thought this somewhat quick UI change would be handy for now. 

I agree about the section operator, but the real problem was real estate. I did have it a little different and in between, but I wasn't a fan of how it looked. 

The next time I mess with this page, it will be much better, but I think it will be this way for a little while. Until I can get around to really changing things.

jorenn92:
Nice change! I tried it out for a bit, and it looks great.
I was concerned it might not work well on mobile, but I'm glad to see you took care of that.

The only issue I noticed is when importing rules. The UI doesn't update to reflect the changes, even though the rules are correctly saved after hitting "Save."

Could you update this PR with the latest from main?

ydkmlt84:
> Nice change! I tried it out for a bit, and it looks great. I was concerned it might not work well on mobile, but I'm glad to see you took care of that.
> 
> The only issue I noticed is when importing rules. The UI doesn't update to reflect the changes, even though the rules are correctly saved after hitting "Save"
> 
> Could you update this PR with the latest from main?

I could not recreate that importing behavior. Could you give me an example? I will pull all of the latest chanages and update.

jorenn92:
> > Nice change! I tried it out for a bit, and it looks great. I was concerned it might not work well on mobile, but I'm glad to see you took care of that.
> > The only issue I noticed is when importing rules. The UI doesn't update to reflect the changes, even though the rules are correctly saved after hitting "Save"
> > Could you update this PR with the latest from main?
> 
> I could not recreate that importing behavior. Could you give me an example? I will pull all of the latest changes and update.

I can’t reproduce the issue anymore. I retried using the PR image generated above, and it works now.
When I reported it, I had only tested it in VS Code. Something was probably still off with my setup there.

Sorry for sending you on a wild goose chase!

ydkmlt84:
> > > Nice change! I tried it out for a bit, and it looks great. I was concerned it might not work well on mobile, but I'm glad to see you took care of that.
> 
> > > The only issue I noticed is when importing rules. The UI doesn't update to reflect the changes, even though the rules are correctly saved after hitting "Save"
> 
> > > Could you update this PR with the latest from main?
> 
> > 
> 
> > I could not recreate that importing behavior. Could you give me an example? I will pull all of the latest changes and update.
> 
> 
> 
> I can’t reproduce the issue anymore. I retried using the PR image generated above, and it works now.
> 
> When I reported it, I had only tested it in VS Code. Something was probably still off with my setup there.
> 
> 
> 
> Sorry for sending you on a wild goose chase!

Ooohh. I think that is because in dev mode useEffects fire twice. I was seeing something like that too, and chased it down for a while, but never could find the problem. Since it wasn't happening when not in dev, I let it go. It depends on what you're importing and what rules were there already, for you to see it happen.

ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
```diff
@@ -20,13 +21,21 @@ const CommunityRuleUpload = (props: ICommunityRuleUpload) => {
...
-      await PostApiHandler(`/rules/community`, {
+      const requiredServices = detectRequiredServices(props.rules)
+      const payload: any = {
         name: nameRef.current.value,
         type: props.type,
         description: descriptionRef.current.value,
         JsonRules: props.rules,
         uploadedBy: uploadedByRef.current.value || undefined,
-      })
+        requiredServices,
```
benscobie, line 31: I cannot see a need to save the required services. You have access to the JsonRules for each upload so you can infer the applications (detectRequiredServices) at the point of display/import.

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
```diff
@@ -362,201 +355,205 @@ const RuleInput = (props: IRuleInput) => {
...
+            className="w-full rounded-lg p-2 text-zinc-100 focus:border-amber-500 focus:ring-amber-500"
+          >
+            <option value={undefined}> </option>
+            <optgroup label="Custom values">
+              {ruleType === RuleType.DATE ? (
+                <>
+                  <option value={CustomParams.CUSTOM_DAYS}>
+                    Amount of days
+                  </option>
+                  {action &&
```
benscobie, line 439: This must've been a merge conflict as it was recently changed from this to fix a bug

```suggestion
                  {{action != null &&
```

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
```diff
@@ -362,201 +355,205 @@ const RuleInput = (props: IRuleInput) => {
...
+              ) : undefined}
+              {ruleType === RuleType.TEXT ? (
+                <option value={CustomParams.CUSTOM_TEXT}>Text</option>
+              ) : undefined}
+              <MaybeTextListOptions ruleType={ruleType} action={action} />
+            </optgroup>
+            {ConstantsCtx.constants.applications?.map((app) => {
+              return (app.mediaType === MediaType.BOTH ||
+                props.mediaType === app.mediaType) &&
+                action &&
```
benscobie, line 462:
```suggestion
                action != null &&
```

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
```diff
@@ -362,201 +355,205 @@ const RuleInput = (props: IRuleInput) => {
...
+                type="number"
+                name="custom_val"
+                id="custom_val"
+                onChange={updateCustomValue}
+                value={customVal ? +customVal / 86400 : undefined}
+                placeholder="Amount of Days"
+                className="w-full rounded-lg p-2 text-zinc-100 focus:border-amber-500 focus:ring-amber-500"
+              />
+            ) : customValType === RuleType.TEXT &&
+              secondVal === CustomParams.CUSTOM_TEXT ? (
```
benscobie, line 513: Merge conflict? The RuleType.TEXT_LIST condition was removed

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
```diff
@@ -441,627 +432,671 @@ const AddModal = (props: AddModal) => {
...
-                    id="description"
-                    rows={5}
-                    defaultValue={props.editData?.description}
-                    ref={descriptionRef}
-                  ></textarea>
-                </div>
-              </div>
-            </div>
+        <form className="flex flex-col">
+          <div className="grid grid-cols-1 gap-x-5 gap-y-3 md:grid-cols-2">
```
benscobie, line 436: I think the breakpoints need some tweaking. I personally think that `md:grid-cols-2` should be `xl:grid-cols-2` or even `2xl:grid-cols-2` as most input labels are wrapping after 1 or 2 words. I think leaving all child md breakpoints as-is is fine, it gives the larger screens a less compressed view.

```suggestion
          <div className="grid grid-cols-1 gap-x-5 gap-y-3 xl:grid-cols-2">
```

ui/src/utils/CommunityRuleMaps.tsx
```diff
@@ -0,0 +1,43 @@
...
+  JELLYSEERR,
+}
+
+export enum TVLevel {
+  SHOW = 'show',
+  SEASON = 'season',
+  EPISODE = 'episode',
+}
+
+export const applicationNames: Record<number, string> = {
```
benscobie, line 18: It's better to reference the enum as the key here. When Application is updated, this line will error until the new value is added. It's also just generally more readable.

```suggestion
export const applicationNames: Record<Application, string> = {
```

ui/src/components/Settings/Plex/index.tsx
```diff
@@ -250,39 +254,29 @@ const PlexSettings = () => {
 
   const refreshPresetServers = async () => {
     setIsRefreshingPresets(true)
-    let toastId: string | undefined
```
benscobie, line 253: They have quite a nice API for promises which cleans this up:

```
    const toastId = 'plex-refresh-preset-servers'

    const serverPromise = GetApiHandler<PlexDevice[]>(
      '/settings/plex/devices/server2s',
    )

    try {
      const response = await toast.promise(
        serverPromise,
        {
          pending: 'Retrieving server list from Plex',
          success: 'Plex server list retrieved successfully!',
          error: 'Failed to retrieve Plex server list.',
        },
        {
          toastId,
        },
      )

      if (response) {
        setAvailableServers(response)
      }
    } finally {
      setIsRefreshingPresets(false)
    }
```

ui/src/components/Common/CommunityRuleModal/index.tsx
```diff
@@ -325,6 +330,9 @@ const CommunityRuleModal = (props: ICommunityRuleModal) => {
             onSubmit={handleUpload}
             type={props.type}
             rules={props.currentRules ? props.currentRules : []}
+            {...(props.type === 'show' && props.level
+              ? { level: props.level }
+              : {})}
```
benscobie, line 335:
```suggestion
            level={
              props.type === 'show' && props.level ? props.level : undefined
            }
```

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
benscobie: The grid/flex layout is a bit wonky, is there negative margins somewhere? Options also has quite a bit of space on the right that could be used up on the smaller screens.

image

ui/src/components/Common/CommunityRuleModal/index.tsx
ydkmlt84: I think I may have "fixed" my merge conflicts backwards. I applied your suggestion.

ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
ydkmlt84: Technically it would be the same either way wouldn't it? Or you think it would be better/easier to check it on the other end, instead of at the upload? I do however like the idea of adding "metadata" to the rule uploads for other uses.

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
ydkmlt84: I confirmed that there is only one opening `{`

ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
ydkmlt84: I can only imagine yes, since it appears elsewhere that I did not remove. 95% of my changes here were supposed to be div placement and CSS. I must have "fixed" the merge conflicts incorrectly. 

It should be put back to the way it should be now.

ui/src/components/Settings/Plex/index.tsx
ydkmlt84: Nice. That is much cleaner and works great!

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
ydkmlt84: I agree with your statements, after review. I am not completely sold on 2xl:grid-cols-2 but I did change it to xl. I also made some other changes that I think you may like better. Let me know what you think.

ui/src/components/Rules/RuleGroup/AddModal/index.tsx
```diff
@@ -441,627 +432,677 @@ const AddModal = (props: AddModal) => {
...
-                  <div className="ml-0">
-                    <h3 className="heading">Rules</h3>
-                    <p className="description">
-                      Specify the rules this group needs to enforce
-                    </p>
+          </div>
+          <hr className="mt-6 h-px border-0 bg-gray-200 dark:bg-gray-700"></hr>
+          <div className="grid grid-cols-1">
+            <div className="flex justify-center">
+              <div className={`section ${useRules ? `` : `hidden`} md:w-3/4`}>
```
benscobie, line 982: We can pick this up in a future PR, but I think we could keep the rules section here full width until a larger breakpoint like XL. Some of the rule names are quite long so this change would allow the user to see more of it. I did this locally and it looks like some other styling needs changing elsewhere, hence I think we can pick this up later.

ui/src/components/Rules/Rule/RuleCreator/CommunityRuleUpload/index.tsx
benscobie: I just believe it's cleaner to not upload something we don't need, as well as keeping the schema slimmer.

PR #1697, opened 2025-04-08
fix: Unable to save rules with 'Bigger' actions

benscobie:
### Description

This PR fixes saving rules that use the 'Bigger' action. Introduced in #1684 - a condition needed to be updated as the value swapped from '0' to 0.


ydkmlt84 (approved):
👍tested

PR #1716, opened 2025-04-15
chore(ci): Fix Docker image tags

benscobie:
### Description

I poo poo'd the Docker image tags and a bunch of badly named ones got pushed (see https://hub.docker.com/r/jorenn92/maintainerr/tags - can we delete these?). I created a temporary branch to re-run just the Docker image build for 2.14.0 after I fixed the workflow.

I've added major and major.minor tags here as well. I like to pin to a major version as upgrading might involve manual steps. I think using `:latest` is a bad idea in most cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -131,7 +131,6 @@ jobs:
             ${{ env.GHCR_SLUG }}
           tags: |
             type=raw,value=main
-            type=raw,value=develop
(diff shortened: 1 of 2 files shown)


jorenn92 (approved):

jorenn92:
I've removed the faulty tags from Docker hub & ghcr.io.

PR #1754, opened 2025-04-29
feat: YAML Importer Modal additions

ydkmlt84:
### Description

Added an upload button to the YAML import modal, to upload a YAML file. Added a copy button to the export modal. Added a copy button to Test Media output.

This is to address iOS/Safari copy and paste issues with the monaco editor. Gives users the option to upload a rule YAML file instead of using copy and paste.

#################

**Note: I cannot test this functionality as I don't have a Mac for emulation purposes.**

#################

### Checklist

### How to test

1. Open a rule export and use the copy button to copy the contents of the export modal.
2. Use the upload button to upload a YAML file into the import modal.
3. Copy the output of Test Media with the copy button.


ui/src/components/Collection/CollectionDetail/TestMediaItem/index.tsx
```diff
@@ -168,6 +170,14 @@ const TestMediaItem = (props: ITestMediaItem) => {
...
   }
 
+  const copyToClipboard = () => {
+    const value = (editorRef.current as any)?.getValue?.()
+    if (value?.trim()) {
+      navigator.clipboard.writeText(value)
+      toast.success('Copied to clipboard')
+    }
+  }
+
```
benscobie, line 180: So copying to clipboard is kind of a nightmare in certain situations, like in Firefox it is only available in "[Secure Contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)", meaning it doesn't work for URLs like `http://maintainer.home.arpa`. This also appears to be the same for Chrome.

You could check if `navigator.clipboard` is not undefined at the top of the component, and then toggle the button display based on that.

`const isClipboardAvailable = navigator.clipboard !== undefined`

There are other hacks to workaround around this issue if you wanted to go down that route, such as: https://stackoverflow.com/a/65996386

`navigator.clipboard.writeText` can also fail so it's worth wrapping in a try catch and having an error toast to handle that case:

```suggestion
  const copyToClipboard = async () => {
    const value = (editorRef.current as any)?.getValue?.()
    if (value?.trim()) {
	try {
            await navigator.clipboard.writeText(value)
            toast.success('Copied to clipboard')
        } catch (error) {
            toast.error('Failed to copy to clipboard')
        }
    }
  }

```

benscobie (approved):
Nice, works good 🚀

PR #1774, opened 2025-05-16
chore: Inject logger and fix incorrect parameters

benscobie:
### Description

This PR replaces new'd up Logger instances with dependency injected MaintainerLogger's. Some usages of the Logger functions were passing objects to provide extra information, but the Winston transports aren't setup to handle this so you'd get messed up log messages such as:
`[maintainerr]  |  22/04/2025 00:13:47  [INFO] [[object Object]] Couldn't fetch Tautulli metadata!`

By using MaintainerLogger which directly wraps Winston we have much better control around this, as well as following good practice by using DI. I have fixed those incorrect usages, and moved the additional information to the log message where applicable.

I've created `MaintainerrLoggerFactory` for the places where we create new API clients inline. If we passed the original injected Logger in, messages would have an incorrect context. In the future I'd like to move API client instantiation to factories as well as I think that's cleaner.

### How to test

1. Load up the app and verify log messages still output


ydkmlt84 (approved):
Works fine on my end.

PR #1775, opened 2025-05-18
feat: Retry failed Plex API requests

benscobie:
### Description

This PR adds retrying of failed Plex API requests. I had done this for the other API integrations [a few months ago](https://github.com/jorenn92/Maintainerr/commit/35c36d9ffd2eab695df6795c0a45aa4b121634fe) but did not tackle Plex as it was using the fetch API instead. We now use axios here too, with the same retry mechanism. I have also removed redundant XML response handling code as we always receive JSON back. This was cherry-picked from the work being carried out in https://github.com/jorenn92/Maintainerr/pull/1705 - a user confirmed it solved their add & remove issue, they were receiving intermittent timeouts calling the Plex API.

Retries if it is a network error or a 5xx error on an idempotent request (GET, HEAD, OPTIONS, PUT or DELETE).

### How to test

It's hard to test a retry-then-succeed scenario given the intermittent nature of these failures. What I did was change my Plex URL to something invalid in Maintainerr and then run rules. You should see log messages 'Retrying...'  indicating a retried request. It might be possible to restart Plex mid rule run to try and simulate a failure but it might be tough.


ydkmlt84 (approved):
Tested. 👍

PR #1776, opened 2025-05-18
fix: Remove floating promises

benscobie:
### Description

This PR removes floating promises from the code base. These can cause several issues, such as ignored exceptions (catch block isn't used), potentially resulting in incorrect control flow, as well as spamming API requests when used in for loops.

This was cherry-picked from the work in #1705

### How to test

Almost all changes are simply `await` additions so don't really need any specific testing. Running rules & collection handling will suffice I think.


ydkmlt84 (approved):
Tested 👍

Issue #1446, opened 2024-12-19
Items get repeatedly removed and added back to collections


Zipties:
I am having an issue where all items in the collection are being removed each night, and then being added again shortly after

![image](https://github.com/user-attachments/assets/091fa261-ec97-411b-bb3c-ed1c94a993d9)

All my rules are exhibiting this behavior and I am pretty stumped over it. Here is one of them

 ```mediaType: MOVIES
rules:
  - "0":
      - firstValue: Plex.seenBy
        action: CONTAINS
        lastValue: Overseerr.addUser
      - operator: AND
        firstValue: Overseerr.isRequested
        action: EQUALS
        customValue:
          type: boolean
          value: "true"
      - operator: AND
        firstValue: Overseerr.mediaAddedAt
        action: BEFORE
        customValue:
          type: custom_days
          value: "90"
      - operator: AND
        firstValue: Radarr.tags
        action: NOT_CONTAINS_PARTIAL
        customValue:
          type: text
          value: dnd
  - "1":
      - operator: OR
        firstValue: Overseerr.mediaAddedAt
        action: BEFORE
        customValue:
          type: custom_days
          value: "60"
      - operator: AND
        firstValue: Overseerr.isRequested
        action: EQUALS
        customValue:
          type: boolean
          value: "true"
      - operator: AND
        firstValue: Radarr.tags
        action: NOT_CONTAINS_PARTIAL
        customValue:
          type: text
          value: dnd
  - "2":
      - operator: OR
        firstValue: Plex.lastViewedAt
        action: BEFORE
        customValue:
          type: custom_days
          value: "365"
      - operator: AND
        firstValue: Radarr.tags
        action: NOT_CONTAINS_PARTIAL
        customValue:
          type: text
          value: dnd
  - "3":
      - operator: OR
        firstValue: Plex.viewCount
        action: EQUALS
        customValue:
          type: number
          value: 0
      - operator: AND
        firstValue: Plex.addDate
        action: BEFORE
        customValue:
          type: custom_days
          value: "730"
      - operator: AND
        firstValue: Radarr.tags
        action: NOT_CONTAINS_PARTIAL
        customValue:
          type: text
          value: dnd
```

benscobie:
Could you enable debug logging:  https://docs.maintainerr.info/Installation.html#enabling-debug-logging

Wait for this to happen a few more times and then attach the logs here.

[...]

PR #1777, opened 2025-05-18
feat: Save rule results against collection log

benscobie:
### Description

This PR introduces a meta column to collection logs for storing arbitrary, but typesafe data. In this PR I am saving the rule evaluation results against the added & removed media collection log. These can then be viewed by clicking the meta link next to the log message. Manual additions and removals are also labelled with `(manual)`.

This was cherry-picked from the work in #1705.

### Related issue

This should help in debugging #1446 - and users often wonder why something got added or removed so this should help those guys out too.

### How to test

1. Create a collection with some rules that will match some media
2. Run rules
3. Collection -> Info
4. A meta link should appear next to the log message. Clicking that should bring up a modal similar to test media with the rule results displayed.


ydkmlt84 (commented):
Functionally, this worked. However the meta information seems a little off, because it is adding an operator at the end. I dont think it does anything, just something I noticed. 

This is from Test Media
```
- plexId: 110403
  result: true
  sectionResults:
    - id: 0
      result: true
      ruleResults:
        - operator: OR
          action: in_last
          firstValueName: Plex - Date added
          firstValue: 2025-05-18T23:22:20.000Z
          secondValueName: custom_days
          secondValue: 2025-05-15T14:45:21.428Z
          result: true
```

and this is from the meta on that same item
```
plexId: 110403
result: true
sectionResults:
  - id: 0
    result: true
    ruleResults:
      - operator: OR
        action: in_last
        firstValueName: Plex - Date added
        firstValue: 2025-05-18T23:22:20.000Z
        secondValueName: custom_days
        secondValue: 2025-05-15T14:43:31.345Z
        result: true
    operator: OR
```

also the `(meta)` link wasn't super obvious that is what I should be seeing/clicking on. maybe an icon/button?

benscobie:
@ydkmlt84 I've fixed the operator incorrectly appearing on the first section. Turns out this was an existing bug but didn't show up as test media only runs rules for 1 media item.

I've also replaced the link with a button icon. I've kept the `(manual)` text next to the message as it's just a label.

ydkmlt84 (approved):
👍

PR #1787, opened 2025-05-23
fix: Rule groups not in consistent order

### LINKED ISSUE #1786: Updating Rules Changes Their Order
**Describe the bug**

Each time I make an update to a rule, its position among other rules is changed and it is moved to the end of the order. This makes it confusing to find it the next time and also to keep rules sorted.

**To Reproduce**
1. Note a rule's location in the order among other rules.
2. Edit that rule and save.
3. Return to the rules overview. Observe the edited rule is now at the end of the order.

**Expected behavior**
Editing a rule should not change its position. There is no way to sort rules, so without that functionality, this gets rules out of order.

**Screenshots**

 - OS: Docker
 - Version 2.14.0

benscobie:
### Description

Rule groups were only sorted by their related rule IDs. These rules are re-created on rule group update which has the side effect of moving that rule group to the end of the list. This PR adds an additional sort on the rule group ID which orders them from oldest to newest.

### Related issue

Fixes #1786

### How to test

1. Updating a rule group should keep it in the same place in the list
2. New rule groups should be placed at the end of the list


ydkmlt84 (approved):
👍Tested!

PR #1790, opened 2025-05-24
chore: Fix release Docker tags

benscobie:
### Description

The reference to the version was incorrect in the workflow, meaning only the latest tag got pushed. I created a temporary branch off of the 2.15.0 tag, fixed the issue and re-ran that part of the release workflow to push tags for that release.

I also remembered you (@ydkmlt84) said the docs were deploying twice and found out why, so that is now fixed.


ydkmlt84:
> ### Description
> The reference to the version was incorrect in the workflow, meaning only the latest tag got pushed. I created a temporary branch off of the 2.15.0 tag, fixed the issue and re-ran that part of the release workflow to push tags for that release.
> 
> I also remembered you (@ydkmlt84) said the docs were deploying twice and found out why, so that is now fixed.

Yeah, it happened again yesterday too. Awesome that you found the problem!
image

ydkmlt84 (approved):
👍 ❤️

PR #1802, opened 2025-05-27
fix: Test media failing due to missing AbortSignal

benscobie:
### Description

Test media doesn't pass an abort signal (by design) so we just need to check if it's available before using. Another bug that would've been caught with strict null checks.

### How to test

1. Validate test media now works


ydkmlt84 (approved):
👍 Tested!

PR #1804, opened 2025-05-27
fix: Tautulli settings not updating until restart

benscobie:
### Description

When updating Tautulli settings, the in-memory values were not being updated and the API client was not re-built. This meant that any calls (excluding the Test endpoint) to the Tautulli API would use the old settings, which could be empty on a fresh install.

### How to test

1. Empty your Tautulli settings out
2. Restart the app
3. Add Tautulli settings
4. Add a Tautulli rule group
5. Run rules and check that everything worked ok


ydkmlt84 (approved):
👍Tested!

PR #1807, opened 2025-05-29
fix: Prevent undefined base path appearing in path

benscobie:
### Description

undefined gets output in template literals. In the `release-workflow` branch NEXT_PUBLIC_BASE_PATH is always defined which is where this change originated from.

### How to test

1. The stream endpoints should now work + the logos


ydkmlt84 (approved):
👍fixed

PR #1808, opened 2025-05-30
Improve UI memory handling

p1carats:
### Description

This PR addresses several memory and performance issues across the codebase.

Notably, it resolves an unbounded log accumulation issue that caused instability during long-running sessions. Logs are now properly capped or cleared to prevent excessive memory usage.

It also fixes cases where event listeners were added without proper cleanup on component unmount, and ensures timers use stable closures and are reliably cleared to avoid memory leaks or delayed side effects.

### Related issue

No issue has been opened yet for these problems.

### Checklist

- [X] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document.
- [X] I have performed a self-review of my code.
- [X] I have linted and formatted my code.
- [X] My changes generate no new warnings.
- [X] New and existing unit tests pass locally with my changes.

### How to test

Run the app as intended—nothing in behavior should have changed—but you should notice a reduced memory footprint, especially during longer sessions or prolonged usage. Keep an eye on devtools and system resource usage to confirm the improvements.

### Additional context

As I'm not fully familiar with every part of the codebase, there may still be other sources of memory or performance issues. This PR is a first step that addresses the most immediate and obvious problems.

Given the impact of these fixes, I recommend releasing a new hotfix or patch version shortly after merging.

PR #1813, opened 2025-05-31
fix: Patch remaining runtime and cleanup issues

p1carats:
### Description

This PR continues the work initiated in [#1808](https://github.com/jorenn92/Maintainerr/pull/1808), addressing several additional runtime and cleanup issues that were identified during testing.

Specifically:

- Prevents a crash when the Overseerr/Jellyseerr URL is left empty.
- Cleans up various event listeners on component unmount to avoid memory leaks.
- Cancels untracked Plex-related Axios requests to avoid unexpected behavior on component teardown.

These changes contribute to the broader goal of stabilizing long-running sessions and reducing memory bloat in edge scenarios.

### Related issue

Follow-up to #1808 — no formal issue created.

### Checklist

- [X] I have read the [CONTRIBUTING.md](CONTRIBUTING.md) document.
- [X] I have performed a self-review of my code.
- [X] I have linted and formatted my code.
- [X] My changes generate no new warnings.
- [X] New and existing unit tests pass locally with my changes.

### How to test

Run the app normally and test the following:

- Leave the Overseerr or Jellyseerr URL blank (which is often the case on first boot or when no data is configured). One of the two is typically empty, as using both simultaneously is not a common use case. Ensure no runtime crash occurs in this scenario.
- Trigger navigation across tabs and validate that event listeners are correctly cleaned up (e.g., no memory bloat, no duplicate events).
- Open the Plex tab, wait for API calls, then switch views — verify no cancelled Axios warning or memory leaks are triggered.

### Additional context

This is part of an ongoing effort to tighten the app's memory and lifecycle management.
While these patches resolve some remaining edge cases missed in #1808, there may still be other latent issues. A deeper audit may be worthwhile, but this PR might at least help improve long-session reliability and reduce noisy errors or warnings.

As previously mentioned, the removed commit related to timer handling and cleanup is still pending reintroduction. It was causing unexpected behavior and needs further refinement before being safely merged. This follow-up PR focuses on other isolated fixes, and that timer-related patch will be revisited in a future contribution.

*A follow-up hotfix or patch release is again recommended after merging.*


ui/src/components/Settings/Plex/index.tsx
```diff
@@ -213,6 +213,8 @@ const PlexSettings = () => {
   const verifyToken = (token?: string) => {
     const authToken = token || settingsCtx.settings.plex_auth_token
     if (authToken) {
+      const source = axios.CancelToken.source()
```
benscobie, line 216: This method of cancelling is deprecated, AbortController is the way to go now:

```suggestion
      const controller = new AbortController();
```

You also need to pass this into axios for it to work

ui/src/components/Settings/Plex/index.tsx
```diff
@@ -226,6 +228,11 @@ const PlexSettings = () => {
           setTokenValid(response.status === 200 ? true : false)
         })
         .catch(() => setTokenValid(false))
+
+      // Cancel the request if component unmounts
+      return () => {
+        source.cancel('Component unmounted')
```
benscobie, line 234:
```suggestion
        controller.abort()
```

ui/src/components/Settings/Logs/index.tsx
```diff
@@ -136,21 +136,27 @@ const Logs = () => {
...
+    }
+
+    es.addEventListener('log', handleLog)
 
     es.onerror = (e) => {
       console.error('EventSource failed:', e)
     }
 
     return () => {
+      es.removeEventListener('log', handleLog)
```
benscobie, line 156: Oops, good spot

benscobie:
I'm interested in your workflow for tracking memory leaks. My first thought would be to use the memory tool in dev tools, but comparing snapshots doesn't seem to be the easiest?

ui/src/contexts/events-context.tsx
```diff
@@ -44,7 +44,11 @@ export const useEvent = <T,>(
     }
 
     const parserListener = (ev: MessageEvent) => {
-      listenerAdded.current(JSON.parse(ev.data))
+      try {
+        listenerRef.current(JSON.parse(ev.data))
+      } catch (error) {
```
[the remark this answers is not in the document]
p1carats: I actually renamed `listenerAdded` to `listenerRef` to make the naming a bit more consistent and reflective of its purpose. And yeah, I wrapped it in a try/catch just to be safe — I didn’t run into a specific error as far as I remember, but I stumbled upon it by chance and figured it wouldn’t hurt to guard against any edge cases. Better safe than sorry!

p1carats:
Just pushed an update! I applied your suggestions regarding the deprecated pattern in the Plex Axios request — thanks again for catching that. I’ve added you as a co-author for that commit @benscobie 🙂

Also added a new commit with a few small fixes, mostly React minor warnings related to controlled vs uncontrolled form states.

Let me know if anything else needs tweaking!

benscobie:
> Full transparency: I’m running Maintainerr outside of Docker right now, in a pretty tightly constrained environment with a hard ulimit on memory. The app would mostly work, but I kept getting random crashes tied to some glib issue from a Node module. That’s what got me digging in the first place (and explains my contributions to the release-workflow branch too, haha).

> I’ve just been putting the app under memory pressure and watching how it behaves. 

Are you running the server + your client (browser) on the same memory-constrained system? The reason I ask as if you're only talking about the server-side, then I am surprised you saw any reduction in memory with the changes made as they're all targeting the client.

ui/src/components/Common/MediaCard/index.tsx
```diff
@@ -132,6 +132,7 @@ const MediaCard: React.FC<IMediaCard> = ({
               alt=""
               src={`https://image.tmdb.org/t/p/w300_and_h450_face${image}`}
               fill
+              sizes="9.375rem"
```
benscobie: How did you come up with the number? I'm looking at the generated srcset and I can't see any relation to this 😆 

I also noticed a problem with the CachedImage implementation. It is referencing settingsCtx which only gets populated on the settings pages by the looks of it... So this leads to unoptimized being true as it's defaulted to an empty object. Of course if you swap to a settings page and then back, it'll then reference the cached image instead.

I think we could move that fetch and set of settings to the layout so we can get access to the context everywhere. The other alternative is loading the settings within the CachedImage component, and use a fetch library that de-dupes requests like react-query (already a dependency). The settings provider should be moved to the root settings page (\ui\src\components\Settings\index.tsx) in this case though, as it's only used in that tree.

p1carats: It's the value from [ui/styles/globals.css](https://github.com/jorenn92/Maintainerr/blob/ec110b9116606a29726cce14abb63589acfed84f/ui/styles/globals.css#L87)! React was complaining about not having a default size for this. 

The current image optimization workflow is pretty intensive as far as I've seen (I've set image to unoptimize on next.config on my constrained dev env, squoosh (the lib used to optimize images on Next) uses huge WASM mallocs). This has basically eliminated all the crashs I could encounter. Maybe we could switch to a lighter engine if possible?

**Edit :** Next themselves [recommend switching to sharp instead of squoosh for production use](https://nextjs.org/docs/messages/sharp-missing-in-production), so this is definetely something to consider.

**Edit 2:** I saw multiple old commits about fixing sharp for prod use on Docker, but I couldn't find anything in the current code about it. I'll try to spin up a PR this weekend for that, although this will require moving off alpine-based images as sharp doesn't support musl and requires libgc.

p1carats: > It is referencing settingsCtx which only gets populated on the settings pages by the looks of it... So this leads to unoptimized being true as it's defaulted to an empty object. Of course if you swap to a settings page and then back, it'll then reference the cached image instead.
> 
> I think we could move that fetch and set of settings to the layout so we can get access to the context everywhere. The other alternative is loading the settings within the CachedImage component, and use a fetch library that de-dupes requests like react-query (already a dependency). The settings provider should be moved to the root settings page (\ui\src\components\Settings\index.tsx) in this case though, as it's only used in that tree.

Nice catch! Should I let you implement this, or do you want me to add another commit to this PR?

benscobie: I swear they've changed the sharp stuff several times now. I'm sure at some point they started shipping it as part of their package. It's the first time I've seen squoosh mentioned as well.

I've used sharp in the past for a production app but we ended up disabling image optimization entirely as it was hurting performance at times. You can certainly give it a go with your setup and see if it's any better than squoosh.

We can leave the cached image fix I mentioned for a different PR 😊

benscobie: Oh and on the sizes change, I'll need to check if that's a large enough value, as that CSS was using minmax. 1fr could be larger just before the next breakpoint hits when the cards are in a single column.

benscobie: I would probably go with the default of `100vw` here for now. As the component is referenced in a few places that contain them in a grid, with varying different breakpoints, I think it will be a bit of a pain (or potentially impossible) to generate a sensible list. The smallest I saw a card go was around 1/2 the size of the source image (which is 300x450). With `sizes=100vw`, I'm getting 24KB~ optimised files, with the src image being 37KB~

p1carats:
I've updated the size value to `100vw` as requested and force-pushed the changes. The PR looks good on my end!

benscobie (approved):
Thank you 🚀

PR #1943, opened 2025-09-29
Add per-rule group execution and UI controls

jcputney:
### Description

Introduces API and backend logic to execute individual rule groups, including endpoint, service changes, and tests. Adds a RunButton component and integrates execution controls into the RuleGroup UI, allowing users to trigger execution for specific rule groups and view execution status.

### Checklist

### How to test

1. Create a Rule
2. Go to the Rules page
3. You should see a new Run button next to the Edit button

### Additional context

After clicking the Run button on a Rule, all other Run buttons are disabled while the Rule is executing.

image


server/src/modules/rules/rules.controller.ts
```diff
@@ -102,6 +103,47 @@ export class RulesController {
...
+  @ApiResponse({
+    status: 409,
+    description: 'The rule executor is already running.',
+  })
+  async executeRuleGroup(@Param('id') id: string, @Res() res: Response) {
+    const ruleGroupId = Number(id);
+
+    if (Number.isNaN(ruleGroupId)) {
+      throw new NotFoundException('Rule group not found');
+    }
```
benscobie, line 125:
```suggestion
  async executeRuleGroup(@Param('id', ParseIntPipe) ruleGroupId: number, @Res() res: Response) {
```

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -97,7 +102,16 @@ const RuleGroup = (props: {
             </div>
           </div>
         </div>
-        <div className="grid w-full grid-cols-1 xl:grid-cols-2">
+        <div className="grid w-full grid-cols-1 xl:grid-cols-3">
+          <div>
+            <RunButton
```
benscobie, line 107: I'm not 100% sold on the Run button being so prominent. It's also in place of the Edit button so I suspect it'll be an annoyance for people until they learn that it's moved. Do you have any other ideas?

ui/src/components/Rules/RuleGroup/index.tsx
```diff
@@ -97,7 +102,16 @@ const RuleGroup = (props: {
...
+              onClick={() => props.onExecute(props.group)}
+              text="Run"
+              executing={props.isExecuting}
```
benscobie, line 110: I think this is supposed to make the run button show a spinner like the top level Run Rules button when executing? It's not appearing for me though.

server/src/modules/rules/tasks/rule-executor.service.ts
```diff
@@ -199,7 +233,9 @@ export class RuleExecutorService extends TaskBase {
 
     this.eventEmitter.emit(
```
benscobie, line 234: In the rare case getRuleGroupsForExecution throws, there would have been no RuleHandler_Started event prior to this Finished event now. I think we should move it into the catch above, what do you think?

benscobie:
Closing this as the rule executor was rewritten in #2148 and this functionality has been delivered in #2173.

PR #2367, opened 2026-02-12
refactor(logging, settings): parse settings using schemas for validation

ydkmlt84:
Summary
Fixes CI build failures (yarn turbo build) caused by DTO type errors in server services.

**What changed**

- Instead of directly reading properties from createZodDto classes, services now parse DTO inputs with the matching Zod schema.

### DIFF
--- a/apps/server/src/modules/logging/logs.service.ts
+++ b/apps/server/src/modules/logging/logs.service.ts
@@ -1,4 +1,4 @@
-import { LogSettingDto } from '@maintainerr/contracts';
+import { LogSettingDto, logSettingSchema } from '@maintainerr/contracts';
 import { Injectable, LoggerService, Scope } from '@nestjs/common';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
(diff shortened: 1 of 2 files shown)


enoch85:
@ydkmlt84 Hmm, this was never an issue before? Build always worked in the jellyfin-dev branch (yes, I know this is main).

> The PR papers over a TypeScript compilation issue by introducing unnecessary runtime work (double Zod parsing on every request). It "works" because schema.parse() returns a cleanly-typed plain object, satisfying the type checker — but the actual validation is wasted CPU cycles since ZodValidationPipe already did it. If the build was working before and something broke, the root cause should be investigated rather than adding redundant parsing throughout the service layer.

enoch85:
I will not implment this in main.

Based on research in changelogs from the updated dependencies and online zod / nestjs documentation:

> The nestjs-zod v5+ rewrite improved createZodDto type inference, but the root issue in this project is the tsconfig mismatch — contracts uses strict: true while server uses strictNullChecks: false. Even with v5.1.1, that mismatch causes the types to not resolve properly.
>
> Keep our z.infer approach. It's:
>
>The documented best practice per nestjs-zod and Zod communities
>Zero runtime cost (vs redundant .parse() calls)
>Already the pattern used for JellyfinSetting and PlexSetting in this codebase
Proper separation of concerns (DTOs for controllers, inferred types for services)

Would be great if you could cherry-pick this commit on jellyfin-dev and remove the Jellyfin settings, that way we are as even as possible: 

https://github.com/Maintainerr/Maintainerr/pull/2330/commits/25f496b31545fd7143b0169a6650af9010665b6d
https://github.com/Maintainerr/Maintainerr/pull/2330/commits/0d51e0ba1b6f99216cf6c4bfe424d3ed301a26ab
https://github.com/Maintainerr/Maintainerr/pull/2330/commits/fcdf6b831792a4707c10aa04e00d3e2d5603511a

ydkmlt84:
> @ydkmlt84 Yeah, but build is failing. I will sort it soon.

I am working on it as well. I will look through your changes to make sure we are as close as possible...to ease future merging.

PR #2442, opened 2026-03-01
fix(server): reject null/undefined in numeric rule comparisons

### LINKED ISSUE #2441: Plex IMDb Rating Often Not Being Populated
**Describe the bug**
I created a rule which actions on `Plex - IMDb rating (scale 1-10)`. I was surprised to see a bunch of highly rated movies being added. When I checked the details I would see results like this:

```
mediaServerId: "5633"
result: true
sectionResults:
  - id: 0
    result: true
    ruleResults:
      - action: before
        firstValueName: Plex - Date added
        firstValue: 2021-09-04T12:20:37.000Z
        secondValueName: custom_days
        secondValue: 2023-02-28T21:16:29.103Z
        result: true
  - id: 1
    result: true
    operator: AND
    ruleResults:
      - action: equals
        firstValueName: Plex - Times viewed
        firstValue: 0
        secondValueName: number
        secondValue: 0
        result: true
  - id: 2
    result: true
    operator: AND
    ruleResults:
      - action: smaller
        firstValueName: Plex - IMDb rating (scale 1-10)
        firstValue: null
        secondValueName: number
        secondValue: 6
        result: true
```

Note the `null` for the IMDb rating. When I view the metadata from plex by going to "Get Info" then "View XML" I see IMDb ratings. 

Not *all* of my matches had a `null` value but most of them in a spot check. Not sure how to check more broadly.

**To Reproduce**
Create a rule which checks Plex IMDb rating, then evaluate and see if it is properly enforced.

**Expected behavior**
I would have expected it to pull the IMDb rating from Plex metadata.

 - OS: Debian 12

enoch85:
## Summary

Fixes #2441 — Plex IMDb rating (and other nullable values) incorrectly matching `SMALLER`/`BIGGER` rules when the value is missing.

## Root Cause

Two issues in `rule.comparator.service.ts`:

1. **Broken null guard** — The expression `(firstVal !== undefined || null)` does not reject `null` due to JS operator precedence. Since getters return `null` for missing data, null values pass through to comparisons.

2. **Unsafe numeric coercion** — `BIGGER`/`SMALLER` use raw `val1 > val2` / `val1 < val2`. JS coerces `null` to `0`, so `null < 6` evaluates to `true`, causing false matches.

## Changes

- **`rule.comparator.service.ts`**: Fix the null guard to `firstVal != null` (rejects both null and undefined). Add `typeof` number checks in BIGGER/SMALLER (consistent with how BEFORE/AFTER already guard for Date types).
- **`rule.comparator.service.doRuleAction.spec.ts`**: Replace individual BIGGER/SMALLER tests with a single table-driven `numericComparisonData` array (follows existing `listCountData` pattern), covering normal cases and null/undefined edge cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,11 +48,11 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Docker meta
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: |
             ${{ env.DOCKERHUB_SLUG }}
@@ -60,7 +60,7 @@ jobs:
 
       - name: Log in to GitHub Container registry
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -68,14 +68,14 @@ jobs:
 
       - name: Log in to Docker Hub
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME  }}
           password: ${{ secrets.DOCKERHUB_TOKEN  }}
 
       - name: ${{ github.event_name == 'pull_request' && 'Build image' || 'Build & push by digest' }}
         id: build
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           context: .
           platforms: ${{ matrix.platform }}
@@ -115,24 +115,24 @@ jobs:
           merge-multiple: true
 
       - name: Log in to GitHub Container registry
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.repository_owner }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME  }}
           password: ${{ secrets.DOCKERHUB_TOKEN  }}
 
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v3
+        uses: docker/setup-buildx-action@v4
 
       - name: Docker meta
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: |
             ${{ env.DOCKERHUB_SLUG }}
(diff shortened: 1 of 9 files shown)


apps/server/src/modules/rules/tests/rule.comparator.service.doRuleAction.spec.ts
```diff
@@ -207,28 +207,35 @@ describe('RuleComparatorService', () => {
...
+      [false, 5, 5, RulePossibility.SMALLER],
+      // null/undefined must not match numeric comparisons
+      [false, null, 6, RulePossibility.SMALLER],
+      [false, null, 6, RulePossibility.BIGGER],
+      [false, 6, null, RulePossibility.SMALLER],
+      [false, 6, null, RulePossibility.BIGGER],
+      [false, undefined, 6, RulePossibility.SMALLER],
+      [false, undefined, 6, RulePossibility.BIGGER],
+      [false, null, null, RulePossibility.SMALLER],
+      [false, null, null, RulePossibility.BIGGER],
```
SynVisions: I know it makes this change more complicated, but to be honest I don't think that shifting it to "always false" is *that* much better than "always true". Depending upon how one writes their rules this will be just as insidious. If you have a bug or dependency failure users will potentially experience mass deletion due to unexpected bypasses in logic checks. IMO a few options that make sense to me as a user:

1. If you are getting non-numbers in numerical comparisons, you should raise an error and fail to evaluate the rule group on that entry entirely.
1. Users could be given the choice in rule definition (or rule group definition) to default null values to something.

Realistically I think the first option is the most "logical" and simple, but the maintainers could decide if they want to add more optionality here.

enoch85: I had another look at it, thanks for input!

**CHANGES**

- rule.comparator.service.ts:212-219 — The null guard now lets null/undefined through for BIGGER/SMALLER actions. In v2.27.0, the guard was broken (x !== undefined || null is always truthy), so null always reached doRuleAction. We preserve that behavior for numeric comparisons while keeping the fixed guard for all other comparison types.

- rule.comparator.service.ts:428-449 — BIGGER/SMALLER now use Number(val1) / Number(val2) instead of returning false for non-numbers. This matches v2.27.0's implicit JS coercion (null → 0, undefined → NaN). A debug log is emitted whenever a non-number value is encountered, so users/maintainers can spot these cases in logs.

- Tests updated to match v2.27.0 behavior: null acts as 0 (so null < 6 is true), undefined produces NaN (all comparisons false).

- Log as WARN so that it's visible when we hit a null value.

ydkmlt84: I am in agreement with this comment.

> If you are getting non-numbers in numerical comparisons, you should raise an error and fail to evaluate the rule group on that entry entirely.

If for some reason NULL is returned from a service, we should not use that value to do anything. It should never get evaluated and compared to secondValue (user input).

This is not 100% always the case but in MOST instances, NULL is coming from some sort of communication error.  We did not receive the correct information from the services API, or any information depending on what it was.

enoch85:
@ydkmlt84 I've been on this issue now for the past 6 hours. It became a bigger than I initally had planned for, but it also makes the code leaner, and pushes more code to the shared abstraction layer, which to me is a good thing. The more we can share between the media servers, the better.

I've tested in my prod, and it behaves as expected. I have nothing to add here now. Consider this PR to be done. :+1:

image

PR #2453, opened 2026-03-03
fix(server): Use shared watch state for isWatched and Plex viewCount

wortmanb:
## Summary

Addresses reviewer feedback from the original PR. The previous approach replaced the watch history API entirely with `libItem.viewCount`, which is per-user (admin token only) and not suitable for server-wide view counts.

### viewCount (revised approach)

- **Primary**: watch history API (`/status/sessions/history/all`) — server-wide, counts all users
- **Fallback**: `libItem.viewCount` when history returns 0 or the API fails — per-user (admin), but better than returning 0 when Plex has purged watch history
- Debug logging when fallback is used, noting the per-user limitation

| Scenario | Before (original code) | After (this PR) |
|----------|----------------------|-----------------|
| History available | Uses history count | Uses history count (same) |
| History purged | Returns 0 (wrong) | Falls back to admin viewCount with warning |
| API failure | Uncaught error | Falls back to admin viewCount with warning |

### isWatched (new property)

New boolean rule option: **Plex - Is Watched** (id: 43, movie type).

Returns `true` if the item has been watched by anyone:
1. Checks watch history first (server-wide)
2. Falls back to `libItem.viewCount > 0` if history is empty/fails

This provides a simple "has this been watched?" check that is resilient to Plex history purging.

### Re: Jellyfin parity (enoch85)

Noted that Jellyfin viewCount is derived from aggregated watch history, not a native counter. Aligning Plex and Jellyfin semantics is worthwhile but orthogonal to this fix — suggest addressing in a follow-up.

### Re: remove/re-add behavior (ydkmlt84)

When an item is removed from Plex and re-added, the watch history is lost and `viewCount` resets to 0. Neither the history API nor the native counter survives this. The `isWatched` boolean has the same limitation. This is a Plex-level constraint.

## Test plan

- [x] 10 unit tests covering viewCount (5 cases) and isWatched (5 cases): history available, history empty with fallback, history empty without fallback, API failure with fallback, API failure without fallback
- [x] All 510 existing tests pass (0 regressions)
- [x] Formatted with Prettier


wortmanb:
I'll look into addressing these issues today. I appreciate the detailed comments.

wortmanb:
Updated the PR based on the review feedback. Here's what changed:

**@ydkmlt84** — You were right that the original approach (replacing watch history with `libItem.viewCount`) doesn't work for multi-user servers since it's per-admin-token. The revised approach:

- Restores watch history API as the primary data source (server-wide)
- Adds `libItem.viewCount` as a *fallback only* when history returns 0 or the API fails, with debug logging that notes the per-user limitation
- This is strictly better than the original code which had no fallback at all

On the remove/re-add question: both watch history and native viewCount reset when an item is removed and re-added to Plex. That's a Plex-level constraint neither approach can work around.

**@enoch85** — Agreed on Jellyfin parity. The Jellyfin viewCount derives from aggregated watch history rather than a native counter, so the semantics already differ. Aligning them by intent would be a good follow-up but felt out of scope for this fix.

**New addition**: Added a `Plex - Is Watched` boolean property (id: 43) that returns true if watch history has entries OR `libItem.viewCount > 0`. Provides a simple "has this been watched by anyone?" check that's resilient to history purging.

All 510 tests pass, 10 new tests covering both viewCount and isWatched.

enoch85:
@ydkmlt84 

> Note to self: https://github.com/Maintainerr/Maintainerr/pull/2466#issuecomment-4021846607

The last two commits finish the Plex-side fix from that comment. The functional change removes the fallback to metadata viewCount and uses the item-scoped Plex history endpoint directly, with cache bypassed for watched-rule evaluation. That means the rule no longer depends on the admin-scoped viewCount field and is not stale for up to 5 minutes.

The newest commit is test-only. It extracts shared Plex fixtures so the new history-based behavior is covered more clearly and with less duplicated setup.

PR #2458, opened 2026-03-04
feat: clean up empty Sonarr shows after season actions

### LINKED ISSUE #892: TV shows are missing from Maintainerr when they are monitored but have no files
I currently have a rule setup that will automatically delete and unmonitor a season of a TV show if no one watches any episodes from that season for 90 days. This functionality works perfectly, no issues there.

However, in my circumstance, TV shows can get into a state where they are monitored in Sonarr, but none of the seasons are monitored and there are no files (because my previous rule has deleted and unmonitored the seasons).

If the show is a continuing show, it's not a problem because chances are my users will want to watch future seasons, so keeping it within Sonarr is useful because it'll grab those new episodes when they are available.

The issue I am running into is when shows have finished and have an Ended state in Sonarr. This leaves the show itself monitored in Sonarr, no seasons monitored, no episodes downloaded and no future episodes coming. In this case, I'd like to just have a rule that would delete the entire show from Sonarr since my users don't care about the show any more.

This seems pretty straight forward, however I noticed that shows that don't exist in Plex (which in my circumstance they won't any more, because the files are deleted) aren't interactable in the Maintainerr overview, or via Maintainerr rules/collections.

I have tried the following rules and none of them pickup the shows in that state:

```
mediaType: SHOWS
rules:
  - "0":
      - firstValue: Sonarr.ended
        action: EQUALS
        customValue:
          type: boolean
          value: "true"
      - operator: AND
        firstValue: Plex.sw_episodes
        action: EQUALS
        customValue:
          type: number
          value: 0
```

```
mediaType: SHOWS
rules:
  - "0":
      - firstValue: Sonarr.ended
        action: EQUALS
        customValue:
          type: boolean
          value: "true"
      - operator: AND
        firstValue: Sonarr.diskSizeEntireShow
        action: EQUALS
        customValue:
          type: number
          value: 0
```

Here are some screenshots the show's state in Sonarr.

image
image

image
image

I am running Maintainerr on Docker, and my current version is v2.0.1.

enoch85:
## Summary

Closes #892

- adds two season-level Sonarr actions:
  - Unmonitor and delete season + delete show if empty
  - Unmonitor season + unmonitor show if empty
- deletes ended empty shows immediately after season cleanup
- for continuing shows, checks Seerr for remaining requested seasons and only deletes when none remain
- the delete-show-if-empty path inspects Seerr state instead of deleting Seerr requests, preserving Seerr history and audit trail
- delete-show-if-empty is not gated by Force Seerr; for continuing shows it only needs Seerr to be configured and reachable to confirm that no other season requests remain
- moves ServarrAction into shared contracts and updates the UI to use the shared enum
- adds focused tests for Sonarr empty-show cleanup and Seerr request filtering

## Test plan

- [ ] Verify ended shows are deleted when the last season is removed with delete-show-if-empty
- [ ] Verify a continuing show in the reviewer scenario (for example, only one requested season remains and Sonarr is not monitoring future seasons by default) is deleted when that last requested season is removed
- [ ] Verify continuing shows are deleted only when Seerr shows no other requested seasons remain
- [ ] Verify continuing shows are not deleted when Seerr still has another requested season
- [ ] Verify continuing shows are not deleted when Seerr state is unavailable or Seerr is unreachable
- [ ] Verify delete-show-if-empty does not remove Seerr requests or request history
- [ ] Verify delete-show-if-empty still works without Force Seerr enabled
- [ ] Verify unmonitor-show-if-empty only unmonitors ended shows with no monitored seasons left
- [ ] Verify existing delete and unmonitor actions still behave as before


apps/server/src/modules/api/seerr-api/seerr-api.service.ts
```diff
@@ -296,6 +296,33 @@ export class SeerrApiService {
...
+        ),
+      );
+    } catch (err) {
+      this.logger.warn(
+        'Seerr communication failed. Is the application running?',
+      );
+      this.logger.debug(err);
+      return undefined;
+    }
+  }
```
andrew-kennedy, line 352: Here's some alternate pseudocode to get at what I mean:


```ts
public async hasRemainingSeasonRequests(
  tmdbid: string | number,
  removedSeasonNumber: number,
): Promise<boolean | undefined> {
  try {
    const media = await this.getShow(tmdbid);

    if (!media?.mediaInfo) {
      return undefined;
    }

    const requests = media.mediaInfo.requests ?? [];

    return requests
      .filter((request) => request.status !== MediaRequestStatus.DELETED)
      .flatMap((request) => request.seasons)
      .some((season) => season.seasonNumber !== removedSeasonNumber);
  } catch (err) {
    this.logger.warn(
      'Seerr communication failed. Is the application running?',
    );
    this.logger.debug(err);
    return undefined;
  }
}
```

Your original code means:

 "Is there any request that contains at least one season other than `removedSeasonNumber`?"

but this rewrite proposal means:

"Is there any whole request left that does not contain `removedSeasonNumber` at all?"


Put another way, what we really want is:

"Among non-deleted requests for this show, does any request still contain any season whose number is not `removedSeasonNumber`?"

This would mean that we don't need to have Maintainerr mess with deleting Seerr requests itself at all, and we maintain the "audit trail" of cleaned up, deleted requests in the Seerr UI.

enoch85: Thanks for the review. Reverted the old commit and refactored a bit further. Please check it it's aligning wit your intentions.

andrew-kennedy (commented):
This looks good to me when it comes to non-ended shows, but I had a question about ended shows: why do we care if they are ended at all? 

Imagine I requested season 1 of breaking bad, an ended show. Someone else requested season 2, but for whatever reason (indexer rate limits, etc) that request hasn't been fulfilled yet. My maintainerr rule eventually gets triggered to delete season 1. Why should it matter if the show is ended? Shouldn't deleting the show only happen regardless of if it's "ended" if two things are *both* true:
1. Show is empty of seasons in sonarr.
2. No unfulfilled seasons besides the current one being acted on exist in any seerr requests.


I suspect we don't need to take into account ended status at all.

enoch85:
> Any chance I can help further towards getting this merged @enoch85? Would really smooth out a rough edge I experience a lot with Maintainerr leaving a lot of empty shows today.

I want this as much as you. But I don't control the main repo, only jellyfin-dev.

So use 'jellyfin-dev' until it gets merged into main. I run it myself in production.

Cc @ydkmlt84 as we discussed already.

apps/server/src/modules/actions/sonarr-action-handler.ts
```diff
@@ -242,4 +293,91 @@ export class SonarrActionHandler {
...
+
+    if (!series || !this.isShowEmpty(series, 'files')) return;
+
+    if (series.status === 'ended') {
+      await sonarrApiClient.deleteShow(series.id, true, listExclusions);
+      this.logger.log(
+        `[Sonarr] Show '${series.title}' is ended with no files remaining — deleted from Sonarr`,
+      );
+      return;
+    }
```
andrew-kennedy, line 319: This will lead to ended shows getting deleted if they have another in-flight request for another season. 

Here's the scenario:
1. Someone requests season 1 of an ended show
2. Season 1 Downloads
3. Someone requests season 2 of the same ended show
4. Due to some unknown issue, Season 2 does not yet dowload
5. The rule runs and unmonitors and deletes season 1 in sonarr as all criteria are met
6. This function runs as the show's files are "empty", and it deletes the show

In this case, there was still an outstanding unfulfilled request for season 2 of the show, and this function deletes the show before it can ever be filled. We still need to check Seerr or the monitored status of other seasons in sonarr to ensure we don't delete the show when pending requests still exist. Preferably seerr, as that's the true source of truth.

andrew-kennedy (changes requested):
Currently `deleteShowIfEmpty` is too aggressive in deleting shows. Left a comment explaining the issue and how to fix.

apps/server/src/modules/actions/sonarr-action-handler.ts
enoch85: Great catch!

Feel free to dig in to this PR. Help is welcome! :rocket:

enoch85:
Superseded by #2618. Rebuilt from current development so the valid fixes and review intent could be carried forward cleanly.

PR #2461, opened 2026-03-05
feat(rules): add ARR disk target path selection for disk space rules

tumeden:
## Summary
This PR adds path targeting for Radarr/Sonarr disk space rules.

- add a **Disk Target** dropdown for Radarr/Sonarr disk space rules
- the dropdown appears **only when the first value is a Radarr/Sonarr disk space field**
- support **Aggregate** (default) or a specific Arr-reported path
- evaluate disk space against the selected path when configured
- keep existing rules unchanged (Aggregate mode by default)

## Why
Users running containerized Arr setups need to evaluate free space on media mounts (for example `/movies`, `/shows`) instead of only using a generic aggregate value.

Feature request:
https://features.maintainerr.info/posts/151/disk-space-on-other-mounts-not-just-root-containerized-radarr-sonarr-deployments

## Notes
- Backward compatible: existing rules continue to work without changes.
- Path options come from Arr disk space API results.

image

I did basic testing for these changes. I don’t currently run Sonarr/Radarr in Docker, so I couldn’t directly verify the Docker-specific mount-path scenario but this change adds more granular configuration when it comes to disk space rules.

### DIFF
--- a/apps/ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
+++ b/apps/ui/src/components/Rules/Rule/RuleCreator/RuleInput/index.tsx
@@ -474,10 +474,10 @@ const RuleInput = (props: IRuleInput) => {
   }, [firstval])
 
   useEffect(() => {
-    if (!isSelectedArrDiskspaceRule && arrDiskPath) {
+    if (!isSelectedArrDiskspaceRule) {
       setArrDiskPath('')
     }
-  }, [isSelectedArrDiskspaceRule, arrDiskPath])
+  }, [isSelectedArrDiskspaceRule])
 
   useEffect(() => {
     if (secondVal) {


enoch85:
This is now included in the `jellyfin-dev` docker container - among some other fixes. To upgrade, change tag in docker from `latest` (or whatever you have) to `jellyfin-dev`. Then do:

```
docker compose pull jellyfin-dev && docker compose up -d
```

You can check the latest commits here: https://github.com/Maintainerr/Maintainerr/commits/jellyfin-dev

1. Consider the jellyfin-dev branch to be early development, and make a backup before switching to that branch!
2. Please test that your issue is fixed
3. Report back here, and tag me and @ydkmlt84 

Thank you very much! :rocket:

azurecrash:
Just pulled this down to test. 


The fix does work for Radarr, but not Sonarr, and I can now show the exact asymmetry between the two which may be helpful for you all to be aware of, though I don't know if anything further can happen on this side.

Environment:

- Radarr 6.1.1.10317 (linuxserver.io, develop tag, Docker)
- Sonarr 4.0.16.2946 (linuxserver.io, develop tag, Docker)
- Both running on the same host, both with NAS-backed media paths mounted via bind mount
- Both have their respective media paths configured as root folders in Media Management

Radarr (working):

System Status → Disk Space correctly shows /movies at 6.9 TiB free / 26.2 TiB total — the NAS filesystem, not the container root.

Sonarr (broken):

System Status → Disk Space only lists /, /config, and /downloads, all resolving to the container root (232.2 GiB). /tv does not appear, despite being correctly mounted and showing 6.9 TiB free in Sonarr's own Root Folders settings under Media Management.

Root cause hypothesis:
This appears to be a difference in how the two apps expose paths via their APIs. Radarr surfaces /movies through the Disk Space endpoint; Sonarr does not surface /tv there — it only appears via the Root Folders endpoint. I am unclear whether this is a function of how the container images have been put together or Radarr and Sonarr differences at the application level. Maybe someone else can weigh in here?

PR #2488, opened 2026-03-13
fix(server): improve Jellyfin watch data aggregation

### LINKED ISSUE #2427: Add result caching to getItemFavoritedBy and getTotalPlayCount
## Summary

`getWatchHistory()` caches its processed results to avoid repeated per-user API calls across rule runs. `getItemFavoritedBy()` (#2393) and `getTotalPlayCount()` use the same pattern — iterating all users via `getItemUserData()` in batches — but neither caches their results.

For a rule group scanning N items with M users, each uncached method adds N × M API calls per rule run.

## Why per-method caching (not a shared `getItemUserData` cache)

All three methods call `getItemUserData()` internally, so caching at that level seems appealing. However, downstream consumers of `getWatchHistory()` depend on the **processed `WatchRecord[]` format** — `Date` objects parsed from ISO strings, filtering by `Played: true`, and defaults applied via `JellyfinMapper.toWatchRecord()`. A raw `getItemUserData()` cache would not eliminate the need for per-method caching since the transformation would still need to run on every call.

Each method should cache its own processed results, following the existing `getWatchHistory()` pattern:

```typescript
// Existing pattern in getWatchHistory()
const cacheKey = `${JELLYFIN_CACHE_KEYS.WATCH_HISTORY}:${itemId}`;
if (this.cache.data.has(cacheKey)) {
  return this.cache.data.get<WatchRecord[]>(cacheKey) || [];
}
// ... compute ...
this.cache.data.set(cacheKey, records, JELLYFIN_CACHE_TTL.WATCH_HISTORY);

### LINKED ISSUE #2486: jellyfin lastWatched is null on unwatched items
**Describe the bug**
When using the jellyfin "last watched"/"last view date" rule, if a show hasn't been watched, the date is null and the custom days become 864000:
- operator: AND
          action: before
          firstValueName: Jellyfin - Last view date
          firstValue: null
          secondValueName: custom_days
          secondValue: 864000
          result: false

**To Reproduce**
1. add a show rule for jellyfin. specifically a last view date rule
2. set the second value to custom days before 10 days
3. save
4. Go to go to collections
5. Click on a show collection
6. Click on "test media"
7. write a show nobody on your server has ever watched
8. note that it doesn't have a working lastViewDate**Expected behavior**A clear and concise description of what you expected to happen.

**Screenshots**

 - OS: unraid (docker
 - Version: 3.1.0

**Additional context**

enoch85:
### Description & Design

This switches Jellyfin per-user item lookups to the dedicated user-data endpoint, centralizes the shared batching logic, and adds caching for favorites and total play count.

This fixes incorrect show `lastViewedAt` aggregation in Jellyfin and reduces repeated per-user API work in the same adapter path.

### Related issue

Fixes #2486
Fixes #2427

### Checklist

### How to test

1. Run `yarn format`.
2. Run `yarn test`.
3. Verify a Jellyfin show rule using `lastViewedAt` picks up the latest watched episode date.


enoch85:
Tested and works! Manual validation against Jellyfin looks good:
- Unwatched shows still return `lastViewedAt = null` and are skipped by BEFORE/AFTER comparisons.
- Watched shows are now matched correctly by show-level `Jellyfin.lastViewedAt` rules, which confirms episode watch data is being aggregated into the show value.

PR #2492, opened 2026-03-15
Fix Sonarr episode handling when media indexes are missing

enoch85:
Fix https://github.com/Maintainerr/Maintainerr/issues/2491

## Summary
- fix Sonarr episode handling for items that do not have a usable episode index by falling back to air date matching
- make Sonarr and Radarr actions report success and failure explicitly instead of allowing false positives
- only remove media from Maintainerr collections after the external action and any forced Seerr cleanup succeed
- fix collection worker reporting so failed items are not reported as successfully handled
- add regression coverage for episode matching, failure propagation, and collection handling order

This issue started as a Sonarr episode identification bug, but review of the full handling flow uncovered a second correctness problem in the same path: Maintainerr could mark media as handled and remove it from the collection before external actions had actually succeeded.

Fixing the issue safely required correcting both:
- how Sonarr identifies certain episodes
- how success and failure are propagated through the Servarr helpers, action handlers, collection handler, and worker flow

That is why the PR touches more than the original Sonarr lookup point.

## Testing
- yarn workspace @maintainerr/server test

### DIFF
--- a/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.ts
+++ b/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.ts
@@ -107,9 +107,7 @@ export class RadarrApi extends ServarrApi<{ movieId: number }> {
       if (options?.monitored !== undefined) {
         movieData.monitored = options.monitored;
       }
-      if (
-        !(await this.runPut(`movie/${movieId}`, JSON.stringify(movieData)))
-      ) {
+      if (!(await this.runPut(`movie/${movieId}`, JSON.stringify(movieData)))) {
         return false;
       }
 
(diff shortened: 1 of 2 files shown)


enoch85:
Superseded by #2653.\n\nThe valid Sonarr fixes from this branch were ported onto current \ in a smaller follow-up PR that preserves the newer metadata-based lookup and current collection-handling behavior.\n\nPlease continue review on #2653 instead of this stale branch.

enoch85:
Superseded by #2653.

The valid Sonarr fixes from this branch were ported onto current development in a smaller follow-up PR that preserves the newer metadata-based lookup and current collection-handling behavior.

Please continue review on #2653 instead of this stale branch.

PR #2494, opened 2026-03-15
fix: added batch collection adding of media for Jellyfin

### LINKED ISSUE #2489: Jellyfin collection sync can trigger collection.xml file lock errors
### Problem

When Maintainerr syncs Jellyfin collections, it adds and removes collection members one item at a time. Jellyfin appears to persist collection metadata after each mutation, which can cause repeated writes to the same `collection.xml` file in quick succession.

In practice this shows up as Jellyfin errors like:

System.IO.IOException: The process cannot access the file '.../collections/TEST [boxset]/collection.xml' because it is being used by another process.

### Current behavior

Maintainerr collection sync is sequential, but still performs multiple consecutive remote mutations for the same collection:

- resync existing items into a newly created collection
- add new items one by one
- remove items one by one

Relevant code paths:
- Collections add flow: [apps/server/src/modules/collections/collections.service.ts](apps/server/src/modules/collections/collections.service.ts#L736)
- Collections remove flow: [apps/server/src/modules/collections/collections.service.ts](apps/server/src/modules/collections/collections.service.ts#L860)
- Resync loop after collection creation: [apps/server/src/modules/collections/collections.service.ts](apps/server/src/modules/collections/collections.service.ts#L807)
- Per-item add call: [apps/server/src/modules/collections/collections.service.ts](apps/server/src/modules/collections/collections.service.ts#L1084)
- Per-item remove call: [apps/server/src/modules/collections/collections.service.ts](apps/server/src/modules/collections/collections.service.ts#L1161)

The Jellyfin adapter already uses endpoints that accept an `ids` array:
- add: [apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.ts](apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.ts#L1005)
- remove: [apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.ts](apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.ts#L1025)

But the media-server abstraction only exposes single-item add/remove methods:
- [apps/server/src/modules/api/media-server/media-server.interface.ts](apps/server/src/modules/api/media-server/media-server.interface.ts#L183)

### Root cause hypothesis

This is likely not a Maintainerr parallelism bug. The collection sync code is already sequential.

The likely issue is that we force Jellyfin to perform one collection mutation per item, and Jellyfin persists boxset metadata after each mutation. That creates write contention around `collection.xml` during larger syncs.

### Proposed fix

Batch Jellyfin collection mutations per sync pass instead of issuing one request per item.

Suggested implementation:

1. Extend the media-server abstraction with bulk collection mutation methods, or add a Jellyfin-specific batching path.
2. For Jellyfin, send all ids for a given add/remove operation in one API call.
3. Update `CollectionsService` to gather item ids for:
   - initial resync into a newly created collection
   - additions
   - removals
4. Keep database writes and collection log records per item after the remote batch succeeds.
5. If needed, fall back to per-item mutation only when a bulk call fails and we need best-effort recovery.

### Acceptance criteria

- Jellyfin collection sync no longer performs one remote add/remove request per item when syncing the same collection.
- Reproducing a multi-item collection sync no longer triggers repeated `collection.xml` file lock errors in Jellyfin logs, or reduces them materially.
- Plex behavior remains unchanged.
- Collection DB rows and collection log records remain correct after bulk add/remove operations.

AlexS54:
### Description & Design

This fixes potential Jellyfin errors caused by collection mutations requested by Maintainerr in quick succession.
The solution implemented here is taking advantage of Jellyfin's support for sending multiple media ids at the same time in one collection creation/media addition/media removal request.

The new batch methods are exposed in the media server abstractization and leverage the support described above for the Jellyfin implementation, and makes sequential calls in the Plex implementation.

This PR introduces a new `MediaServerFeature`: `COLLECTION_CREATION_WITH_ITEMS`, which represents whether the Media Server supports sending a list of media ids to be automatically added on collection creation.

### Related issue

Fixes #2489 

### AI-Assisted Development

I used Copilot for tab autocomplete in a few parts of the code, but I read each suggestion before accepting it. Did not have the AI do any complex/agent work.

### Checklist

### How to test

1. Set up a Jellyfin server with multiple media items and connect it to the Maintainerr instance.
2. Create a rule that matches multiple media items from Jellyfin and run it so a collection is created in Jellyfin.
3. Delete the collection IN Jellyfin (not in Maintainerr).
4. In Maintainerr, add a new media item to the collection.
5. The `Syncing ${collectionMedia.length} existing items to newly created media server collection` message will not be displayed, but the collection will be recreated properly with all the items in Jellyfin.


apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
```diff
@@ -277,6 +286,15 @@ export class PlexAdapterService implements IMediaServerService {
...
 
+  async removeBatchFromCollection(
+    collectionId: string,
+    itemIds: string[],
+  ): Promise<void> {
+    for (const itemId of itemIds) {
+      await this.removeFromCollection(collectionId, itemId);
+    }
+  }
+
```
AlexS54, line 317: Right now the batch methods throw in the Plex adapter on *any* failed operation, I don't know if we should do this another way

enoch85: This should be fixed now.

Instead of throwing on the first Plex failure, the batch methods now return the failed item IDs. That lets the collection service handle partial success properly, while still treating Plex `404` removes as success and Jellyfin failures as all-or-nothing.

enoch85 (dismissed):

enoch85 (approved):

PR #2534, opened 2026-03-26
Finalize Jellyfin hardening for collections, search, and setup

enoch85:
### Why

This branch started as a fix for a few Jellyfin stability issues, then was extended to close the still-valid gaps that remained after the Jellyfin merge into 3.0.

### What this PR fixes

- scopes Jellyfin collection lookup to the requested library instead of returning cross-library collections
- validates collection endpoint payloads with Zod before they reach the service layer
- avoids implicitly selecting a media server when both Plex and Jellyfin are configured without an explicit `media_server_type`
- makes Jellyfin search results hydrate through real metadata instead of returning fabricated partial media items
- fails fast when a media server adapter still is not set up after `initialize()` instead of returning a broken adapter
- adds focused regression coverage for the affected Jellyfin and factory behaviors

### Why this is better

This makes Jellyfin behavior more predictable in the places that matter most:

- collection operations stay library-correct
- invalid collection requests fail at the API boundary
- mixed Plex/Jellyfin configs no longer produce unsafe implicit selection
- search results now carry real provider IDs and metadata instead of placeholders
- adapter startup failures surface immediately instead of leaking into later request handling

### Validation

- targeted Jellyfin adapter and media server factory specs passed
- full server test suite passed: 39 suites, 651 tests

This PR targets `jellyfin-dev` and is intended to be the final Jellyfin hardening pass for the remaining verified issues in that branch line.


enoch85:
Closing in favor of #2617, which carries forward the remaining valid hardening fixes from this branch against current development. The stale or broader feature work from #2534 was intentionally left out so the replacement PR stays narrowly scoped and reviewable.

PR #2616, opened 2026-04-06
fix: track collection membership provenance

### LINKED ISSUE #2611: reused manual collection ui bug
**Describe the bug**

when using a "manual collection" with two+ rules adding media to the same collection, the media in rule 1 shows in the collection for rule 2, but with the "manual" tag implying it was manually included

**To Reproduce**
1. create a custom collection in Jellyfin (untested in plex)
2. create two separate rules (different libraries for example) that both add media to the same custom collection
3. open the maintainer "collections view" 
4. open one of the rule based collections
5. see that the media from rule 1 shows in collection 2 and media from rule 2 show in collection 1 

**Expected behavior**
I expected the collections view on the maintainerr side to only show the media that belongs to rule 1 in collection 1. (and the items that were actually manually created, but I'm unsure of how that should be done best)

**Screenshots**
my collections;
Image

the "Leaving Media (manual)" collection from rule Foreign Movies:

Image

 - OS: [e.g. Ubuntu] unraid (docker)
 - Version [e.g. 1.2.0] 3.3.0
 - Server: [e.g. Jellyfin/Plex] jellyfin

**Debug logs**
```
Add your debug logs here
```
don't think logs say anything sadly. this is simply a ui issue

**Additional context**
The solution is likely to base what's shown in the collection on the library the rule is built on. it wouldn't be perfect in case someone had two separate rules for the same library, but should at least be good enough maybe?

(and some way to properly track what is and isn't manually added)

enoch85:
## Summary

Fixes #2611 by tracking rule-owned and manual collection membership separately. When multiple rules target the same manual media server collection, each Maintainerr collection now keeps its own rule-owned membership while still showing genuinely shared manual items.

### Root cause

The old model stored a single `isManual` flag per `CollectionMedia` row. During media server sync, items discovered in a shared server collection but not tracked locally were imported as manual, even when they actually belonged to another rule targeting the same media server collection. That erased provenance and caused unrelated items to appear permanently pinned as manual.

### What this changes

- Replaces the single `isManual` flag with two independent fields:
  - `includedByRule`
  - `manualMembershipSource` (`legacy`, `local`, `shared`)
- Preserves rule and manual membership independently so one can be removed without deleting the other.
- Reconciles shared manual collections without re-importing rule-owned sibling items as manual.
- Preserves local and legacy manual provenance during sync instead of collapsing everything into shared membership.
- Applies scoped membership removal (`rule`, `manual`, `all`) so rule execution and media server sync only clear the intended membership.

### Backward compatibility

`isManual` remains a derived field via entity hooks, so existing read paths continue to behave correctly while the new provenance fields drive collection membership behavior. The migration backfills `includedByRule` and `manualMembershipSource` from existing rows and remains reversible.

## Validation

- `yarn format:changed`
- `yarn check-types`
- `yarn lint` (4 existing UI warnings only)
- `yarn test`
- `yarn build`


Simon-Eklundh:
I gave this a quick test with my simple ruleset, and it looks correct. will test with my complicated ruleset and give the code a proper onceover (eventually)

edit:

complicated ruleset technically works, but because of the bug with media item resolution in main, it won't fill any show related collections

enoch85:
> I gave this a quick test with my simple ruleset, and it looks correct. will test with my complicated ruleset and give the code a proper onceover (eventually)
> 
> edit:
> 
> complicated ruleset technically works, but because of the bug with media item resolution in main, it won't fill any show related collections

Ok, so it works - it's just that your bug is stopping it from working - but it would work on a system without your specific bug?

Simon-Eklundh:
It should work, yes.

My bug is already in main, so this shouldn't make anything worse

Mind you, I haven't reviewed the actual code

enoch85:
> It should work, yes.
> 
> My bug is already in main, so this shouldn't make anything worse
> 
> Mind you, I haven't reviewed the actual code

Please do a real code review. Would help a lot!

This will be in 3.5.0 anyway, so no rush. :)

Simon-Eklundh:
Will do (eventually, probably not tonight)


Will test with my advanced rules this evening Tho (as soon as my libraries are done refreshing metadata)

enoch85:
> Will do (eventually, probably not tonight)
> 
> Will test with my advanced rules this evening Tho (as soon as my libraries are done refreshing metadata)

Any update here? Looking to merge to development this weekend.

Simon-Eklundh:
I believe I tested it

But I'll want to run a second test since I don't remember the result etc properly...

Simon-Eklundh:
And I got through some of the code review, just got stuck on other things like sleep ;)

Simon-Eklundh:
@enoch85 
image


seems to be an issue on latest version

it worked before tho, so maybe a bad merge.

I've tested with multiple manual collections. it seems to be specifically a problem with manual collections


EDIT: checked out oldest commit and it seems I was wrong. it fails on that too.... 

Dug further into this; it seems to be a bug on main. somewhere between current main and 40 commits back

Simon-Eklundh:
https://github.com/Maintainerr/Maintainerr/pull/2657

this is a combination of both the fixes I did for collections being deleted, and what happens here.

enoch85:
Superseded by #2663 (https://github.com/Maintainerr/Maintainerr/pull/2663) after renaming the branch from poc/collection-membership-provenance to fix/collection-membership-provenance. Closing this PR so review continues on the renamed branch PR.

PR #2618, opened 2026-04-07
feat: clean up empty Sonarr shows after season actions

### LINKED ISSUE #892: TV shows are missing from Maintainerr when they are monitored but have no files

enoch85:
## Summary

Closes #892
Supersedes #2458

- adds two season-level Sonarr actions:
  - Unmonitor and delete season + delete show if empty
  - Unmonitor season + unmonitor show if empty
- rebuilds the empty-show cleanup on top of current development and the newer metadata lookup flow
- only deletes the show when Seerr explicitly reports that no other requested seasons remain, preserving the existing fail-safe behavior on ambiguous or failed Seerr lookups
- falls back to the ended-and-no-monitored-seasons case when Seerr is not configured
- excludes the delete-show-if-empty action from Force Seerr request mutation
- consolidates `ServarrAction` onto the shared contracts enum and re-exports it from the existing server import path
- simplifies Sonarr action handling by giving the season-only empty-show actions their own switch branches
- adds focused Sonarr, Seerr, and collection-handler tests


enoch85:
@andrew-kennedy this is the fresh rebuild of the still-valid carryovers from #2458 on top of current development. I kept the empty-show delete path conservative around remaining Seerr season requests, including the ended-show review scenario you called out.

enoch85:
This is double checked now. All pass!

enoch85:
@andrew-kennedy Please have a last look at this.

I'm happy, but would be nice with some confirmation. :+1:

enoch85:
I've been over this several times now. It checks out.

I will merge to development if you want to try it out even further. :+1: This will be in 3.5.0.

PR #2658, opened 2026-04-11
fix: retain items from different libraries in jellyfin

Simon-Eklundh:
- Two libraries share a manual collection in Jellyfin
- One of the libraries' rules is changed (e.g. manual → automatic)
- Before fix: the entire Jellyfin collection is deleted — including items from the other library
- After fix: only items from the changed library's rule are removed, the collection remains if it is manual, and automatic collections are deleted when they become empty


apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
```diff
@@ -167,6 +167,18 @@ export class PlexAdapterService implements IMediaServerService {
...
+   * 
+   * @param id 
+   * @param libraryId 
+   * @returns 
+   */
+  async itemIsInLibrary(id: string, libraryId: string): Promise<boolean> {
+   console.log ("not sure what to do about this")
+   return false
+  }
+
```
Simon-Eklundh: please observe

enoch85 (dismissed):

PR #2663, opened 2026-04-11
fix: track collection membership provenance

### LINKED ISSUE #2611: reused manual collection ui bug

enoch85:
## Summary

Fixes #2611 by tracking rule-owned and manual collection membership separately. When multiple rules target the same manual media server collection, each Maintainerr collection now keeps its own rule-owned membership while still showing genuinely shared manual items.

### Root cause

The old model stored a single `isManual` flag per `CollectionMedia` row. During media server sync, items discovered in a shared server collection but not tracked locally were imported as manual, even when they actually belonged to another rule targeting the same media server collection. That erased provenance and caused unrelated items to appear permanently pinned as manual.

### What this changes

- Replaces the single `isManual` flag with two independent fields:
  - `includedByRule`
  - `manualMembershipSource` (`legacy`, `local`, `shared`)
- Preserves rule and manual membership independently so one can be removed without deleting the other.
- Reconciles shared manual collections without re-importing rule-owned sibling items as manual.
- Preserves local and legacy manual provenance during sync instead of collapsing everything into shared membership.
- Applies scoped membership removal (`rule`, `manual`, `all`) so rule execution and media server sync only clear the intended membership.

### Backward compatibility

`isManual` remains a derived field via entity hooks, so existing read paths continue to behave correctly while the new provenance fields drive collection membership behavior. The migration backfills `includedByRule` and `manualMembershipSource` from existing rows and remains reversible.

### Test 1

Test:
Create two Maintainerr rule groups that both target the same manual Plex or Jellyfin collection, then add media that only matches rule A.

Expected:
The item appears in rule A's Maintainerr collection only. It must not appear in rule B's Maintainerr collection as a manual item.

Old:
The item could bleed into rule B's Maintainerr collection and appear there as manual, even though it was only added by rule A.

### Test 2

Test:
Add media that only matches rule B in the same shared manual collection setup.

Expected:
The inverse of test 1. The item appears in rule B's Maintainerr collection only and does not show up in rule A as manual.

Old:
The same cross-rule bleed could happen in the opposite direction, making rule-owned items look manually pinned in the sibling collection.

### Test 3

Test:
Manually add an item directly in the media server collection shared by both Maintainerr collections.

Expected:
The item appears in both linked Maintainerr collections as manual/shared membership.

Old:
Manual and rule-owned provenance were not separated cleanly, so shared items and sibling rule-owned items could be conflated.

### Test 4

Test:
Remove a rule match from an item that also has manual membership.

Expected:
The item remains in the collection because the manual membership still exists.

Old:
Removing the rule match could remove the entire row and lose the surviving manual membership.

### Test 5

Test:
Remove a manually added item from the media server collection when that item is still rule-owned in Maintainerr.

Expected:
The manual membership is removed, but the item stays in the Maintainerr collection because the rule membership still exists.

Old:
The removal handling could treat the item as fully removed or otherwise corrupt the remaining membership state.

### Test 6

Test:
Add an exclusion in one linked collection for an item present in the shared media server collection.

Expected:
That item is not re-imported into the excluded collection as shared manual membership.

Old:
Shared/manual reconciliation could re-import items too broadly and ignore the intended per-collection exclusion boundary.

### Test 7

Test:
Run rule execution again after add/remove operations in a shared manual collection.

Expected:
Recently removed items are not resurrected, and recently added manual items are not dropped because of stale child enumeration from the media server.

Old:
Stale collection children could cause just-removed items to come back or just-added manual items to be cleared incorrectly.


enoch85:
@Simon-Eklundh Please test this, polished it with the final fixes I think is valid. :+1:

Simon-Eklundh:
just tested  @enoch85 

it seems to work to me at least.

I did need to run the rules and move out and into manual collection, but that was most likely because of an old rule having put it there, and I run two different instances depending on if I'm testing or not

PR #2671, opened 2026-04-12
feat: add explicit exists rule operators

enoch85:
## Summary
- add explicit EXISTS and NOT_EXISTS operators to the shared rule contracts and surface them for supported rule types
- keep normal comparisons fail-closed on missing values while treating empty strings and empty lists as missing for EXISTS and NOT_EXISTS
- omit fake second-value placeholders from unary comparison stats output and add clearer missing-value reasons in stats output
- add direct RuleConstanstService coverage for the dynamic null-reason builder
- update rule validation and the RuleInput editor so unary rules do not require a second value, while keeping the existing custom-value and reopen flows stable
- add focused server and UI regression coverage, including the null BEFORE fail-closed case and unary RuleInput behavior

## Example

```
    - id: 1
      result: true
      operator: OR
      ruleResults:
        - action: exists
          firstValueName: Jellyfin - Date added
          firstValue: 2025-12-14T00:55:08.013Z
          result: true
```


enoch85:
@d-Rickyy-b Have you tried this out yet?

enoch85:
Works in my testing, let's get this in development. :+1:

PR #2721, opened 2026-04-19
Fix storage totals for shared Arr filesystems

### LINKED ISSUE #2720: Storage totals double-count shared filesystems across Arr root folders
## Summary

Maintainerr's storage headline appears to double- or triple-count available space when Sonarr/Radarr expose multiple root-folder paths that all live on the same underlying filesystem.

In my case, `/movies`, `/tv`, and `/downloads` are all on the same 23.8 TB media filesystem, but Maintainerr reported about 66.4 TB available because those paths were summed independently.

## Reproduction

Environment shape:
- Radarr root folder: `/movies`
- Sonarr root folder: `/tv`
- Download path: `/downloads`
- All three paths point to the same underlying filesystem

Live filesystem reality:
```text
$ df -h /movies /tv /downloads
Filesystem  Size  Used Avail Use% Mounted on
data         22T  577G   22T   3% /movies
data         22T  577G   22T   3% /tv
data         22T  577G   22T   3% /downloads
```

Radarr diskspace payload:
```json
[
  {"path":"/downloads","freeSpace":23214726447104,"totalSpace":23833675956224},
  {"path":"/movies","freeSpace":23214726447104,"totalSpace":23833675956224}
]
```

Sonarr diskspace/root-folder payload includes:
```json
[
  {"path":"/tv","freeSpace":23214726447104,"totalSpace":23833675956224},
  {"path":"/downloads","freeSpace":23214726447104,"totalSpace":23833675956224}
]
```

Maintainerr `/api/storage-metrics` totals came back as:
```json
{
  "freeSpace": 69721478995968,
  "totalSpace": 73006456774656,
  "mountCount": 6
}
```

Expected totals should have been roughly:
```json
{
  "freeSpace": 23214726447104,
  "totalSpace": 23833675956224,
  "mountCount": 1
}
```

## Suspected Cause

From reading the current source:
- `ServarrApi.getDiskspaceWithRootFolders()` merges Arr `/diskspace` and `/rootfolder` entries by normalized path
- `StorageMetricsService.computeTotals()` dedupes mounts by `host + normalizeDiskPath(path)`

That means different paths on the same filesystem, like `/movies`, `/tv`, and `/downloads`, are treated as distinct storage targets and summed independently.

Relevant files:
- `apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts`
- `apps/server/src/modules/storage-metrics/storage-metrics.service.ts`

## Expected Behavior

The storage headline should count a shared filesystem once, even if Arr exposes it through multiple root-folder paths.

## Notes

I have a local workaround, but I have not opened a PR because I wasn't confident the dedupe is universally safe upstream. In particular, two different filesystems could theoretically have identical `totalSpace`/`freeSpace`, and Arr does not appear to expose a stable filesystem/device identifier in this payload.... unlikely but wanted to call it out.

I'm opening this issue first in case there is a preferred approach for:
- selecting only root-folder-backed mounts for headline totals, or
- teaching the totals logic how to identify shared filesystems more reliably

nbellowe:
## Summary

Fix storage headline totals when Sonarr and Radarr expose multiple root-folder paths that all point at the same underlying filesystem.

This update now:
- prefers root-folder-backed mounts when computing the headline total
- keys root-folder selection by `instanceType + instanceId` so Sonarr and Radarr IDs do not collide
- keeps the Arr root-folder merge logic shared in `ServarrApi`
- dedupes accurate mounts per host using the best available filesystem-like signature:
  - `label + totalSpace` when a stable volume label is present
  - otherwise `totalSpace + MiB-bucketed freeSpace` to absorb tiny cross-instance drift without collapsing obviously different same-size disks

Closes #2720.

## Root Cause

`getDiskspaceWithRootFolders()` can return multiple distinct paths like `/movies`, `/tv`, and `/downloads` for the same shared filesystem.

Before this change, `computeTotals()` deduped by normalized path, so those entries were summed independently even when they referred to the same media disk.

## Notes

Arr APIs do not expose a stable filesystem or device identifier in these payloads, so the totals logic still has to rely on a heuristic.

For accurate mounts without a label, the fallback now uses `totalSpace + MiB-bucketed freeSpace` as the filesystem signature. That fixes the shared-filesystem double-count while avoiding the stricter `totalSpace`-only collision on equal-capacity disks with materially different usage. Two separate unlabeled filesystems on the same host could still theoretically collide if they expose the same total space and land in the same free-space bucket.

## Validation

- Added focused unit coverage for:
  - shared-filesystem root folders being counted once
  - overlapping `radarr(1)` / `sonarr(1)` IDs not colliding
  - same-host free-space drift still merging shared filesystems
  - same-host equal-capacity disks with different usage staying distinct
  - inaccurate-total path dedupe
  - `getRootFolders()` returning `undefined` in the shared Servarr helper
- Ran:
  - `yarn workspace @maintainerr/server test --runInBand src/modules/storage-metrics/storage-metrics.service.spec.ts src/modules/api/servarr-api/common/servarr-api.service.spec.ts`
  - `yarn workspace @maintainerr/server tsc --noEmit`
  - `yarn workspace @maintainerr/server build`
- Result: all passed

### DIFF
--- a/apps/server/src/modules/api/servarr-api/common/servarr-api.service.spec.ts
+++ b/apps/server/src/modules/api/servarr-api/common/servarr-api.service.spec.ts
@@ -1,0 +1,45 @@
+import type { ArrDiskspaceResource } from '@maintainerr/contracts';
+import type { MaintainerrLogger } from '../../../logging/logs.service';
+import { ServarrApi } from './servarr-api.service';
+
+class TestServarrApi extends ServarrApi<Record<string, never>> {}
+
+describe('ServarrApi', () => {
+  let api: TestServarrApi;
+  let logger: jest.Mocked<MaintainerrLogger>;
+
+  beforeEach(() => {
+    logger = {
+      setContext: jest.fn(),
+      log: jest.fn(),
+      warn: jest.fn(),
+      debug: jest.fn(),
+    } as unknown as jest.Mocked<MaintainerrLogger>;
+
+    api = new TestServarrApi(
+      { url: 'http://localhost:7878', apiKey: 'test' },
+      logger,
+    );
+  });
+
+  it('returns diskspace mounts unchanged when root folders are unavailable', async () => {
+    const diskspace: ArrDiskspaceResource[] = [
+      {
+        id: 1,
+        path: '/movies',
+        label: null,
+        freeSpace: 100,
+        totalSpace: 200,
+        hasAccurateTotalSpace: true,
+      },
+    ];
+
+    jest.spyOn(api as any, 'getDiskspace').mockResolvedValue(diskspace);
+    jest.spyOn(api as any, 'getRootFolders').mockResolvedValue(undefined);
+
+    await expect(api.getDiskspaceAndRootFolders()).resolves.toEqual({
+      mounts: diskspace,
+      rootFolderPaths: new Set(),
+    });
+  });
+});
--- a/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
+++ b/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
@@ -114,26 +114,29 @@ export abstract class ServarrApi<QueueItemAppendT> extends ExternalApiService {
    * These merged entries are safe for remaining-space calculations and for the
    * UI path picker. Total-space rule evaluation must use raw /diskspace data.
    */
-  public getDiskspaceWithRootFolders = async (): Promise<
-    DiskSpaceResource[]
-  > => {
+  public getDiskspaceAndRootFolders = async (): Promise<{
+    mounts: DiskSpaceResource[];
+    rootFolderPaths: Set<string>;
+  }> => {
     const [diskspace, rootFolders] = await Promise.all([
       this.getDiskspace(),
       this.getRootFolders(),
     ]);
 
-    const results: DiskSpaceResource[] = [...(diskspace ?? [])];
+    const mounts: DiskSpaceResource[] = [...(diskspace ?? [])];
     const existingPaths = new Set(
-      results.filter((d) => d.path).map((d) => normalizeDiskPath(d.path!)),
+      mounts.filter((d) => d.path).map((d) => normalizeDiskPath(d.path!)),
     );
+    const rootFolderPaths = new Set<string>();
 
     for (const folder of rootFolders ?? []) {
       if (!folder.path) continue;
 
       const normalized = normalizeDiskPath(folder.path);
+      rootFolderPaths.add(normalized);
       if (!existingPaths.has(normalized)) {
         existingPaths.add(normalized);
-        results.push({
+        mounts.push({
           id: folder.id,
           path: folder.path,
           label: null,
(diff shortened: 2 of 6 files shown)


enoch85 (changes requested):
Thanks for the PR! :pray: 

A few things before we can merge:

### Need to fix
- Lint: `ArrDiskspaceResource` import is unused now (the explicit type annotation came off). ESLint trips on it.
- Prettier: the `(diskspace ?? []).map(...)` block needs a re‑wrap.

### Worth a second look
- Double API call. `client.getDiskspaceWithRootFolders()` already calls `getRootFolders()` internally, and the PR calls it again in parallel. `getRolling`'s cache won't save you on a cold fetch since both fire at once. Could you either pull root folders out of the existing helper, or call `getDiskspace()` + `getRootFolders()` once and merge here?
- Related, and medium‑sized: we're now doing a second `/rootfolder` fetch per Arr instance — `client.getRootFolders()` runs alongside `client.getDiskspaceWithRootFolders()`, even though the shared Servarr helper already does that merge over in `servarr-api.service.ts:117-123`. And since `external-api.service.ts:113-145` only does a cache lookup (no in‑flight dedupe), a cold storage‑metrics refresh ends up firing two `/rootfolder` calls plus one `/diskspace` call per instance. Would be nice to keep this in one place — either extend the Servarr helper to hand back both the merged mounts and the normalized root‑folder paths, or just fetch raw `getDiskspace()` and `getRootFolders()` once here and skip `getDiskspaceWithRootFolders()` entirely. Not strictly in scope for this PR (which is really about the filesystem‑signature dedupe), but worth tidying up while we're here.
- Dedupe key includes `freeSpace`. That value drifts between back‑to‑back Radarr/Sonarr queries on the same filesystem, so a one‑byte difference breaks the merge and the bug quietly pops back up. `totalSpace` alone is a lot more stable — or bucket `freeSpace` to MB.

### Nice to have
- A test for `getRootFolders()` returning `undefined`, and one for the `hasAccurateTotalSpace=false` path branch.
- Tiny stray blank line at line 283 — unrelated to the fix.

Really appreciate the writeup in the PR body, especially calling out the shared‑signature edge case up front. Once lint/format and the duplicate fetch are sorted, this is good to go.

enoch85:
@nbellowe I've decided to help you fix this. :+1:

enoch85:
Checks pass --> merge to `development` for further testing.

nbellowe:
Thanks @enoch85 ! I can test it locally as well, when I get a few minutes in the next week.

enoch85:
> Thanks @enoch85 ! I can test it locally as well, when I get a few minutes in the next week.

Would be great!

nbellowe:
Hey @enoch85 there is still an issue if I have multiple containers, with different hostnames, and the same volumes and some double counts. I think we should remove the "hostname" from the key. Any objections? Happy to open up a PR.

enoch85:
@nbellowe If you open a PR, use GPT 5.4 please. 

Test the changes locally first and make it tight, minimal and tell the AI to do fact based changes. :+1: 

That said, we want to keep hostnames.

nbellowe:
Not really sure how else to fix it (besides removing hostname, which has the issue of not supporting multi-host setups using the same disks on multiple hosts well). There doesn't appear to be additional disk metadata available to use for grouping. We could use "free disk space" but that could have issues as well.

enoch85:
> Not really sure how else to fix it (besides removing hostname, which has the issue of not supporting multi-host setups using the same disks on multiple hosts well). There doesn't appear to be additional disk metadata available to use for grouping. We could use "free disk space" but that could have issues as well.

Please open a new issue and describe it as detailed as possible.

PR #2733, opened 2026-04-21
- remove collection/rule event-driven overlay processing

MrLinford:
- keep overlays running only on the overlay schedule and explicit overlay actions
- replace event-listener tests with scheduled task coverage

### Description & Design

This PR combines a set of related server-side hardening and cleanup changes across collections, Plex integration, task coordination, and overlays.

Problems addressed by this branch:
- shared media-server exports were exposing Plex-specific internals through a server-agnostic boundary
- Collection controller request bodies were not being validated tightly enough at the boundary
- rules and collections relied on a shared execution lock concept without a single canonical key
- parts of the Plex API layer still depended on broad `any`-style typing
- Collection write paths had looser contracts than needed
- Overlays could be triggered indirectly by rule or collection membership events, even though overlays already have their own schedule and explicit actions

This approach was chosen because it improves correctness by tightening boundaries instead of layering on more conditional behavior:
- provider boundaries are made explicit
- Request validation happens at the controller edge
- Lock coordination uses one shared constant
- Plex request and response handling is typed more narrowly
- Collection write contracts are clearer and better tested
- overlay execution is owned by the overlay task rather than by unrelated rule and collection events

At a high level, the branch works as follows:
- stops exporting Plex-specific internals from the shared media-server barrel
- adds stronger request validation for collection controller write operations
- centralizes the shared rules/collections execution lock key
- replaces broad Plex API `any` usage with more explicit types
- adds focused tests that lock current Plex client request behavior
- makes collection write contracts more explicit and adds supporting tests
- removes event-driven overlay processing from collection/rule membership changes so overlays run only from the overlay task schedule or explicit overlay actions
- replaces the removed overlay listener tests with scheduled-task coverage

Important trade-offs and limitations:
- overlays no longer react immediately to collection membership changes from rules or collection handling
- that change is intentional and aligns overlay behavior with its dedicated schedule and explicit controls
- This branch includes a mix of behavioral fixes and maintenance-oriented refactors, so reviewers should pay attention to the intended scope of each commit rather than treating the branch as overlay-only work

### Related issue

None linked.

### AI-Assisted Development

AI-assisted areas:
- tracing the interactions between rules, collections, tasks, Plex integration, and overlays
- drafting and refining some of the server-side changes and focused Jest coverage
- reviewing commit scope and preparing PR/commit summary text

Personally reviewed, changed, or validated:
- verified the boundary changes around media-server exports and shared task locking
- reviewed the collection controller validation and explicit collection write contract updates
- reviewed the Plex typing cleanup and the request-behavior test coverage
- verified the overlay task responsibility change so overlays no longer run from rule/collection membership events
- validated focused test coverage where changes were made
- validated a full workspace build

Why this fits the project:
- it reduces coupling rather than adding more special-case logic
- it keeps shared abstractions provider-agnostic
- it pushes validation to the request boundary
- it improves type clarity without broadening public APIs
- it keeps overlays owned by their own task/schedule instead of relying on side effects from unrelated subsystems

> Note: The author is fully responsible for the correctness, testing, and maintainability of this change, regardless of the tooling used.

### Checklist

### How to test

1. Verify collection controller validation by sending invalid collection write payloads and confirming they are rejected at the API boundary.
2. Run the relevant collection and Plex-focused tests and confirm the request/contract behavior remains stable.
3. Trigger rules and collection handling manually and confirm they no longer trigger overlay execution as a side effect.
4. Use the overlay-specific Run now action and confirm overlays are processed only from that explicit action.
5. Update overlay settings with a cron schedule and confirm the overlay task still uses its own schedule correctly.
6. Run a full build and confirm the monorepo compiles cleanly.

### Additional context

This branch currently contains these areas of change:
- media-server boundary cleanup
- collection controller validation
- shared execution lock cleanup
- Plex request behavior tests and API typing cleanup
- explicit collection write contracts
- overlay task execution boundary cleanup

Validated locally during this work:
- full workspace build via `yarn build`
- focused Jest coverage for overlay task behavior

There are multiple logical commits in this branch, so commit-by-commit review may be easier than reviewing it as a single feature.


MrLinford:
For reviewers, this branch breaks down into these commits from oldest to newest:

- b7f4ed97 — fix(media-server): stop exporting plex internals  
This tightens the shared media-server boundary so Plex-specific internals are no longer exposed through the provider-agnostic export surface.

- e20aff24 — fix(collections): validate controller request bodies  
This adds stronger request validation at the collections controller boundary so invalid write payloads are rejected earlier and more predictably.

- c112da5f — refactor(tasks): share rules collections lock key  
This centralizes the shared rules/collections execution lock key so both flows coordinate through one canonical constant instead of duplicated string values.

- a0f44ccf — test(plex): lock current client request behavior  
This adds regression coverage around current Plex client request behavior so later refactors can be made without accidentally changing request construction.

- f4b72ff5 — refactor(plex): replace broad api any types  
This narrows Plex API typing in the server integration layer to reduce broad any-style behavior and make the code easier to reason about safely.

- 6aed55a5 — refactor(collections): make write contracts explicit  
This makes collection write contracts more explicit and adds focused test coverage so the expected shapes and behaviors are clearer in the write path.

- aa42809a — overlay task cleanup  
This removes event-driven overlay execution from rule and collection membership changes, so overlays now run only from their own schedule and explicit overlay actions, and it updates the overlay task tests to cover scheduled behavior instead of the removed listeners.

Validation completed during this work:
- Full workspace build passed
- Focused overlay task Jest coverage passed

apps/server/src/modules/tasks/execution-lock.service.ts
```diff
@@ -1,5 +1,7 @@
 import { Injectable } from '@nestjs/common';
 
+export const RULES_COLLECTIONS_EXECUTION_LOCK_KEY = 'rules-collections-lock';
```
enoch85, line 3: This is a bit sloppy, why use an alias here?

MrLinford: From what I understand, it is not really an alias in the TypeScript sense; it is a shared key constant.

I’d agree that the placement is sloppy. `execution-lock.service.ts` is a generic lock service, so exporting a domain-specific key from that file mixes the mechanism with a specific consumer. The cleaner version would be:

- keep the constant
- move it to a small constants module, for example, a task-level or rules/collections shared constants file
- Leave ExecutionLockService responsible only for lock behavior

So the constant is justified; the file it lives in is the questionable part. 

Could it be moved to a cleaner constants file and the imports updated?

enoch85: Even better, create a constants file for overlay, and move all the constants there. One place, one single source of truth, easier to find and tweak if needed.

enoch85: I will fix it.

enoch85: I misread this, it's not an alias, it's a key, so yes, it makes sense.

Sorry, tired from work.

enoch85 (approved):

PR #2766, opened 2026-04-25
fix: stop cross-rule contamination for same-titled automatic collections

### LINKED ISSUE #2765: Rules targeting same Plex collection are populating both Maintainerr collections
**Describe the bug**
When two different rules target the same (non-manual) collection, items matches by one rule appear in the other Maintainerr collection as manual. 

**To Reproduce**
- We have rule A and rule B, they both go to (non-manual) collection "Going Soon".
- For simplicity sake, let's say those rules have no intersection (they can never both be true).
- Rule A has action "delete" after 14 days, Rule B has action "delete" after 7 days.
- Let's say our item matches rule A.
- Once both rules are run, the collection for rule A will include this item
- The collection for rule B (with the same name) will also include this item, where it will be marked manual.
- the movie will be deleted after 7 days in the Plex collection, despite it being matched with rule A (should delete after 14 days) and not rule B.

**Expected behavior**
- An item added to a mutual collection will follow the action rules and "action after days" of the rule that originally added it to the collection
- In any case, I think it'd be nicer if the item didn't show in the collection at all, or something other than "manual" showed up.

**Screenshots**

 - OS: Linux
 - Version: 3.7.0
 - Server: Plex

**Debug logs**
Irrelevant.

**Additional context**

enoch85:
## Summary

Fixes #2765. Two automatic rule groups with the same title currently end up linked to the same media server collection (because `checkAutomaticMediaServerLink` relinks by name), and the rule executor's manual-import path then pulls the sibling rule's items into the wrong rule's `collection_media` as `manual`. The collection worker doesn't filter on membership type, so those items get handled at the wrong rule's `deleteAfterDays`. There's also a related landmine where whichever rule's local rows clear first deletes the shared media server collection out from under the sibling rule.

This PR keeps the existing same-title sharing behavior intact (it's still useful for relinking after a Plex restart), but makes it safe:

- New `CollectionsService.getSiblingRuleOwnedMediaServerIds(collection)` returns the set of media server IDs that are rule-owned by any other Maintainerr collection sharing this collection's `mediaServerId`.
- The rule executor's manual-import loop ([`rule-executor.service.ts`](apps/server/src/modules/rules/tasks/rule-executor.service.ts)) skips children that fall in that set, so a sibling rule's items never get imported as manual into the wrong rule.
- `removeFromCollection` and `checkAutomaticMediaServerLink` now check `isMediaServerCollectionShared` before deleting an empty media server collection. When shared, the local link is cleared but the media server collection is preserved for the sibling rule.

This mirrors the provenance approach used by #2616 for manual collections, applied to the automatic-collection same-title case it didn't cover.

## Validation

- `yarn workspace @maintainerr/server test` — 1021/1021 passing (4 new)
- `yarn workspace @maintainerr/server lint` — clean
- `yarn build` — clean

### DIFF
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -163,6 +163,94 @@ describe('CollectionsService', () => {
     );
   });
 
+  it('returns rule-owned media server ids from sibling collections sharing a media server collection', async () => {
+    const collection = createCollection({
+      id: 1,
+      mediaServerId: 'remote-collection',
+    });
+    const sibling = createCollection({
+      id: 2,
+      mediaServerId: 'remote-collection',
+    });
+    collectionRepo.find.mockResolvedValue([sibling]);
+    collectionMediaRepo.find.mockResolvedValue([
+      createCollectionMedia(sibling, {
+        mediaServerId: 'rule-owned',
+        includedByRule: true,
+        manualMembershipSource: null,
+      }),
+      createCollectionMedia(sibling, {
+        mediaServerId: 'manual-only',
+        includedByRule: false,
+        manualMembershipSource: CollectionMediaManualMembershipSource.LOCAL,
+      }),
+    ]);
+
+    const result = await service.getSiblingRuleOwnedMediaServerIds(collection);
+
+    expect(Array.from(result)).toEqual(['rule-owned']);
+  });
+
+  it('does not delete a shared media server collection when one rule empties locally', async () => {
+    const collection = createCollection({
+      id: 11,
+      mediaServerId: 'remote-collection',
+      manualCollection: false,
+    });
+    const collectionMedia = [
+      createCollectionMedia(collection, { mediaServerId: 'item-1' }),
+    ];
+
+    collectionRepo.findOne.mockResolvedValue(collection);
+    collectionMediaRepo.find
+      .mockResolvedValueOnce(collectionMedia)
+      .mockResolvedValue([]);
+    collectionRepo.save.mockImplementation(async (value) => value as Collection);
+    jest
+      .spyOn(service as any, 'checkAutomaticMediaServerLink')
+      .mockResolvedValue(collection);
+    jest
+      .spyOn(service as any, 'removeChildrenFromCollection')
+      .mockResolvedValue(['item-1']);
+    jest
+      .spyOn(service, 'isMediaServerCollectionShared')
+      .mockResolvedValue(true);
+
+    await service.removeFromCollection(collection.id, [
+      { mediaServerId: 'item-1' },
+    ]);
+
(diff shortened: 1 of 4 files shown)


SHxKM:
@enoch85 Thanks for the super fast response to #2765  🙏 this looks awesome.

One reservation I have is: does that mean shared collection will NEVER be deleted? My concern with this is that we'll end up with cruft / users will see empty collections. I think it's OK for a collection handler to delete the collection as long as it checks (with a lock) that it has just deleted the last item(s), then the handler for the next rule (and assuming it needs to) would just recreate it, no? 

I'm unfamiliar with the codebase so I may be offering suboptimal suggestions here.

enoch85:
@SHxKM Please test the latest changes.

> One reservation I have is: does that mean shared collection will NEVER be deleted? My concern with this is that we'll end up with cruft / users will see empty collections. I think it's OK for a collection handler to delete the collection as long as it checks (with a lock) that it has just deleted the last item(s), then the handler for the next rule (and assuming it needs to) would just recreate it, no?
> 
> I'm unfamiliar with the codebase so I may be offering suboptimal suggestions here.

Yeah, fair concern - but it's not "never deleted", just "not deleted while a sibling still points at it". Once the last sibling rule group is gone (deleted/renamed/moved library), the empty collection gets cleaned up on the next run via the normal path. Plex collections are also recreated by name, so the recreate side is already free.

Your lock-based idea is cleaner in theory, but Maintainerr doesn't have a cross-rule lock today and adding one felt heavy for what's basically "two rule groups with the same title in the same library" — pretty rare. The sibling-count check is the passive version of the same idea: no other rule holds the link → safe to delete.

If empty-collection cruft ever becomes a real problem, a periodic sweep would be the easy follow-up.

SHxKM:
Thanks @enoch85. 

> The sibling-count check is the passive version of the same idea: no other rule holds the link → safe to delete.

But wouldn't there _always_ be a sibling when two rules point to the same collections? Or is this referring to actual items in the Maintainerr collection.

enoch85:
> Thanks @enoch85.
> 
> > The sibling-count check is the passive version of the same idea: no other rule holds the link → safe to delete.
> 
> But wouldn't there _always_ be a sibling when two rules point to the same collections? Or is this referring to actual items in the Maintainerr collection.

@SHxKM "Sibling" here means another Maintainerr collection row pointing at the same Plex mediaServerId - not items inside the collection. So yes, while both rule groups exist and both link to the shared Plex collection, there's always a sibling and we won't delete it. The sibling goes away when you delete one of the rule groups, rename one so it no longer collides, or unlink one from Plex. At that point the remaining rule is no longer "shared" and the normal delete-if-empty path kicks back in.

enoch85:
All tests pass here, and I've been over it a couple of times.

Merging to `development`. :rocket:

PR #2800, opened 2026-04-28
feat: cache jellyfin collections

natekspencer:
### Description & Design

Stores jellyfin collections and collection children to cache to avoid excessive repeat queries during rules that look at collections.

### Checklist

### How to test

1. Add a rule targeting Jellyfin collections
2. Run rule

### Additional context

Currently, rules that target Jellyfin collections result in a query for all collections and a query for each collection's children, repeated for every item processed. This quickly balloons and results in excessively long runtimes. For example, targeting a library with 500 items and 50 collections would result in 25,500 queries per rule run. With caching, this is reduced to 51 calls.

### DIFF
--- a/apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.spec.ts
+++ b/apps/server/src/modules/api/media-server/jellyfin/jellyfin-adapter.service.spec.ts
@@ -24,6 +24,7 @@ const jellyfinApiMocks = {
   getItems: jest.fn(),
   getItem: jest.fn(),
   getItemUserData: jest.fn(),
+  updateItem: jest.fn(),
   refreshItem: jest.fn(),
   getItemImage: jest.fn(),
   setItemImage: jest.fn(),
@@ -72,10 +73,12 @@ jest.mock('@jellyfin/sdk/lib/generated-client/models', () => ({
     ProviderIds: 'ProviderIds',
     Path: 'Path',
     DateCreated: 'DateCreated',
+    ChildCount: 'ChildCount',
     MediaSources: 'MediaSources',
     Genres: 'Genres',
     Tags: 'Tags',
     Overview: 'Overview',
+    ParentId: 'ParentId',
     People: 'People',
   },
   LocationType: {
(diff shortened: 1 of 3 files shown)


natekspencer:
> I'm not sure about this.
> 
> Cache is good for fairly static objects, but collections isn't one of them. Cache can be very error prune. What's your motivation for this PR?

My direct motivation: I have a rule that takes up to 20 minutes without the cache and drops to under 2 with the cache.

More details:
Right now, a rule that targets items in a collection such as:
```yaml
- firstValue: Jellyfin.collection_names
  action: CONTAINS_PARTIAL
  customValue:
    type: text
    value: Cleanup Eligible
```
currently causes excessive calls to collections and their children for every single item processed, which results in a significant performance problem at scale. 

Collections don't need to be re-queried for every item within a single rule run. If a collection changes mid-run and that is a concern, every previously processed item would need to be re-evaluated, resulting in a cyclic problem that isn't worth solving. The `jellyfin:` cache items appear to be purged between rule group runs anyway, ensuring it is always fresh at the start of each rule group run.

enoch85:
> > I'm not sure about this.
> > Cache is good for fairly static objects, but collections isn't one of them. Cache can be very error prune. What's your motivation for this PR?
> 
> My direct motivation: I have a rule that takes up to 20 minutes without the cache and drops to under 2 with the cache.
> 
> More details: Right now, a rule that targets items in a collection such as:
> 
> ```yaml
> - firstValue: Jellyfin.collection_names
>   action: CONTAINS_PARTIAL
>   customValue:
>     type: text
>     value: Cleanup Eligible
> ```
> 
> currently causes excessive calls to collections and their children for every single item processed, which results in a significant performance problem at scale.
> 
> Collections don't need to be re-queried for every item within a single rule run. If a collection changes mid-run and that is a concern, every previously processed item would need to be re-evaluated, resulting in a cyclic problem that isn't worth solving. The `jellyfin:` cache items appear to be purged between rule group runs anyway, ensuring it is always fresh at the start of each rule group run.

Yeah, sorry, deleted my comment when I read your descripton. :see_no_evil: 

I will review this soon. Almost ready to say LGTM though. :laughing:

natekspencer:
> Yeah, sorry, deleted my comment when I read your descripton. 🙈
> 
> I will review this soon. Almost ready to say LGTM though. 😆

All good, haha. ^_^

enoch85:
OK, I will fix some minor stuff here:

1. Both new caches can go stale because the mutation paths don't del them:

createCollection (service.ts:1337) → should invalidate jellyfin:collections:<libraryId>
deleteCollection (service.ts:1377) → should invalidate both jellyfin:collections:<libraryId> and jellyfin:collections:children:<collectionId>
addToCollection / addBatchToCollection (service.ts:1483, :1487) → should invalidate jellyfin:collections:children:<collectionId> (and arguably the parent collections cache, since childCount is now stale)
removeFromCollection / removeBatchFromCollection (service.ts:1560, :1580) → same as above

2. Empty results are cached 

3. Tests

enoch85:
@natekspencer Thanks for this PR! I'm sure users will be happier now when it just says "swosch" and rules are done. :rocket: 

I'll merge this when tests pass. :+1:

natekspencer:
> @natekspencer Thanks for this PR! I'm sure users will be happier now when it just says "swosch" and rules are done. 🚀
> 
> I'll merge this when tests pass. 👍

I know I will be! Thanks for your work on this, too!

Issue #2807, opened 2026-04-29
OverlayProccessorService crashing Maintainerr when reverting item no longer in collection


Michael-Busse:
**Describe the bug**

**To Reproduce**
1. Simply crashes whenever Overlay processing starts

**Device (please complete the following information):**
 - OS: Unraid / Docker
 - Version 3.9.0
 - Server: Plex

**Debug logs**
```
[maintainerr]  |  29/04/2026 09:35:57  [INFO] [OverlayProcessorService] === Overlay processor started ===
[maintainerr]  |  29/04/2026 09:35:57  [INFO] [OverlayProcessorService] Processing 5 overlay-enabled collection(s)
[maintainerr]  |  29/04/2026 09:35:57  [INFO] [OverlayProcessorService] Item 78052 no longer in any overlay collection, reverting
[maintainerr]  |  29/04/2026 09:35:57  [ERROR] [main] The server has crashed because of an uncaughtException. This is likely a bug, please report this issue on GitHub: write EPIPE | code=EPIPE
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)

```

This is running on docker on an Unraid server. Plex is running on a stand-alone server in this case.

Michael-Busse:
A bit more debug info:

[maintainerr]  |  29/04/2026 10:14:24  [INFO] [OverlayProcessorService] === Overlay processor started ===
[maintainerr]  |  29/04/2026 10:14:24  [INFO] [OverlayProcessorService] Processing 5 overlay-enabled collection(s)
[maintainerr]  |  29/04/2026 10:14:24  [INFO] [OverlayProcessorService] Item 78052 no longer in any overlay collection, reverting
[maintainerr]  |  29/04/2026 10:14:24  [DEBUG] [PlexApiService] getPosters(78052) failed: Error: GET http://172....56:32400/li...ers failed with exception: Plex Server didnt respond with a valid 2xx status code, response code: 404
[maintainerr]  |  29/04/2026 10:14:24  [ERROR] [main] The server has crashed because of an uncaughtException. This is likely a bug, please report this issue on GitHub: write EPIPE | code=EPIPE
Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)


So it seems Plex is responing with a 404 for this specific request for some reason.

enoch85:
[...]

PR #2812, opened 2026-04-29
fix: harden server logging against stdio EPIPE

enoch85:
## Summary

Prevent broken `stdout`/`stderr` pipes from crashing the server during logging by installing stdio guards before bootstrap and covering the behavior with focused tests.

Related: #2807

## Validation

- `yarn workspace @maintainerr/server jest --runTestsByPath src/modules/logging/winston/stdioPipeGuard.spec.ts src/modules/events/events.controller.spec.ts src/modules/logging/logs.controller.spec.ts src/modules/overlays/overlay-processor.service.spec.ts`

### DIFF
--- a/apps/server/src/main.ts
+++ b/apps/server/src/main.ts
@@ -8,6 +8,13 @@ import path from 'path';
 import { AppModule } from './app/app.module';
 import { dataDir } from './app/config/dataDir';
 import { MaintainerrLogger } from './modules/logging/logs.service';
+import { installStdioPipeGuards } from './modules/logging/winston/stdioPipeGuard';
+
+// Pre-bootstrap guard so the console.warn/console.error calls below — and any
+// other write before LogsModule loads — cannot crash the process on a broken
+// stdio pipe. The logging module re-installs these (idempotent) for
+// defence-in-depth.
+installStdioPipeGuards();
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule, {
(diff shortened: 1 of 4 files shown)


enoch85:
@SmolSoftBoi Please double check this.

apps/server/src/main.ts
```diff
@@ -95,6 +102,21 @@ process
...
+        : undefined;
+
+    // A broken stdout/stderr pipe (EPIPE/ERR_STREAM_DESTROYED) is an
+    // output-sink failure, not an application correctness failure. The
+    // logging module installs guards on process.stdout/stderr to swallow
+    // these; this is a silent fallback in case one slips through. We don't
+    // log here because writing to the same broken sink could re-trigger the
+    // crash we're avoiding.
+    if (code === 'EPIPE' || code === 'ERR_STREAM_DESTROYED') {
+      return;
```
SmolSoftBoi: When a non-stdio stream or integration lets an `EPIPE` or `ERR_STREAM_DESTROYED` escape to the process handler, this new branch treats it as a logging sink failure and returns before the existing fatal logging and exit path. That silently keeps the server running after an uncaught exception from unrelated code, the stdio listener already handles `process.stdout` and `process.stderr`, so this fallback should be scoped to known stdio failures.

SmolSoftBoi (changes requested):

apps/server/src/main.ts
enoch85: Solved it. Thanks for pointing that out!

PR #2814, opened 2026-04-29
feat(overlays): upload images for overlay Image elements from the UI

enoch85:
## Summary
- New compact image picker in the overlay Properties panel — `[label] [Select] [Upload]` — replacing the manual filename text input. Backed by `POST /api/overlays/images`. Bytes are validated through `sharp` so a renamed text file is rejected (400). On-disk files placed in `data/overlays/images/` continue to be discovered automatically because the new endpoint lists that directory.
- Single shared `IMAGE_UPLOAD_MAX_BYTES` constant (500 KB) backs both the collection poster and overlay image upload limits.
- Bootstrap creates `data/overlays/{fonts,images}/` so the list/delete endpoints work on a fresh install.
- Extracted `ResourceField` is reused by the existing font picker and the new image picker.


apps/ui/src/components/OverlayEditor/OverlayCanvas.tsx
```diff
@@ -391,8 +475,35 @@ function ElementRenderer({
...
+        const offsetY = (h - drawH) / 2
+        return (
+          <Group {...commonProps}>
+            <KonvaImage
+              image={loaded}
+              x={offsetX}
+              y={offsetY}
+              width={drawW}
+              height={drawH}
+              listening={false}
```
SmolSoftBoi: Once an image loads, the placeholder hit target is replaced by a group whose only child has `listening={false}`. Konva groups do not draw their own hit area from `width`/`height`, so the image element becomes difficult or impossible to click, tap, or drag after loading. Keep the image listening, or add a transparent listening rect covering the element bounds.

apps/ui/src/components/OverlayEditor/PropertiesPanel.tsx
```diff
@@ -442,67 +472,31 @@ function FontFields<
...
-      </label>
+      <ResourceField
+        label="Font"
+        value={selectValue}
+        options={fonts}
+        onSelect={applyFontByName}
+        onUpload={async (file) => {
+          const uploaded = await onUploadFont(file)
+          if (uploaded) applyFontByName(uploaded.name)
+          return uploaded
```
SmolSoftBoi: For font uploads, this wrapper calls `applyFontByName`, then `ResourceField.handleUpload` calls `onSelect(uploaded.name)` again after the upload promise resolves. A single font upload therefore emits duplicate element updates and extra undo history entries. Let `ResourceField` own the post-upload selection, or add an option to suppress it for callers that already select.

SmolSoftBoi (changes requested):
Found two actionable issues in PR #2814 both in the overlay editor UI path.

apps/ui/src/components/OverlayEditor/OverlayCanvas.tsx
enoch85: Confirmed and fixed in fdf8d09a — dropped `listening={false}` from the loaded `KonvaImage` so it acts as the hit target itself, matching how Text/Variable/Shape elements work in this file. The other `listening={false}` instances on the canvas background layer are intentional and unchanged.

apps/ui/src/components/OverlayEditor/PropertiesPanel.tsx
enoch85: Confirmed and fixed in fdf8d09a — replaced the inline wrapper with `onUpload={onUploadFont}`. `ResourceField.handleUpload` already calls `onSelect(uploaded.name)` after a successful upload, and `onSelect` is `applyFontByName`, so the post-upload select now happens exactly once. Matches how the new image picker is wired (`onUpload={onUploadImage}`).

apps/server/src/modules/overlays/overlays.controller.ts
```diff
@@ -316,14 +326,125 @@ export class OverlaysController {
...
+    return path.join(configDataDir, 'overlays', 'images');
+  }
+
+  @Get('images')
+  listImages() {
+    const imagesDir = this.getImagesDir();
+    return fs
+      .readdirSync(imagesDir)
+      .filter((f) => this.isSupportedImageFile(f))
+      .map((name) => ({ name, path: path.join(imagesDir, name) }));
```
SmolSoftBoi, line 361: `listImages()` returns every supported extension even when the filename is not safe, but `getImage()` and the render service reject those same names via `isSafeFilename`. A manually dropped file like `My Logo.png` appears in the picker, then previewing `/api/overlays/images/My%20Logo.png` returns 400 and the final overlay omits it. Filter with `isSafeFilename` here, or normalise manual filenames consistently before exposing them.

SmolSoftBoi (changes requested):

apps/server/src/modules/overlays/overlays.controller.ts
enoch85: Confirmed and fixed in 230a0a73 — `listImages()` now filters with `isSafeFilename` in addition to the supported-extension check, so the picker, the `GET /api/overlays/images/:name` endpoint, and the render service all agree on the same allow-list. Manually-dropped files like `My Logo.png` are simply hidden instead of appearing as broken picks.

Added a regression test covering safe vs space/traversal/separator names.

enoch85: Followed up in bf170cd8 — applied the same `isSafeFilename` filter to `listFonts()` so both pickers behave consistently with the GET endpoints and the render service. Regression test added for fonts too.

PR #2815, opened 2026-04-29
fix(overlays): persist overlay template on rule group save

### LINKED ISSUE #2805: [BUG] Overlay template selection reverts to "Default Poster Template" after saving rule group
**Describe the bug**
When editing a rule group and selecting any overlay template other than "Default Poster Template", the selection does not persist after saving. The save completes without error, but on checking the rule group again the overlay template has reverted to "Default Poster Template".

**Steps to reproduce**
1. Open an existing rule group
2. Enable overlays on the rule group
3. Select the built-in "Countdown" template from the overlay template dropdown (template is visible and selectable)
4. Click Save
5. Observe: save completes with no error
6. Re-open the rule group — overlay template has reverted to "Default Poster Template"

**Expected behavior**
The selected overlay template should persist after saving.

**Additional context**
This was tested on the `main` tag (image built 2026-04-28, sha256: 4d4eb1ae52084900f89ac144b90607747346c73698cd209ea4beaa19a8ee2296) after confirming the previous `Failed to save rule group` error (filed separately) is resolved in this build.

The "Default Poster Template" option is always selected regardless of what was chosen before saving. All other rule group fields save correctly.

**Environment**
- Maintainerr version: main branch (3.9.0 RC, built 2026-04-28)
- Image: ghcr.io/maintainerr/maintainerr:main
- Host: Synology DS920+, DSM 7.x
- Deployment: Docker Compose

enoch85:
## Summary
- Fixes #2805 — overlay template selection silently reverted to "Default" after saving an existing rule group.
- Root cause: the effect that clears an unavailable template ran before the async `getOverlayTemplates()` fetch resolved, found an empty list, and reset the form value to `null`.
- Gated the effect on a new `overlayTemplatesLoaded` flag so it only runs once the fetch has returned.


enoch85:
Tested and works!

PR #2816, opened 2026-04-29
fix(overlays): show locked-templates hint when overlays disabled

enoch85:
## Summary
iOS Safari greys disabled `<option>` entries in the native picker silently — `change` never fires, so the toast-only feedback for blocked templates routes was easy to miss on mobile (Discord report: \"can't select anything in the Overlay settings\"). Adds a persistent inline note under the tab/dropdown when overlays are disabled.

The earlier resize-clamp commit was reverted: the bounds did fire in tests but didn't address the underlying iOS resize bug, so it's removed from this PR to keep the change focused.


SmolSoftBoi (changes requested):
The `AddModal` template persistence fix does not add a focused regression test for issue #2805 so that behaviour relies on inspection and broader CI rather than a targeted component test.

enoch85:
Addressed in b8002448: extracted the effect's gating logic into a pure `shouldClearOverlayTemplateSelection` helper and added four cases — loading, no selection, matching selection, and stale selection. The first one is the #2805 regression.

enoch85:
This needs to be tested first. Highly uncertain it will work, Timelord on Discord needs to confirm.

PR #2852, opened 2026-05-06
fix(storage-metrics): merge shared volumes across hosts when byte-exact

### LINKED ISSUE #2841: Storage space being counted twice
**Describe the bug**

**To Reproduce**
1. Go to 'Storage Tab'
2. Scroll down to 'Total Capacity'

**Additional context**
My total capacity is being reported as 37.45 TB which is not correct, as it is just adding up the capacity of Radarr and Sonarr which use the same storage backend but operating from different LXC containers.  

Image

enoch85:
Closes #2841.

Storage totals were doubled when Radarr and Sonarr ran in separate LXC containers against the same backend volume — different hostnames meant PR #2721's per-host dedupe could not collapse them.

Adds a second pass after the per-host dedupe that merges mounts across hosts only when the signature is implausibly tight for unrelated filesystems: byte-exact `totalSpace` + `freeSpace`, or matching volume label + byte-exact `totalSpace`. Any block-level write diverges unrelated hosts byte-for-byte, so the false-positive surface is narrow.

Per-host dedupe still uses the existing 1 MiB freeSpace bucket to absorb same-host poll drift between Arr clients.

### DIFF
--- a/apps/server/src/modules/storage-metrics/storage-metrics.service.spec.ts
+++ b/apps/server/src/modules/storage-metrics/storage-metrics.service.spec.ts
@@ -478,7 +478,37 @@ describe('StorageMetricsService', () => {
       expect(totals.mountCount).toBe(2);
     });
 
-    it('does not merge identical capacities across different hosts', () => {
+    it('merges shared storage across different hosts when totalSpace and freeSpace match byte-for-byte', () => {
+      const totals = compute(
+        [
+          mount({
+            instanceType: 'radarr',
+            instanceId: 1,
+            path: '/data/media/movies',
+            freeSpace: 2761668689920,
+            totalSpace: 4000000000000,
+          }),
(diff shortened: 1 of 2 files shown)


enoch85:
@rstuke82 thanks for the detailed report — the byte-exact `freeSpace` match across both rootfolder payloads was exactly the signal this fix uses to detect shared backend storage across separate hosts.

A `pr-2852` Docker tag is being built now. Once it's up, would you give it a try and confirm your storage card no longer double-counts?

nbellowe:
hey thanks for this. This is the same issue that I was trying to point out. Appreciate it!

enoch85:
@nbellowe so does it work? Did you test the pr-tag? 🙂

nbellowe:
Tested locally - works!
Thanks!

PR #2856, opened 2026-05-07
Feature/plex collection sort

00Scooby:
> **Note to contributors:** > This template is part of our review process. Pull requests that do not meaningfully complete the sections below, or that do not demonstrate understanding of the change, may be closed without detailed review.

### Description & Design

Please include:

- **What problem this change solves:** Users currently cannot automatically sort the items inside a media server collection created by Maintainerr. This PR adds the ability to define a custom sort order (e.g., Release Date, Date Added, Alphabetical) directly from the Rule Group settings.
- **Why this approach was chosen:** We leverage Plex's custom sort capabilities. By applying the sort directly after the batch add process during rule execution, we ensure that the collection immediately reflects the user's desired order on the media server.
- **How the solution works at a high level:** 1. A `mediaServerSort` column is added to the Collection database entity.
  2. The UI (`AddModal`) is updated with a dropdown to select the sort criteria (conditionally rendered for Plex).
  3. The `PlexApiService` and `PlexAdapterService` are extended with `setCollectionCustomSort` and `moveCollectionItem` endpoints.
  4. In `collections.service.ts`, after adding new items to the collection, the backend retrieves the hydrated metadata, sorts the items locally according to the chosen criteria, and pushes the exact new order to the media server.
- **Any important trade-offs, edge cases, or limitations:** This feature is currently limited to Plex, as Jellyfin handles custom collection sorting differently. The UI field is gated behind the `isPlex` check, and the backend logic is protected by `mediaServer.supportsFeature(MediaServerFeature.COLLECTION_SORT)`.

### Related issue

no related issue

### AI-Assisted Development

If you used AI tools (e.g. ChatGPT, Copilot, Claude) while working on this PR, please describe:

- **Which parts were AI-assisted:** Generating the boilerplate for the Plex API integration, adjusting the UI forms, and structuring the local sorting logic in `collections.service.ts`.
- **What you personally reviewed, changed, or validated:** I manually verified all Plex API calls via network traces, ensured database migrations worked cleanly on a fresh setup, debugged and fixed caching/startup errors (`ECONNREFUSED` and empty DB states), and rigorously tested the end-to-end flow on my local instance.
- **Why this solution fits this project’s coding standards and design:** The implementation cleanly follows the existing Adapter pattern (`IMediaServerService`), hides Plex-specific logic behind feature flags, and seamlessly integrates into the existing UI components and Rule Executor lifecycle.

> Note: The author is fully responsible for the correctness, testing, and maintainability of this change, regardless of tooling used.

[...]

PR #2860, opened 2026-05-08
fix(collections): apply collection sort on save, harden Plex reorder

enoch85:
## Summary

Follow-up to #2856. Three fixes flagged in post-merge review plus one related improvement.

**Save-time apply (HIGH).** Picking or changing the collection sort in the rule form now takes effect at save time. Previously the executor only pushed the order on cycles where `newMedia.length > 0`, so the UI could show a configured sort that didn't exist on Plex until membership happened to change.

**Per-item resilience (MEDIUM).** Plex `reorderCollectionItems` no longer aborts the entire move loop on a single failed move. Each `moveCollectionItem` is wrapped individually; failures are accumulated and surfaced in a single summary warning at the end.

**No-op short-circuit (MEDIUM).** Plex adapter reads current child IDs first and returns early if they already match the requested order — zero PUTs in the matching case (was 1 prefs PUT + N move PUTs).

**Alphabetical tiebreaker for tied sorts.** When `airDate` / `rating` / `watchCount` / `deleteSoonest` returns 0, the comparator now falls through to its own `case 'title':` branch (A→Z, direction-agnostic). Reuses the existing title-sort logic via recursion — no new helper. Affects every consumer of `compareMediaItemsBySort`, so Maintainerr's UI display and the Plex collection-sort push now order tied items the same way. Addresses #2856 (comment).

Status sorts (`manual` / `excluded`) intentionally skip the tiebreaker — their existing comment documents that they only partition.

## Test plan

- [x] `yarn workspace @maintainerr/server test --runInBand` — 1160 tests pass (was 1151)
- [x] `yarn workspace @maintainerr/ui test --run` — 195 tests pass
- [x] `yarn lint` — 0 errors / 0 warnings
- [x] `yarn format:check` — clean
- [x] New `sort-utils.spec.ts` covers the comment scenario (`G,H,S,X,A,C,Z`) and asserts manual/excluded preserve partition
- [x] `plex-adapter.service.spec.ts` adds short-circuit and mid-loop-failure cases
- [x] `rules.service.updateRules.spec.ts` covers apply-on-enable and no-op-when-unchanged transitions

### DIFF
--- a/apps/server/src/modules/api/media-server/media-server.interface.ts
+++ b/apps/server/src/modules/api/media-server/media-server.interface.ts
@@ -272,6 +272,10 @@ export interface IMediaServerService {
    * (or no-op) before applying. Gated by MediaServerFeature.COLLECTION_SORT.
    * Throws if not supported. Caller is responsible for filtering out
    * smart collections (Plex rejects move on smart).
+   *
+   * Implementations should short-circuit when the current child order
+   * already matches `orderedItemIds`, and continue through the full list
+   * if individual moves fail (logging a summary at the end).
    */
   reorderCollectionItems(
     collectionId: string,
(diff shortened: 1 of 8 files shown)


enoch85:
@timelordx :)

PR #2868, opened 2026-05-10
fix(collections): sort by addDate and group episodes by show

enoch85:
Reported in Discord on `main`: with "Delete Soonest" / "Delete Latest" the Plex collection order is wrong, the alphabetical tiebreaker never fires, and TV episodes are interleaved by episode title instead of grouped by show.

**Maintainerr's DB is the source of truth for ordering.** `applyCollectionSort` pushes Maintainerr's order to the media server; the UI reads from the same DB. Both must agree on the primary sort key (`collection_media.addDate`).

## Key changes

- **`deleteSoonest` reads `collection_media.addDate`** via a `deleteSoonestDate` callback on the comparator options, replacing the previous `MediaItem.addedAt` (which is the unrelated library-add date).
- **Day buckets align with the visible "Leaving in X days" countdown.** New `deleteSoonestReferenceTime` option (`now − deleteAfterDays × dayMs`) makes `ceil((addDate − reference) / dayMs)` the bucket key, so two items with the same overlay countdown tie even if their `addDate`s straddle UTC midnight. Within a tied bucket the show-aware title order is the tiebreaker. Used by `applyCollectionSort`.
- **`deleteSoonest` UI paging stays SQL-paginated** — orders by `collection_media.addDate` directly at the DB level instead of hydrating every row in the collection. Restored after the initial PR removed it; hydrate-then-sort on a 127-item Jellyfin collection blocked the UI page for minutes. Trade-off: same-day items on the UI page may show in a slightly different within-bucket order than the polished Plex/Jellyfin collection (which has the title tiebreak applied by `applyCollectionSort`). Primary sort key matches; both views agree on which items leave first.
- **Title sort groups episodes under their show** via `grandparentTitle ?? parentTitle ?? title`, then `title`. Movies are unaffected. Shared with the library Overview comparator, so general library title sorts also group episodes under their show — matches Plex's own behavior.

## Comparator hardening (same surface)

- Items missing the sort value (no air date / no rating / no view count / no add date) now sort to the end regardless of asc/desc. Previously they coerced to `0` and silently sorted to the front in `asc` mode.
- `airDate`, `rating`, `watchCount`, and `deleteSoonest` collapse onto one `compareNumericWithTitleFallback` helper.
- `airDate` now day-buckets like `deleteSoonest`.

## Other

- Pinned `tsBuildInfoFile` in `apps/server/tsconfig.build.json` so nest-cli's type-check pass stops emitting a stray `tsconfig.build.tsbuildinfo` at the package root.
- Removed unused `baseUrl` + `@/*` paths from `apps/ui/tsconfig.json` (cleared a TS 6 deprecation in the IDE).

## Caveat

Existing collections won't visually re-sort in Plex until membership changes — `applyCollectionSort` only re-runs when `newMedia.length > 0`. Editing a rule (touch + save) triggers it. Worth a release-notes line.

## Test plan

- [ ] On a movie collection with `deleteSoonest.asc`, items leaving today appear before items leaving in 7+ days, and the page loads in under a second on a collection with 100+ items
- [ ] In the actual Plex/Jellyfin collection, items added at 23:00 vs. 01:00 the next morning showing the same overlay countdown sort alphabetically together
- [ ] Episode collection sorted by Title A–Z keeps episodes from the same show contiguous
- [ ] Items with no `addDate` / no air date / no rating sort to the **end** in both `asc` and `desc`
- [ ] `apps/server/tsconfig.build.tsbuildinfo` does NOT exist after `yarn workspace @maintainerr/server build` (only `apps/server/dist/tsconfig.build.tsbuildinfo`)
- [ ] `yarn lint`, `yarn test`, `yarn build`, `yarn dev` all clean

### DIFF
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -1220,21 +1220,16 @@ describe('CollectionsService', () => {
     });
   });
 
-  it('orders deleteSoonest paging by collection_media.addDate so the UI matches the media-server push', async () => {
-    // Regression for #2867: the API response and applyCollectionSort must
-    // agree, otherwise the user sees one order in Maintainerr and another
-    // in the media-server collection. MediaItem.addedAt is held identical
-    // across all rows here so any path that falls back to it would tie
-    // and the result order would depend on Map iteration.
+  it('paginates deleteSoonest at the SQL level by collection_media.addDate', async () => {
+    // `deleteSoonest` is equivalent to ordering by `collection_media.addDate`
+    // because `deleteAfterDays` is constant across a collection. SQL does the
+    // pagination so we don't have to hydrate every row in the collection
+    // before slicing — critical for collections with hundreds of items where
+    // hydrating all rows would block the UI for minutes.
     const collection = createCollection({
       id: 9,
       mediaServerId: 'remote-collection',
       type: 'movie',
-    });
-    const libraryAddDate = new Date('2024-06-01T00:00:00Z');
-    const leavesLatest = createCollectionMedia(collection, {
-      mediaServerId: 'leaves-latest',
-      addDate: new Date('2024-02-01T10:00:00Z'),
     });
     const leavesSoonest = createCollectionMedia(collection, {
       mediaServerId: 'leaves-soonest',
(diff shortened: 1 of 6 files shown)


enoch85:
@timelordx

Please test the PR tagged release.

enoch85:
@00Scooby would be great with some confirmation on this! 🙏🏼

enoch85:
Tested on Jellyfin (Collections sorting) and it works as expected).

Nitpick, some flicker when sorting. Might be my browser though.

timelordx:
@enoch85 ... tested `A-Z`, `Z-A`, and `Delete Soonest` sorting, and everything seems to be working perfectly. Well done!

00Scooby:
Hey @enoch85,  
Sorry I missed the tag earlier (Was sleeping)! Glad to see @timelordx already confirmed it and it's merged. Thanks for jumping on this so quickly, the TV show grouping is a much-needed improvement! 🚀

PR #2880, opened 2026-05-12
fix(security): add yarn resolutions to patch transitive dependency vulnerabilities

blixten85:
## Problem

Several transitive dependencies are pinned to vulnerable versions in the lockfile by parent packages that use exact version constraints. These cannot be fixed by updating the parent packages directly — the pinning is deliberate in those parent packages.

| Package | Vulnerable version | Pinned by | CVEs |
|---|---|---|---|
| dompurify | 3.2.7 | `monaco-editor@0.55.1` (exact pin) | XSS (multiple) |
| picomatch | 4.0.2 | `@angular-devkit/core@19.1.7` (exact pin) | ReDoS |
| postcss | 8.4.49, 8.5.8 | build tools via `^8.4.47` / `^8.5.8` | XSS |
| lodash | 4.17.21 | multiple packages via `^4.17.x` | Code Injection via `_.template` |
| lodash-es | 4.17.21 | multiple packages via `^4.17.x` | Code Injection via `_.template` |

## Fix

Adds yarn [`resolutions`](https://yarnpkg.com/configuration/manifest#resolutions) to root `package.json`. This forces yarn to install the specified minimum versions for all instances of these packages, regardless of what parent packages request.

```json
"resolutions": {
  "dompurify": ">=3.4.0",
  "picomatch": ">=4.0.4",
  "postcss": ">=8.5.10",
  "lodash": ">=4.18.0",
  "lodash-es": ">=4.18.0"
}
```

The `yarn.lock` is regenerated with `yarn install --mode=update-lockfile` to apply the resolutions without touching node_modules.

## Not fixed here

The following vulnerable packages require major-version bumps that could introduce breaking changes and are left for separate PRs:

- `undici` 5.29.0 (via `@actions/http-client`) — fixed in 6.24.0, breaking API change
- `undici` 6.22.0 (via `cheerio@1.0.0`) — fixed in 6.24.0
- `undici` 7.16.0 (via `@semantic-release/github`) — fixed in 7.24.0
- `ip-address` 9.0.5 (via `socks`) — fixed in 10.1.1, major version
- `nodemailer` 6.x / 7.x (transitive) — fixed in 8.0.5, major version

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@semantic-release/commit-analyzer": "^13.0.1",
     "@semantic-release/exec": "^7.1.0",
     "@semantic-release/git": "^10.0.1",
-    "@types/node": "^22.19.17",
+    "@types/node": "^25.7.0",
     "knip": "^6.9.0",
     "semantic-release": "^25.0.3",
     "turbo": "^2.9.6",
@@ -83,5 +83,12 @@
     "publish": [
       "@semantic-release/github"
     ]
+  },
+  "resolutions": {
+    "dompurify": ">=3.4.0",
+    "picomatch": ">=4.0.4",
+    "postcss": ">=8.5.10",
+    "lodash": ">=4.18.0",
+    "lodash-es": ">=4.18.0"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2382,57 +2382,58 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


blixten85:
Closing to resubmit a cleaner version. This PR accidentally included an unrelated @types/node major version bump (^22 → ^25) that was in the fork's development branch but not in upstream. Will resubmit with only the yarn resolutions security fix.

PR #2883, opened 2026-05-12
fix(plex): improve error message when library section ID is invalid

### LINKED ISSUE #1302: Removed Plex Library while rules were attached to that library

blixten85:
Fixes #1302.

## Problem

When a Plex library is removed and re-added it receives a new section ID. Rules and collections still reference the old ID, so Plex returns either HTTP 404 or HTTP 200 with a missing `MediaContainer`. Both cases previously logged the misleading:

```
[WARN] [PlexApiService] Plex api communication failure.. Is the application running?
```

That points users at connectivity instead of the actual cause: a stale library section ID.

## Fix

- Null-guard on `MediaContainer` and 404-aware catch routing for every section-scoped endpoint that takes a user-supplied library ID: `getLibraryContentCount`, `getLibraryContents`, `getLibraryLeaves`, `searchLibraryContents`, `getRecentlyAdded`, `getCollections` (the path hit by collection-aware rule evaluation), and the section-scoped writes `createCollection` / `updateCollection`.
- Single `logLibrarySectionError(id, error?)` helper. Detection reads `error.cause.response.status` (typed) and matches only on `status === 404`. 401 and 403 share the same wrapper in `lib/plexApi.ts` but mean auth/permission failures, so they fall through to the generic communication-failure log.
- Spec coverage for `getCollections`: empty `MediaContainer`, 404, 401, 403, and a non-HTTP transport failure.

**Before:**
```
[WARN] [PlexApiService] Plex api communication failure.. Is the application running?
[DEBUG] [PlexApiService] Cannot read properties of undefined (reading 'MediaContainer')
```

**After (404 or empty body):**
```
[WARN] [PlexApiService] Plex library section '16' returned no data. The library may have been removed or its ID changed in Plex. Update any rules or collections that reference this library.
```

### DIFF
--- a/apps/server/src/modules/api/plex-api/plex-api.service.spec.ts
+++ b/apps/server/src/modules/api/plex-api/plex-api.service.spec.ts
@@ -375,6 +375,114 @@ describe('PlexApiService.getMetadata', () => {
   });
 });
 
+describe('PlexApiService.getCollections (invalid section vs auth)', () => {
+  let service: PlexApiService;
+  let settingsService: PlexApiSettingsStub;
+  let logger: Mocked<MaintainerrLogger>;
+  let loggerFactory: Mocked<MaintainerrLoggerFactory>;
+
+  beforeEach(async () => {
+    const { unit, unitRef } = await TestBed.solitary(PlexApiService).compile();
+
+    service = unit;
+    settingsService = unitRef.get(
+      SettingsService,
+    ) as unknown as PlexApiSettingsStub;
+    logger = unitRef.get(MaintainerrLogger);
+    loggerFactory = unitRef.get(MaintainerrLoggerFactory);
+
+    settingsService.plex_hostname = 'plex.local';
+    settingsService.plex_port = 32400;
+    settingsService.plex_ssl = 0;
+    settingsService.plex_auth_token = 'token';
+    loggerFactory.createLogger.mockReturnValue({
+      setContext: jest.fn(),
+      log: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+      debug: jest.fn(),
+    } as any);
+  });
+
+  it('warns about a stale section when Plex returns 200 with no MediaContainer', async () => {
+    (service as any).plexClient = {
+      queryAll: jest.fn().mockResolvedValue({}),
+    };
+
+    await expect(service.getCollections('42')).resolves.toBeUndefined();
+
+    expect(logger.warn).toHaveBeenCalledWith(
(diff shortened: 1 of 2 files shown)


enoch85:
You can't push your dev environment here. ;)

I agree about better logging, but I'm not sure we can assume that it's *always* the same case for the error codes. Also ask Claude to read implementation guidelines and please follow the logging structure in the repo.

blixten85:
> You can't push your dev environment here. ;)
> 
> I agree about better logging, but I'm not sure we can assume that it's _always_ the same case for the error codes. Also ask Claude to read implementation guidelines and please follow the logging structure in the repo.

Ok

enoch85:
@blixten85 I will tidy this up for you.

blixten85:
> @blixten85 I will tidy this up for you.

Thanks 😁
I’ll make sure Claude code reads the .code folder, the whole repo in fact, and tells me what it will push in the future

enoch85:
@blixten85 Depending on which IDE you use it should already be read. Here are for Claude Code for example: https://github.com/Maintainerr/Maintainerr/tree/development/.claude

enoch85 (approved):
Approved with changes.

blixten85:
> @blixten85 Depending on which IDE you use it should already be read. Here are for Claude Code for example: https://github.com/Maintainerr/Maintainerr/tree/development/.claude

Yes, I thought it did that automatically, but I was wrong 😑

PR #2905, opened 2026-05-15
fix(plex): extract logInvalidLibrarySection and fix Octokit type cast

blixten85:
### Description & Design

**What problem this change solves:**

Two related issues in `plex-api.service.ts` and `github-api.service.ts`:

1. **`logLibrarySectionError` was called with one argument in six places** — the no-error path (when `response?.MediaContainer` is missing) — despite the method accepting `error?: unknown`. This relied on an implicit `error === undefined` branch to treat the missing-section case. The intent was unclear and the method signature was misleading.

2. **TypeScript type error in `github-api.service.ts`**: `Octokit.plugin(throttling)` produces a TS2345 error due to a type incompatibility between the `octokit` package's bundled `@octokit/core` and the standalone `@octokit/core` that `@octokit/plugin-throttling` depends on (`retryCount` property missing in one variant).

**Why this approach was chosen:**

- Extract `logInvalidLibrarySection(id)` as a dedicated private method for the "section returned no data" case. The six call sites that detected a missing `MediaContainer` now call it directly, making their intent explicit.
- Make `logLibrarySectionError(id, error)` require `error: unknown` — it now only handles thrown exceptions, with 404 delegating to `logInvalidLibrarySection`.
- Cast `throttling` to `Parameters<typeof Octokit.plugin>[0]` to silence the type incompatibility without changing runtime behaviour.

**How the solution works:**

- No behaviour change — the warn message and routing logic are identical.
- The refactor makes the two cases (no-response vs. thrown error) structurally distinct and easier to follow.

**Trade-offs / limitations:**

- The Octokit cast is a workaround for a package dependency conflict; the proper fix would be aligning `octokit` and `@octokit/plugin-throttling` to share the same `@octokit/core`.

### Related issue

Follows up on #2883 (which introduced `logLibrarySectionError` with structured 404 detection).

### AI-Assisted Development

This PR was developed with Claude Code assistance. I reviewed all changes for correctness, verified the test suite passes, and confirmed the behaviour is identical to the previous implementation.

### Checklist

### How to test

1. Run `yarn check-types` — should complete without errors
2. Run `yarn workspace @maintainerr/server jest --testRegex="plex-api.service.spec|github-api.service.spec"` — all 38 tests pass
3. Configure a Plex library section with an invalid/removed ID and verify the warn log appears correctly

### DIFF
--- a/apps/server/src/modules/api/github-api/github-api.service.ts
+++ b/apps/server/src/modules/api/github-api/github-api.service.ts
@@ -36,7 +36,9 @@ export class GitHubApiService {
     this.loadPersistedCache();
 
     // Create Octokit instance with throttling plugin
-    const OctokitWithPlugins = Octokit.plugin(throttling);
+    const OctokitWithPlugins = Octokit.plugin(
+      throttling as Parameters<typeof Octokit.plugin>[0],
+    );
 
     const octokitOptions: ConstructorParameters<typeof OctokitWithPlugins>[0] =
       {
--- a/apps/server/src/modules/api/plex-api/plex-api.service.ts
+++ b/apps/server/src/modules/api/plex-api/plex-api.service.ts
@@ -499,7 +499,7 @@ export class PlexApiService {
       });
 
       if (!response?.MediaContainer) {
-        this.logLibrarySectionError(id);
+        this.logInvalidLibrarySection(id);
         return undefined;
       }
 
@@ -535,7 +535,7 @@ export class PlexApiService {
       );
 
       if (!response?.MediaContainer) {
-        this.logLibrarySectionError(id);
+        this.logInvalidLibrarySection(id);
         return undefined;
       }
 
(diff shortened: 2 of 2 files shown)


enoch85:
Nope. This doesn't have the quality I'd expect.

What problem does it solve, and why is it needed?

blixten85:
> Nope. This doesn't have the quality I'd expect.
> 
> What problem does it solve, and why is it needed?

It’s from a issue report? 
But ok, I won’t spend tokens in contributing 😋

enoch85:
That's not what I meant. Just that maybe you should take some more care instead of writing a promt and expect wonders. 🙂

In the description there is a tradeoff explained for example. That's not good enough. What would better allign with the repo? What would be using shared functions and be consistent with the *entire* code base? What would be the most minimal, yet most effective change? Etc etc... Just doing like, "please fix this issue" won't work here - and this change looks like it.

I'm happy to accept PRs but we need to share the same philosophy for it to be merged.

PR #2906, opened 2026-05-15
refactor(rules): extract resolveArrTagNames helper for ARR tag resolution

blixten85:
## Summary

- Extracts a shared `resolveArrTagNames(tagIds, client)` utility from the duplicated inline tag-resolution chains in `RadarrGetterService` and `SonarrGetterService`
- Changes `ServarrApi.getTags()` return type from `Tag[]` to `Tag[] | undefined` on failure — callers can now distinguish a transient API error from a server with no tags configured (previously both returned `[]`)
- Adds a 5-test unit spec covering: matching labels, no match, empty IDs, API failure (`undefined`), and Radarr/Sonarr consistency

## Test plan

- [ ] `yarn workspace @maintainerr/server jest --testPathPatterns="arr-tags"` — 5 tests pass
- [ ] `yarn test` — full suite passes with no regressions
- [ ] Prettier check passes (`yarn workspace @maintainerr/server prettier --check .`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
+++ b/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
@@ -169,7 +169,7 @@ export abstract class ServarrApi<QueueItemAppendT> extends ExternalApiService {
     }
   };
 
-  public getTags = async (): Promise<Tag[]> => {
+  public getTags = async (): Promise<Tag[] | undefined> => {
     try {
       const response = await this.axios.get<Tag[]>(`/tag`);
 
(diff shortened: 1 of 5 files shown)


enoch85:
@blixten85 thanks but I'm not sure about this. This feels like a one off.

Have you tested the changes?

blixten85:
> @blixten85 thanks but I'm not sure about this. This feels like a one off.
> 
> Have you tested the changes?

I have tested the image, with plex, radarr, sonarr, overlay. And tag exist/does not exist in the rules

PR #2948, opened 2026-05-20
chore(server): switch jest transform to @swc/jest

### LINKED ISSUE #2946: Overhaul server circular dependencies (unblocks faster @swc/jest test transform)
## Background

The server test job intermittently prints:

> A worker process has failed to exit gracefully and has been force exited. This is likely caused by tests leaking due to improper teardown. Try running with --detectOpenHandles to find leaks. Active timers can also cause this, ensure that .unref() was called on them.

Investigation found this is **not** a leaked handle:
- `jest --detectOpenHandles` is clean across many runs.
- `node-cache` already `.unref()`s its check-period timer.
- The message only appears under CPU contention (e.g. `yarn turbo test` running jest + vitest concurrently).

It's slow `ts-jest` worker teardown under load — each worker holds a full compiled TS program in memory. The warning is left in place intentionally as a reminder for this overhaul (rather than masked with `--forceExit`).

## The real fix

Migrate the jest transform from `ts-jest` to `@swc/jest` (~7x faster, light workers that exit cleanly, no warning). A spike showed this works **but** SWC's stricter CommonJS live-binding exports turn the codebase's existing circular dependencies into TDZ errors:

> ReferenceError: Cannot access 'X' before initialization

So the migration is gated on first breaking those cycles:

- **~24 `forwardRef(() => X)` constructor injections** across ~12 service files (`SettingsService`, `MediaServerFactory`, `MediaServerSwitchService`, `PlexApiService`, etc.). These are genuine circular service dependencies.
- **Bidirectional TypeORM entity relations** (e.g. `Collection` ↔ `RuleGroup`).

## Suggested approach

1. Reduce/break the circular service dependencies where possible (many `forwardRef`s may be avoidable with better module boundaries).
2. For unavoidable entity cycles, use TypeORM's `Relation<>` wrapper + `import type` on relation properties.
3. For remaining `forwardRef` params, type-only-alias the annotation so `decoratorMetadata` doesn't emit an eager class reference.
4. Switch the jest transform to `@swc/jest` and drop the worker-teardown warning.

## Context

The warning was surfaced while fixing an unrelated vitest OOM in #2940 (settings-form render loop, fixed in `fix/settings-form-render-loop`). This overhaul was deliberately scoped out of that fix.

enoch85:
Closes #2946.

Replaces the server's `ts-jest` transform with `@swc/jest` (transpile-only). The server suite drops from ~66s to ~13s and workers exit cleanly — the intermittent "A worker process has failed to exit gracefully" warning is gone (it was slow ts-jest worker teardown under load, not a real handle leak).

SWC's stricter CommonJS live-binding exports turn the existing circular dependencies into TDZ `ReferenceError`s at module load, so this also makes those cycles SWC-safe with no runtime change:

- TypeORM relation property types wrapped in `Relation<>` (TypeORM's documented fix) so `design:type` no longer eagerly references the related entity.
- `forwardRef`-injected constructor params annotated via type-only import aliases so `design:paramtypes` doesn't eagerly reference the class; the value import remains for the `forwardRef` arrow.
- The media-server factory's three adapter injections gain explicit `@Inject(forwardRef(...))` (matching the two sibling params) so their types can be aliased too.

DI tokens/wiring are unchanged. Type-checking still runs via `nest build` in CI. The only behavioural change is to the test transformer.

Note: this adds `@Inject(forwardRef(...))` to the factory's adapter params — worth a quick boot smoke-test of the running server in addition to the green unit suite, since unit tests mock DI.


enoch85:
@SmolSoftBoi Does this make sense? Ran into this several times already. Tests that times out, going slow and whatnot - after this they are not.

Please look through it here, and tell me what you think. :+1:

enoch85:
@SmolSoftBoi Decided to merge, it's a good improvement, and I will follow up with a larger PR refactor later.

PR #2949, opened 2026-05-21
Track rule execution failures for collection media evaluation

### LINKED ISSUE #1603: Errors within rule execution for existing collection items could be handled better
Within the rule executor, if we run into an error getting data from any of the APIs via the valueGetter, the media item we're evaluating will end up being removed from the collection. If the error is transient it will be re-added on the next run.

I propose that we leave media items in the collection when this scenario occurs, in the hopes that it will be corrected on the next run. We should prevent those media items from being handled by the collection handler with some sort of flag, and perhaps surface this in the UI somewhere.

SmolSoftBoi:
## Summary
- Add persisted failure tracking for collection media rule evaluation.
- Update rule execution and collection worker logic to record and surface execution errors more reliably.
- Extend related entities, interfaces, and test fixtures to match the new failure state.
- Add coverage for the rule executor and collection worker error paths.

## Testing
- Added and updated unit tests for rule execution and collection worker behaviour.
- Local validation was not run in this turn.


enoch85:
@SmolSoftBoi Thanks!

Did you test this end to end?

apps/server/src/modules/collections/collection-worker.service.ts
```diff
@@ -118,12 +121,17 @@ export class CollectionWorkerService extends TaskBase {
           new Date().getTime() - +collection.deleteAfterDays * 86400000,
         );
 
-        const mediaToHandle = await this.collectionMediaRepo.find({
+        const dueMedia = await this.collectionMediaRepo.find({
```
enoch85: I like the old naming better. Feels like an unnecessary change.

enoch85:
@SmolSoftBoi Good job on this! Super clean, tight and minimal. :tada:

PR #2955, opened 2026-05-21
refactor(server): decouple SettingsService into data store + coordinator

enoch85:
Eliminates the server's circular-dependency debt at the root. `@swc/jest` (shipped in #2948) surfaced it as TDZ errors that were papered over with `forwardRef` + `import type { X as XType }` aliases. This removes the cause, so those workarounds go away.

## Root cause
`SettingsService` was a god-object: both a hydrated singleton holding ~30 persisted setting fields (read synchronously in ~97 places) **and** a coordinator injecting 7 API services (Plex/Servarr/Seerr/Tautulli/Streamystats/InternalApi/MediaServerFactory). Those API services injected `SettingsService` back just to read config → mutual cycles.

## What this does (the whole overview)
Split the god-object into two services named for their roles:
- **`SettingsDataService`** — owns the settings *data*: the persisted fields, `init()` hydration, `getSettings`/`getPublicSettings` (with secret masking), the read helpers, and low-level persistence. Depends only on the settings repositories — **zero service dependencies**, so anything can inject it.
- **`SettingsOperationsService`** — owns settings *operations*: connection tests + save/remove/reinit flows for every integration. Injects the API services as plain `private readonly` (no `forwardRef`); holds no settings state.
- **Media-server factory + Plex/Jellyfin/Emby adapters** and **all ~97 field reads / read consumers** (API services, schedulers, collection worker, rule executor, metadata, notifications, controller) now inject `SettingsDataService`.

## Result
| | before | after |
|---|---|---|
| `forwardRef` constructor injections | 28 | **2** |
| `@swc/jest` type-only aliases | 28 | **2** |

The only remaining pair is the **genuine mutual `MediaServerFactory ↔ MediaServerSwitchService` cycle** (factory checks `isSwitching()`; switch calls `uninitializeServer()`) — irreducible without further restructuring.

## Behaviour
Unchanged. Public API preserved via delegation; secret masking intact (the public `GET` endpoint still returns masked settings); save/test/connection and media-server-switch flows preserved (save → persist via the data service → re-hydrate).

## Validation
build (tsc) ✓ · 75 suites / 1335 tests ✓ (no worker-exit warning, no TDZ) · lint ✓ · prettier ✓ · DI boot smoke builds the full Nest context and resolves the rewired providers ✓.

Given the breadth, a manual smoke of settings save / connection-test / media-server switch on a running instance is worth doing before merge.


enoch85:
@copilot create an issue that we should do the same for modules as well.

enoch85:
Tested and works!

Issue #2936, opened 2026-05-20
perf: parallelize per-item Plex watch history calls


stormshaker:
## Background

The main rule execution bottlenecks can be addressed by bulk-fetching Sonarr/Radarr libraries and caching Plex watchlists/collections (one fetch per resource instead of one per media item). After those fixes, **Plex watch history** remains the only significant per-item bottleneck with no clean bulk equivalent.

## The problem

`PlexApiService.getWatchHistory(ratingKey)` is called once per media item for properties including `lastViewedAt`, `seenBy`, `sw_lastWatched`, `sw_viewedEpisodes`, `sw_amountOfViews`, `sw_allEpisodesSeenBy`, and `sw_watchers`.

For a rule processing 50,000 episodes, each needing `lastViewedAt`, that's 50,000 sequential `await` calls. At 50ms per round-trip on a local network, that's ~2,500 seconds. The event loop awaits each one before starting the next — fast hardware doesn't help sequential waiting.

There is **no Plex bulk endpoint** for watch history. `/status/sessions/history/all` without a filter is unbounded and not suitable for per-item lookup at scale.

## Proposed approach

Add a configurable concurrency pool to the rule executor's item processing loop using `p-limit` or similar. Processing N items at concurrency=10 reduces wall time by ~10× for watch-history-dependent rules without hammering the Plex server.

Key considerations:
- Default concurrency should be conservative (5–10) to avoid Plex rate limiting
- Error handling needs care: a single item failure shouldn't abort the whole batch
- Shared mutable state in the executor loop needs review before parallelising
- Concurrency limit could be user-configurable in settings

## Expected impact

For a 50,000-episode rule relying on `lastViewedAt`:
- Sequential (current): ~2,500s
- Concurrency=10: ~250s
- Concurrency=20: ~125s

## Related

- Issue #2580 — Sonarr bulk prefetch (same class of problem, addressed by bulk-fetch at the arr layer)
- PR #2800 — Jellyfin collection caching (bulk-cache pattern, analogous to watchlist/collections fix)

PR #2897, opened 2026-05-13
fix(sonarr): don't require never-downloaded seasons to be unmonitored before show cleanup

enoch85:
## Summary

Closes #2891 (and re-closes #2757). The "Unmonitor and delete season + delete show if empty" action deleted a season's files but never deleted the now-empty show. Two distinct bugs were stacked on top of each other.

## Root cause: stale cache (the reason nothing earlier worked)

`ExternalApiService.get()` caches responses for 20 minutes. `getSeriesByTvdbId` went through that cache. The empty-show cleanup's "refetch" — which runs immediately after the unmonitor + episode-file deletes — was a **cache hit**, so it evaluated the *pre-mutation* snapshot and concluded the show still had content.

Confirmed from daviddanko's Sonarr debug log: after the deletes, **no HTTP `GET /series?tvdbId=` was issued at all** — the refetch never hit Sonarr. Every gate-logic change before this was being fed stale input, which is why "reverted to basics" still failed.

Fix: `getSeriesByTvdbId` / `getMovieByTmdbId` now read via `getWithoutCache`. These endpoints are read straight after mutations and also drive rule evaluation — both need current truth. `updateMovie`'s internal read-modify-write read is uncached too, so it can't PUT a stale movie object back to Radarr.

## Secondary fix: the `allSeasonsUnmonitored` gate

Even with a fresh refetch, the no-Seerr fallback required *every* season to be unmonitored. Sonarr carries every TVDB season on the series — including ones the user never downloaded — and those stay monitored forever, permanently blocking deletion of a genuinely empty, ended show (e.g. a library with only seasons 1–4 of a 10-season show).

Fix: the no-Seerr fallback now requires `series.status === 'ended'` and no longer requires all seasons unmonitored. The `episodeFileCount === 0` file gate and the Seerr-authority path are unchanged. `isShowEmpty`'s `'monitored'` mode now also treats a monitored-but-fileless season as empty, and missing season statistics are treated as unknown rather than empty.

## Commits

- `5760c484` — don't require all seasons unmonitored to delete an empty ended show
- `620292f9` — apply the same fix to the unmonitor-show check
- `841c4adf` — treat missing season statistics as unknown, not empty
- `252ad047` — read series/movie lookups uncached so the post-mutation refetch is fresh

## Note

[...]

Issue #2891, opened 2026-05-13
"Unmonitor and delete season + delete show if empty" action not removing ended series from Sonarr


daviddanko:
**Describe the bug**
Opening this because it appears the issue #2757 was not fixed.

I have rule which should delete the show entirely from sonarr after it's been watched and is ended - no more episodes will be released.

<img width="794" height="619" alt="Image" src="https://github.com/user-attachments/assets/586c63c0-b2c9-4a40-8f32-4fb7276f189c" />

Today the show 'The End of the F***ing World' should have been deleted, but it wasn't. The logs say it still has files, but that's not true. Maintainerr correctly deleted the files. I cannot see the show on my disk/plex, however the show remains in sonarr.

**Expected behavior**
I expect maintainerr to delete the files from the disk (this is working correctly) and then delete the show entry from sonarr (this was not done).

**Device (please complete the following information):**
 - OS: Debian
 - Version 13
 - Server: Plex
 - Running in docker

**Debug logs**
```
2:00:00 AMINFO
Started handling of all collections
2:00:00 AMDEBUG
Skipping collection 'Movies' because no media is due for handling
2:00:00 AMDEBUG
Skipping collection 'TV shows (continuing)' because no media is due for handling
2:00:00 AMINFO
Collection handler summary: 3 total (isActive), 0 skipped (Do Nothing), 2 skipped (no due media), 1 queued for handling
2:00:00 AMINFO
Handling collection 'TV shows (ended)'
2:00:01 AMINFO
Unmonitored season 2 from Sonarr show with ID 389
2:00:01 AMINFO
[Sonarr] Removed season 2 from show 'The End of the F***ing World'
2:00:01 AMDEBUG
[Sonarr] Show 'The End of the F***ing World' still has 8 episode file(s) - skipping show deletion
2:00:01 AMDEBUG
[checkAutomaticMediaServerLink] Collection "TV shows (ended)" (DB id: 2, mediaServerId: 6765)
2:00:01 AMDEBUG
[checkAutomaticMediaServerLink] getCollection(6765) returned: id=6765, childCount=10
2:00:01 AMDEBUG
[checkAutomaticMediaServerLink] Trusting Plex metadata childCount=10 for collection 6765, keeping it
2:00:01 AMINFO
Removing 1 media items from collection..
2:00:01 AMINFO
Handling collection 'TV shows (ended)' finished
2:00:01 AMINFO
Sending webhook notification```

[...]

Issue #2757, opened 2026-04-24
"Unmonitor and delete season + delete show if empty" action not removing ended series from Sonarr


daviddanko:
I want to automatically delete a show from Sonarr once it has ended (no more episodes coming out).

Previously, I used the "Delete entire show" action. It worked exactly as I wanted.

<img width="1094" height="934" alt="Image" src="https://github.com/user-attachments/assets/451fe900-0966-45cf-80be-bf03d4ed9ffe" />

A few weeks ago, an update introduced a new option that I liked more: "Unmonitor and delete season + delete show if empty". 

<img width="1112" height="852" alt="Image" src="https://github.com/user-attachments/assets/86c00dac-76db-4594-bec1-12205beb939b" />

I thought this rule would delete seasons as I watch them and then delete the entire show from sonarr once the last episode of the last season is watched/deleted.

I have a [show](https://www.thetvdb.com/series/something-very-bad-is-going-to-happen) which has status ended, and it was correctly put in that collection using this section:

<img width="1510" height="800" alt="Image" src="https://github.com/user-attachments/assets/d7042e8f-cded-4bdf-ad0b-c0ddcfc56b28" />

However, after the rule handled the show only the files were deleted and the show entry is still sitting in Sonarr.

<img width="2179" height="1066" alt="Image" src="https://github.com/user-attachments/assets/f2b988a2-87e0-49b8-883f-8afb49bed622" />

Logs:

<img width="2098" height="595" alt="Image" src="https://github.com/user-attachments/assets/99474723-782b-4575-b2dc-e9106dcba953" />

[maintainerr-2026-04-24.log](https://github.com/user-attachments/files/27056293/maintainerr-2026-04-24.log)

I can see the following entry:

> [maintainerr]  |  24/04/2026 02:00:01  [INFO] [SonarrActionHandler] [Sonarr] Removed season 1 from show 'Something Very Bad Is Going To Happen'

Am I misunderstanding how this new action works, or is this a bug?

andrew-kennedy:
@enoch85  I noticed this as well, since I was involved in reviewing the PR where you worked on this. I can reproduce this, despite seerr having only completed requests for the show, and the show being empty of files once all seasons are deleted, the show itself in sonarr is left untouched. Not seeing any warning or info logs to indicate why though.

daviddanko:
Hi @enoch85,

Still doesn't seem to be working for me unfortunately.

[...]

PR #2957, opened 2026-05-21
perf: parallelize per-item Plex watch-history reads during rule execution

enoch85:
Speeds up rule evaluation on large libraries, and fixes a recent Sonarr/Radarr lookup regression. No new dependencies.

## 1. Parallelize per-item operand resolution
Rules resolve each item's operand one at a time; for watch-history-style properties (no bulk endpoint, #2936) and per-item *arr lookups this dominates runtime. The comparator now resolves `firstVal`/`secondVal` in bounded parallel batches (chunk + `Promise.all` — the same idiom as the existing Plex collection-write batching), then runs the unchanged backward mutation loop.

- `RULE_EVALUATION_CONCURRENCY = 8` — the single global cap and the only batching layer (never nested inside getters), abort checked inside each task.
- Conservative by design: the binding constraint is the slowest co-located backend (e.g. Tautulli's CPU-heavy history queries), not host core count. Higher values over-drive such backends past their request timeout, which then retries and amplifies load.

## 2. Fix the #2897 Sonarr/Radarr lookup regression
#2897 made `getSeriesByTvdbId`/`getMovieByTmdbId` uncached so the empty-show cleanup reads post-deletion truth (#2757 / #2891). Correct for the cleanup — but it also de-cached the **rule-evaluation** path, where the same series/movie resolves once per item, so episode-level rules over large libraries went from a handful of cached lookups to one uncached call per item per section.

`ArrLookupCache` restores the de-duplication without reintroducing the bug: a **run-scoped** memo created for the evaluation loop only and never handed to the collection/action phase, so the cleanup still reads fresh and a stale entry can't survive a deletion. The API-layer lookups stay uncached; only the two uncached *arr identity lookups use it (Tautulli/Seerr/Plex/Jellyfin/Emby already cache at the API layer).

## Real-world result
~13k-episode Tautulli+Sonarr rule on an all-in-one Intel N100:

| | Runtime | CPU | Tautulli errors |
|---|---|---|---|
| Before (stable, sequential) | hours | — | — |
| Concurrency 16 | 28 min | 90–95% | many (timeout → retry storm → items skipped) |
| **Concurrency 8 (this PR)** | **18 min** | **60–85%** | **none** |

8 is faster than 16 because removing the backend overload eliminated the timeout/retry stalls.

Tests: `ArrLookupCache` unit tests, an action-handler test asserting the cleanup reads series fresh (never via the memo), and the existing #2897 regression suites all pass.


enoch85:
Fixing: https://features.maintainerr.info/posts/186/perf-parallelize-per-item-plex-watch-history-calls

enoch85:
@daviddanko since we touch your bugfix in this PR, please confirm it's not reverted by this change. :pray:

daviddanko:
> @daviddanko since we touch your bugfix in this PR, please confirm it's not reverted by this change. 🙏

After a quick test it looks fine!

PR #2986, opened 2026-05-25
fix(rules): harden rule import (TEXT_LIST YAML decode + community-import operator backfill)

enoch85:
Two rule-import robustness fixes found while end-to-end testing the YAML/community import paths against the seeded dev stack.

### 1. TEXT_LIST custom value crashed YAML import (pre-existing)
The encoder serialises a custom value's `type` as the `RuleType` humanName, so a TEXT_LIST value becomes `type: text list` (with a space). `getCustomValueFromIdentifier`'s switch only matched `'TEXT_LIST'`, so the spaced key fell through, left `ruleType` undefined, and threw on `.toString()` — the decode `catch` returned a generic "Import failed". Pre-existing since #1530; not hit by community rules (they import via `/migrate`, not the YAML decoder). **Fix:** normalise spaces to underscores before matching.

### 2. Community import dropped an unset within-section operator (regression, #2971)
`migrateImportedRuleDtos` reasserted section-*boundary* operators but left within-section unset operators as `null`. A community rule authored before #2971 (e.g. id 81, `appVersion 2.19.0`: operators `[null, AND, null]` in one section) therefore kept a `null` operator on a non-first rule, which #2971's new "operator is required for every rule after the first" save validation rejects on import. **Fix:** backfill remaining non-first `null` operators to OR after the boundary pass — the same default the comparator and the `NormalizeRuleSectionOperators` migration apply.

### Verification
- Unit: `constants.service`, `rule-migration.service`, `yaml.service` specs — 50 passing (incl. 2 new regression tests); `check-types` + `format:check` clean; full repo test suite green.
- Live (seeded stack): TEXT_LIST custom value now round-trips (`code:1`, customVal preserved); **all v2.0.0+ community rules** migrate with **0** non-first null operators remaining (was 1 — rule 81, now `[null, AND, OR]`).


enoch85:
@copilot Please run yarn format on this PR.

PR #2989, opened 2026-05-26
test(rules): add regression test for section-operator migration backfill

enoch85:
Adds a regression test to the `NormalizeRuleSectionOperators` migration spec, covering a single group/section layout end-to-end: first-of-group rules stay null, section boundaries backfill to AND (`"0"`), and explicit operators (numeric `0` and string `"0"`) are left untouched.

Test-only — no production change. The migration's behaviour is unchanged.


enoch85:
@stormshaker Please test this when the build is done. Added logging.

Wait for Github Actions to build the PR tag, then use that.

enoch85:
@copilot Please clean this up. Only keep a small regression test that verifies the function and that the outcome is as expected from running the migration.

Remove the actual logging.

PR #2991, opened 2026-05-26
fix(rules): scope exclusions to their rule group under TypeORM 1.0

enoch85:
Adopts TypeORM 1.0's `where`-clause null/undefined semantics (default: throw; use `IsNull()` to match SQL `NULL`) instead of 0.3.x's silent-ignore footgun, and fixes the latent exclusion bugs the `typeorm@1.0.0` bump (#2983) surfaced. No compatibility flag — the code runs on real 1.0 semantics.

## Behaviour changes

| # | Scenario | 0.3.x | 1.0 (this PR) | Status |
|---|---|---|---|---|
| ① | An item excluded in **only** rule group A | excluded from **every** group's collection (the rule-group exclusion leaked), plus duplicate rows | excluded from group A only; other groups no longer filter it out, so they add/act on it | **CHANGED** |
| — | **Global** exclusion | applies everywhere | applies everywhere | SAME |
| — | Rule-group exclusion, in its **own** group | excluded | excluded | SAME |
| — | Add a **global** exclusion when a rule-group one exists | one global row (overwrote the group one) | one global row (the group one is removed) | SAME |
| — | Add a **rule-group** exclusion when already global | created a redundant 2nd row | no-op (the global already covers it) | SAME |
| ② | Make an item **global**, then remove the global, when it had **scoped** rule-group exclusions | rule-group exclusions in **other** groups survived (global-add converted only one) | **none survive** — going global removes all the item's scoped exclusions, so they don't come back | **CHANGED** |
| ③ | Remove a **global** exclusion | bogus "included" log on a random collection | no bogus log; logs the item + group/global | **CHANGED** |
| — | Remove a **rule-group** exclusion | logs "included" on its collection | same (clearer log line) | SAME |
| ④ | `getCollection()` with no/blank id | returned an arbitrary first collection | returns nothing (`undefined`) | **CHANGED** |
| ⑤ | Disconnect a notification config with a **missing/invalid id** | acted on an arbitrary rule group | returns "failed" (guarded) | **CHANGED** (edge/hardening) |
| — | Rule evaluation engine (208 getters + comparator) | — | byte-identical across versions | SAME |

The main real-world impact is ① in **multi-rule-group** setups: an exclusion you set in one group no longer hides the item in your other groups (matching the intent — a rule-group exclusion belongs to that group; a global one applies everywhere).

## Global vs rule-group exclusions

An item is either excluded **globally** (everywhere) or in **specific rule groups** — not both. Adding a global exclusion removes that item's scoped exclusions; adding a rule-group exclusion does nothing if the item is already global. Heads-up: if you later remove the global exclusion, the scoped ones don't come back — you'd re-add them.

## Upgrade note (for release notes)

Excluding an item inside a single rule group used to hide it in **every** group — that was a bug. It now applies only to the group you set it in. So an item you excluded in one group may now show up in (and be acted on by) your other groups. Existing exclusions aren't converted automatically: to exclude something everywhere, exclude it **globally** (no rule group selected).

## Also
- `engines.node` → TypeORM 1.0's floor (`^20.19 || ^22.13 || >=24.11`).
- Docs: object-form `relations`/`select` + `IsNull()` conventions (project-notes, AGENTS, typeorm_instructions).
- Coverage: per-property getter matrix (208 paths, version-independent), a real-DB exclusion-scoping integration test, and setExclusion / global-vs-scoped unit tests.


enoch85:
@stormshaker if you have time and many exclusions this would be a good testing candidate too. 🙏🏼

enoch85:
We now have a warning for the behavior change.


image


image

enoch85:
@SmolSoftBoi Tested this for 2 days now. Can't find any outstanding issue. FYI

stormshaker:
Sorry - been a busy week, I didn't get a chance to look at this one. But I see it's released!

PR #2996, opened 2026-05-27
fix(plex): don't drop auth when plex.tv is unreachable

### LINKED ISSUE #2990: TmdbApiService - Failed to resolve external IDs
Recently (I'm not really sure when, I just noticed that nothing is added to my "Leaving soon" collection), automated rule processing has stopped working. When testing media within the collection manually, it reports back correctly. However, when the rule is triggered (manually or scheduled), the logs are flooded with this error.
[log.log](https://github.com/user-attachments/files/28261023/log.log)
I'm not sure this is needed but, here are my rules for movies and episodes:
[episodes.yaml](https://github.com/user-attachments/files/28261272/episodes.yaml)
[movies.yaml](https://github.com/user-attachments/files/28261271/movies.yaml)

Troubleshooting Performed:
 - Spun up a completely separate container; the exact same behavior.
 - Verified all API connections to the *arrs, Plex, TMDB, and Overseerr. All endpoints report "Success".
 - Triggered a manual test on specific media within the collection, which processed without any errors.
 - Checked my pihole; nothing is blocked.

Device:
 - OS: DSM (Synology)
 - Version: 7.2.2
 - Server: Plex
 - Deployment: Docker container

enoch85:
When plex.tv times out, the stored-token check was reported as an invalid token, so the Plex settings page declared credentials invalid and demanded re-authentication even with a working saved token (issue #2990).

Now the validation distinguishes a real rejection (401/403/422) from a connectivity failure. On unreachable, the account stays authenticated, a non-blocking warning is shown, and the check retries a few times before stopping (revisiting the page re-checks).


enoch85:
@bassterror Please test the pr-tagged release commented here below (soon). Should solve your issues with Plex.

Please note: **do not test on your prod!** This contains irreversible migrations.

enoch85:
Tested on dev, works as intended.

PR #3002, opened 2026-05-28
feat(rules): metadata fallback when series absent from Sonarr

enoch85:
If a show is in your media server but not in Sonarr (you deleted it, never added it, or your library has a movie that Jellyfin/Emby tagged as a series and gave TV-show IDs), rules touching show-level properties get silently skipped right now — Sonarr has no record to read. This PR adds a small fallback: when Sonarr returns nothing, we ask whichever metadata provider you've already got configured (TMDB or TVDB).

Only a few properties get the fallback — the ones that genuinely mean the same thing across Sonarr, TMDB, and TVDB:

- `ended` (true/false)
- `firstAirDate` at show level
- `seasons` at show level — counts real seasons, no specials, matching what Sonarr's `seasonCount` does

I deliberately skipped a few that look similar but actually differ between sources:

- `status` — Sonarr uses a lowercase enum (`continuing`/`ended`/…); TMDB returns free-form strings like `Returning Series`; TVDB returns `Continuing`/`Ended`/…
- `originalLanguage` — Sonarr gives `English`, TMDB gives `en`, TVDB gives `eng`
- `rating` — different scales and aggregations on each side

Letting any of those through would mean rules silently mis-evaluate when the fallback kicks in. Anything that's only in Sonarr's local state (`monitored`, `tags`, `filePath`, disk-size stuff, the file-derived properties, …) keeps returning null like before — providers can't supply that.

A few mechanical bits worth flagging:

- Sonarr is still hit first; the fallback only runs after a confirmed miss, and the existing `ArrLookupCache` still dedupes that miss across the run.
- TVDB's extended series response returns season entries for every alternative ordering (Aired / DVD / Absolute / Alternate / Regional), so I added typing for `seasons[]` and the count filters to the series' default ordering before dropping Season 0. Otherwise the raw length over-counts wildly.
- The existing "not in Sonarr/Radarr" warnings now end with "Is the series/movie tracked in Sonarr/Radarr?" so users have a clear next step instead of a cryptic ID dump.

Tests cover the status mapping in both providers, the specials filter, and the fallback path through `SonarrGetterService`. Full suite passes (1513 tests); format / lint / build all clean.

### DIFF
--- a/apps/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
+++ b/apps/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts
@@ -109,7 +109,14 @@ export class SonarrApi extends ServarrApi<{
   // (unmonitor + file deletes) by the empty-show cleanup, and also drives
   // rule evaluation — both need Sonarr's current truth, not a snapshot that
   // can be up to DEFAULT_TTL stale (see issue #2757 / #2891).
-  public async getSeriesByTvdbId(id: number): Promise<SonarrSeries> {
+  // Returns `null` when Sonarr confirms the series isn't tracked (empty
+  // response) and `undefined` when the lookup itself failed (transport, auth,
+  // 5xx). Callers must keep these distinct: a confirmed miss is safe to fall
+  // back from, a failure must fail closed so a transient Sonarr outage can't
+  // silently change rule evaluation.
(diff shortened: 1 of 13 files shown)


enoch85:
Tested and works in prod!

PR #3029, opened 2026-06-03
feat(server): add /api/health endpoints with DB liveness check

Kartalops:
## Summary
- Add a `HealthController` under `apps/server/src/app/health.controller.ts` exposing three endpoints:
  - `GET /api/health` — combined health check, 200 with `database: "ok"`, 503 with `database: "unreachable"`.
  - `GET /api/health/live` — liveness, 200 as long as the process is up, no DB call. Use for Kubernetes `livenessProbe`.
  - `GET /api/health/ready` — readiness, runs `SELECT 1` against the configured TypeORM datasource, 503 on failure. Use for Kubernetes `readinessProbe` and Docker `HEALTHCHECK`.
- Register the controller in `AppModule`.

## Why
The application has no `/health` endpoint today, which makes it awkward to wire up Kubernetes liveness/readiness probes or Docker `HEALTHCHECK`. `/api/app/status` exists but it also exercises the GitHub releases API, so a slow GitHub response would make the app look unhealthy to a load balancer.

## How tested
- `yarn install` — OK.
- `yarn check-types` (Turbo) — OK across all four packages.
- `yarn workspace @maintainerr/server lint` — OK, no warnings.
- `yarn workspace @maintainerr/server tsc --noEmit` — OK.
- No automated test added in this PR; the controller is small enough to be verifiable by hitting the endpoints once the app is up. A follow-up unit test for the `unreachable` branch (mock the datasource to throw) is straightforward and can be added if reviewers want it.

## Notes
- The DB check uses the same TypeORM datasource the rest of the app uses, so a misconfigured connection string or a full disk surfaces as a clear 503 rather than a silent 200.
- No new dependencies. The controller relies on the existing `@nestjs/typeorm` and `typeorm` packages.
- The endpoints are mounted under `/api/health` to follow the existing `/api/*` prefix convention used elsewhere in the app.


enoch85:
Hey! Thanks for the PR. Valid feature - there's genuinely no health endpoint today and `/api/app/status` reaches the GitHub API, so it's a poor probe. A few things to fix before it's review-OK, roughly in order of importance.

**Should fix first**

- **Controller talks to the DB directly.** No other controller in the codebase injects `DataSource`/repositories — all five do it in a *service* (collections, notifications, rules, settings, database-download). Here `dataSource.query('SELECT 1')` runs inline in the handler, which contradicts the documented "controllers delegate to services" pattern. Move the ping into a small `HealthService` (or a `pingDatabase()` on the existing `AppService`, which already owns app-level status) and let the controller just shape the response.
- **No test.** The unreachable/503 branch is the only non-trivial logic and is exactly what you offered to test. Easy to mock the datasource to throw — and trivial once the logic lives in a service. Matches the existing `*.controller.spec.ts` convention.
- **Swallowed error.** `catch {}` discards the actual DB failure, so operators get `database: "unreachable"` with no reason in the logs. Caught throwables in injectable code should go to `logger.debug(error)` via `MaintainerrLogger`.

**Cleanups**

- **`uptimeSeconds`** measures controller-instantiation time (`startedAt = Date.now()`), not process uptime. Use `process.uptime()` — it's accurate and lets you drop both `startedAt` and the `uptime()` helper.
- **Dead assignment** — in `ready()`, `database = 'unreachable'` is set then only read inside the thrown object literal; collapses out once you log + throw directly.
- **`/api/health` and `/api/health/ready` are identical** (`health()` just calls `ready()`). Harmless, but the combined endpoint adds little over `ready`.
- **`live()` returns `{ status: 'ok' }`** while the others return the fuller `HealthResponse` — minor envelope inconsistency for a client parsing one shape.

**Consistency nits**

- Root route: every other controller uses `@Get()`; this uses `@Get('/')`. Functionally identical, just inconsistent.
- `HealthResponse` is declared inline — response DTOs conventionally live in `@maintainerr/contracts` (e.g. `VersionResponse`).

**Nice to close the loop**

- Doesn't wire up the Dockerfile `HEALTHCHECK` it justifies itself with. The endpoint is the prerequisite, but shipping the actual `HEALTHCHECK` directive (or noting it as a follow-up) would finish the story.

For what it's worth, *not* pulling in `@nestjs/terminus` and using a plain `SELECT 1` are both the right calls here — consistent with preferring the existing idiom over a new dependency.

enoch85 (approved):

enoch85:
Thanks for this. Before merging I pushed a follow-up to address the review points:

- The database check now lives in a dedicated `HealthService`; the controller only shapes the response, in keeping with the rest of the codebase where controllers delegate to services.
- The caught database error is logged via `MaintainerrLogger` instead of being swallowed, so an `unreachable` result leaves a trace in the logs.
- Uptime is reported from `process.uptime()`, and the response shapes (`LivenessResponse`, `HealthResponse`) were moved into `@maintainerr/contracts`. `/live` now returns the same envelope as the other endpoints, and the root route uses `@Get()`.
- Added unit tests for both the service and the controller, covering the reachable and unreachable paths.
- Wired up the Docker `HEALTHCHECK` to `/api/health/ready` (BASE_PATH-aware), which was the original motivation for the endpoint.

Readiness intentionally checks only the database - connected services (media servers, \*arr, Seerr, Tautulli) are deliberately left out so a flaky integration can't take the whole instance out of rotation.

Merged - thanks for the contribution.

PR #3030, opened 2026-06-03
feat(logging): honour LOG_LEVEL env var on startup

Kartalops:
## What

The winston factory now reads an optional `LOG_LEVEL` env var. When set to one of `fatal|error|warn|info|verbose|debug` it overrides the value persisted in the `log_settings` row for the lifetime of the process. Unknown values produce a single warning at startup and fall back to the persisted level. When the env var is unset, behaviour is identical to before.

The factory also tolerates a missing `log_settings` row during the first-boot migration window.

## Why

Today the only way to change verbosity is to flip it in the UI, which writes the database. That is awkward in two common ops situations: triage a single container with extra debug logs, and keep an unattended container on a quieter level without baking it into the DB. The env override is the standard knob for both.

## How

- Read `process.env.LOG_LEVEL`, trim, lower-case.
- If it matches the allowed set, use it; otherwise warn once and fall through.
- Default to the persisted row's `level`, or `'info'` if the row hasn't been created yet.
- Apply the same null-safety to `max_size`/`max_files` so the factory does not throw on a fresh DB.

## Notes

- Pre-existing TypeScript strict-mode warnings in the logging module are untouched (decorators, `logSettings` narrowing, etc.) — they are not introduced by this change.

### DIFF
--- a/apps/server/src/modules/logging/logLevel.spec.ts
+++ b/apps/server/src/modules/logging/logLevel.spec.ts
@@ -1,0 +1,33 @@
+import { LOG_LEVELS } from '@maintainerr/contracts';
+import { resolveLogLevel } from './logLevel';
+
+describe('resolveLogLevel', () => {
+  it('uses the persisted level when LOG_LEVEL is unset', () => {
+    expect(resolveLogLevel(undefined, 'warn')).toEqual({ level: 'warn' });
+  });
+
+  it('lets a recognised LOG_LEVEL override the persisted level', () => {
+    expect(resolveLogLevel('debug', 'info')).toEqual({ level: 'debug' });
+  });
+
+  it('normalises case and surrounding whitespace', () => {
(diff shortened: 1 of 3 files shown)


enoch85:
Took this over and pushed a cleaned-up version. Summary of what changed:

- Wired the resolved level into `createLogger`. The env value was being computed but never applied, so `LOG_LEVEL` had no effect before. It now actually overrides the persisted level.
- Made the factory null-safe for a missing `log_settings` row on first boot, so it no longer throws before the row exists.
- Replaced the hardcoded fallbacks with the shared `DEFAULT_LOG_*` constants, and derived the allowed-level list from `LOG_LEVELS` in contracts instead of a local set.
- Extracted the resolution into a small `resolveLogLevel` helper with unit tests.

The branch is now logging-only - the health endpoint changes were dropped since they already landed via #3029. Full suite passes. Thanks for the original work, kept you credited on the commit.

PR #3031, opened 2026-06-03
fix(notifications): validate webhook URL scheme before posting

Kartalops:
## What

The webhook agent passed `settings.options.webhookUrl` straight to `axios.post`. A misconfigured URL pointing at `file://`, `gopher://`, or any other non-http(s) scheme would be turned into an outbound request by the agent, with axios's behaviour for those schemes varying across Node versions.

## Why

This is a small but real SSRF-shaped guardrail: it keeps a webhook agent pointed at an internal file path or an unexpected scheme from being the way the server leaves the network. The user can still disable the agent entirely; this only narrows what is allowed when it is enabled.

## How

- Parse the configured URL with the platform `URL` constructor; on failure log and return `Failure: invalid webhook URL`.
- Reject anything whose protocol is not `http:` or `https:`; log and return `Failure: unsupported webhook URL scheme`.
- Pass the parsed (and therefore normalised) URL string to `axios.post` instead of the raw user input.

## Notes

- Pre-existing TypeScript strict-mode warnings in the logging/notifications modules are untouched — this change introduces no new ones.

### DIFF
--- a/apps/server/src/modules/notifications/agents/discord.spec.ts
+++ b/apps/server/src/modules/notifications/agents/discord.spec.ts
@@ -18,14 +18,14 @@ jest.mock('axios', () => ({
 describe('DiscordAgent', () => {
   const webhookUrl = 'https://discord.com/api/webhooks/123/abc';
 
-  const createAgent = () => {
+  const createAgent = (url: string = webhookUrl) => {
     const notification = new Notification();
     const settings: NotificationAgentDiscord = {
       enabled: true,
       types: [NotificationType.TEST_NOTIFICATION],
       options: {
         agent: NotificationAgentKey.DISCORD,
-        webhookUrl,
+        webhookUrl: url,
       },
     };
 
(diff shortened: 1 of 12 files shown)


enoch85:
Took this over and pushed an updated version. Summary of what changed:

- Extracted a shared `validateWebhookUrl` helper instead of guarding only the generic webhook agent. The original change left the other agents posting an unvalidated URL.
- Applied the guard in every agent that posts to an operator-supplied URL: the webhookUrl agents (webhook, slack, lunasea, discord) and ntfy's server url. Unparseable URLs and non-http(s) schemes are rejected before any request, and the normalised URL is posted.
- Replaced ntfy's regex slash-trimming with the codebase's char-based idiom, per the no-regex convention.
- Added tests - branch coverage for the helper, plus a wiring test per agent confirming a bad scheme returns a Failure and never calls axios.

Verified end to end via `POST /api/notifications/test`: across all five agents, bad schemes (file://, gopher://) and invalid URLs are rejected with no outbound request, while a valid http URL posts as expected. For ntfy, a trailing-slash URL and leading-slash topic join correctly to a single path.

The branch is now notifications-only - the health and logging commits were dropped since they already landed via #3029 and #3030. Full suite passes. Thanks for the original work, kept you credited on the commit.

enoch85 (approved):

PR #3054, opened 2026-06-06
feat: download client cleanup for Radarr/Sonarr deletions (qBittorrent)

enoch85:
Adds an optional download-client integration that removes the completed download (and optionally its data) from qBittorrent when Radarr/Sonarr deletes media — cleaning up the orphaned, still-seeding download left behind after the library copy is removed.

- Matched via the Radarr/Sonarr download history (`downloadId`), so it only runs for media removed through Radarr/Sonarr. Sonarr is scoped to whole-show deletes (season-pack safety).
- Whether a download has finished seeding is decided by qBittorrent's own ratio / seed-time limits; a fallback ratio (default 0.5, min 0.5) applies only to downloads qBittorrent doesn't limit.
- qBittorrent deletes its own downloaded files, so the common "downloads separate from the library" (hardlink/copy) setup is cleaned without Maintainerr needing any paths. Cross-seed-safe: a shared content path is removed entry-only.
- Settings tab shown only when Radarr/Sonarr is configured; handles qBittorrent's auth-bypass and surfaces an actionable message on a 403.
- Backend sits behind a `DownloadClient` interface + factory so additional clients can be added later.

Includes a TypeORM migration for the new settings columns plus unit and UI tests. Docs (a Configuration section) are prepared as a separate Maintainerr_docs change.


enoch85:
**Reason for dismissing alerts**

Requests intentionally target an administrator-configured download-client URL, exactly like Maintainerr's existing Plex/Jellyfin/Emby/Radarr/Sonarr/Seerr/Tautulli/Streamystats integrations. Maintainerr is self-hosted and by design connects to user-configured services on private networks (localhost/RFC1918); blocking those is not viable (see serviceUrlSchema), and settings endpoints are protected at the network level. Not an exploitable SSRF in this threat model.

Issue #3010, opened 2026-06-01
Maintainerr incorrectly overrides correct TVDB ID via TMDB cross-reference


lluisd:
**Describe the bug**
I have an old tv show in my jellyfin with correct metadata which Maintainerr tries to correct the TVDB id with a wrong one. 

**To Reproduce**
On test over that tv show season from my collection rule.

**Expected behavior**
No needed to correct the TVDB ID.

**Screenshots**

**Device (please complete the following information):**
 - OS: Using docker over Synology DSM 7.3
 - Version  3.13.0
 - Server: Jellyfin

**Debug logs**
```
01/06/2026 13:16:18 [WARN] [MetadataService] Corrected TVDB ID: 280331 to 306261 via TMDB cross-reference. The media server may have incorrect metadata for this item.
01/06/2026 13:16:38 [WARN] [MetadataService] Corrected TVDB ID for "Un principe para...": 280331 to 306261 via TMDB cross-reference. The media server may have incorrect metadata for this item.
01/06/2026 13:16:38 [WARN] [SonarrApi] Could not retrieve show by tvdb ID 306261
01/06/2026 13:16:38 [WARN] [SonarrGetterService] None of the resolved external IDs [TVDB:306261] for 'Un principe para...' matched a series in Sonarr. Is the series tracked in Sonarr?
```

**Additional context**
The TV has IMDB: [tt3648600](https://www.imdb.com/es-es/title/tt3648600/), TMDB: [65351](https://www.themoviedb.org/tv/65351-un-principe-para-tres-princesas?language=es-MX) and TVDB: [280331](https://thetvdb.com/series/280331-show) and release year: 2013 which all are correct in jellyfin. 

Jellyfin metadata:

```json
{
	"Name": "Un principe para...",
	"OriginalTitle": "Un principe para...",
	"ServerId": "907bee2464264fa4b9501e19a8b93251",
	"Id": "bf3ba9f6c02351e96ec5085cb45191cd",
	"DateCreated": "2024-11-14T20:31:44.0000000Z",
	"PremiereDate": "2013-05-12T00:00:00.0000000Z",
	"Path": "/tvshows/un principe para Corina (2013)",
	"ChannelId": null,
	"RunTimeTicks": 72000000000,
	"ProductionYear": 2013,
	"ProviderIds": {
		"Imdb": "tt3648600",
		"Tmdb": "65351",
		"Tvdb": "280331",
		"TvdbSlug": "280331-show"
	},
[...]

PR #3056, opened 2026-06-07
feat(metadata): write corrected release date back to Plex (opt-in)

enoch85:
Opt-in, off by default. Builds on the #3010 fix (#3011, now merged): when two metadata providers agree on a release year the media server got wrong, optionally write the agreed provider's full release date back to the media server — so the source, and everything downstream (the *arrs, other clients), is corrected, not just Maintainerr's in-memory resolution.

**Plex only, by design.** Plex locks a single field (`originallyAvailableAt.locked=1`) so its agent won't revert the correction. Jellyfin/Emby have no per-field date lock (the `MetadataField` enum has no date entry — verified in `@jellyfin/sdk`); the only option there is locking all metadata, which is unacceptable. So write-back is gated to Plex via a new `RELEASE_DATE_WRITEBACK` capability; Jellyfin/Emby keep the per-run in-memory correction only. This avoids the chicken-or-egg where the server's own agent overwrites us again.

- New opt-in setting `metadata_writeback` (server + UI toggle on the Metadata page) + migration.
- `setReleaseDate` on the media-server abstraction (Plex locks the field).
- Writes the provider's full release date; skips when only a year is known (never clobbers day/month); de-duped per process.

Migration verified the prescribed way: v3.14.0 `yarn dev` → branch `yarn dev` applies it on boot → `migration:generate` reports "No changes".


enoch85:
Closing this — after testing it end-to-end against a real Plex, we've decided to drop the release-date writeback. Writing **and locking** `originallyAvailableAt` based on a provider-agreement heuristic is too high-risk for a maintenance tool (a wrong correction pins a wrong date), it requires a DB migration for a single opt-in flag, and it only writes reliably where Maintainerr can reach Plex directly. The provider-agreement year correction from #3011 already gives the rule-evaluation benefit without mutating Plex.

The genuinely useful fixes discovered on this branch were split out into #3063 (connection probe via `/identity`, v2 `/api/v2/resources` server discovery, and a UI banner cleanup).

PR #3072, opened 2026-06-09
perf: bulk-prefetch Plex watch history before rule evaluation

stormshaker:
This is a follow-up PR to https://github.com/Maintainerr/Maintainerr/pull/2972 to further improve Plex performance on large libraries, using a watch history bulk fetch.

## Summary

Replaces O(items × conditions) per-item HTTP calls to Plex's `/status/sessions/history/all?metadataItemID=<id>` with a single paginated bulk fetch at the start of each rule group execution.

**Measured call times (local Plex, same NAS):**
- Per-item call: ~408 ms (Plex executes a SQLite query per request even on localhost)
- Full prefetch (27,205 records): ~67 s cold — builds all three maps in one pass

**Benchmark: vanilla 3.14.0 vs this PR (same NAS, manual trigger, 2026-06-10):**

| Run | Rule | Vanilla 3.14.0 | This PR | Saving |
|-----|------|----------------|---------|--------|
| Cold (first in cron window) | Movies Leaving Soon | 22m 48s | **1m 25s** (67s prefetch + 18s eval) | **−94%** |
| Same window | TV Leaving Soon | 6m 02s | **1m 09s** (reuses cached maps) | **−81%** |
| Warm (repeat in same window) | Movies Leaving Soon | 22m 48s | **17s** | **−99%** |
| Warm (repeat in same window) | TV Leaving Soon | 6m 02s | **47s** | **−87%** |

**Correctness:** patched and vanilla produced identical match sets — the same 10 movies and 152 shows, verified by comparing `mediaServerId` sets from the DB after each run. Re-verified after every revision of this PR.

## How it works

- `prefetchWatchHistory()` fetches the complete watch history once per rule group run and builds **three maps** in a single pass:
  - **leaf map** — movies and episodes keyed by own `ratingKey`
  - **show map** — episode records grouped by show `ratingKey` (from `grandparentKey`)
  - **season map** — episode records grouped by season `ratingKey` (from `parentKey`)
- The maps live in a **dedicated `plexwatchhistory` cache** (1 h TTL). It is `persistent` so the maps survive the per-group `flushAll()`, and it is flushed alongside the other Plex caches whenever a rule group requires a cache reset — so "fresh data" rules rebuild the snapshot. No behaviour of the existing `plexguid` HTTP-response cache is changed.
- `getWatchHistory()` gains an optional `itemType` and routes typed lookups to the correct map; a typed miss is a confirmed never-watched `[]` with no HTTP call. It always returns a **copy** of the cached array, so callers that sort in place cannot corrupt the cache. Untyped callers (e.g. the watch-history API route, which may receive any kind of ratingKey) fall through to the per-item query on a miss.
- `useCache: false` callers (`getWatchState`) are served from the snapshot too: that flag opts out of the stale per-item HTTP response cache, whereas the snapshot's freshness is governed by the rule-run lifecycle above. The item type is threaded through `getWatchState` so never-watched lookups stay in the map.
- Jellyfin/Emby adapters don't implement the optional `prefetchWatchHistory?()`; the executor's optional-chain call is a no-op and their per-item paths are untouched.

## Test plan

- [x] Full suite passes (1,697 server tests) — new coverage for map building, routing, copy-on-read mutation safety, untyped fallthrough, cache-reset flush, concurrent prefetch dedup, and Jellyfin/Emby no-op
- [x] Benchmarks and identical-match-set verification above, running in production for multiple days across PR revisions

## Incidental correctness fix (separate commit)

While testing this PR against a large Plex TV library I found a **pre-existing bug, unrelated to caching**, and fixed it in a standalone commit (5e020dc) so it can be reviewed or cherry-picked independently:

`PlexGetterService` `sw_lastWatched` did `return watchHistory ? new Date(+watchHistory[0].viewedAt * 1000) : null`. An empty array is truthy, so a never-watched show read `viewedAt` off `undefined` and threw `TypeError: Cannot read properties of undefined (reading 'viewedAt')`. The rule comparator catches it and skips the condition, so **match results are unaffected** — but it logs roughly one error per never-watched show evaluated (~900 per run here). The fix guards on `watchHistory?.length`. This exists in current `development` too; happy to split it into its own PR if you'd prefer.

### DIFF
--- a/apps/server/src/modules/api/lib/cache.ts
+++ b/apps/server/src/modules/api/lib/cache.ts
@@ -4,6 +4,7 @@ type AvailableCacheIds =
   | 'tmdb'
   | 'tvdb'
   | 'plexguid'
+  | 'plexwatchhistory'
   | 'plextv'
   | 'seerr'
   | 'plexcommunity'
@@ -24,6 +25,11 @@ type CacheOptions = {
   // If true, this cache is NOT flushed by flushAll(). Use for external metadata
   // caches (e.g. TMDB, TVDB) whose data doesn't change between rule group runs.
   persistent?: boolean;
+  // If false, NodeCache stores and returns object references without cloning.
+  // Required for non-POJO values (Maps, Sets) and for high-frequency lookups
+  // where cloning large objects on every get() would be prohibitively expensive.
+  // Never mutate values returned from caches that set this to false.
+  useClones?: boolean;
 };
 
 export class Cache {
@@ -46,6 +52,7 @@ export class Cache {
     this.data = new NodeCache({
       stdTTL: options.stdTtl ?? DEFAULT_TTL,
       checkperiod: options.checkPeriod ?? DEFAULT_CHECK_PERIOD,
+      useClones: options.useClones ?? true,
     });
   }
 
@@ -71,6 +78,20 @@ class CacheManager {
       persistent: true,
     }),
     plexguid: new Cache('plexguid', 'Plex GUID', 'plexguid'),
+    // Holds the bulk watch-history maps built by PlexApiService.prefetchWatchHistory.
+    // Persistent so the maps survive flushAll() between rule groups in the same
+    // cron window; useClones is off because the values are large Maps —
+    // getWatchHistory returns copies of the per-item arrays instead.
+    plexwatchhistory: new Cache(
+      'plexwatchhistory',
+      'Plex watch history',
+      'plexwatchhistory',
+      {
+        stdTtl: 3600, // 1 hour
+        persistent: true,
+        useClones: false,
+      },
+    ),
     plextv: new Cache('plextv', 'Plex.tv', 'plextv'),
     seerr: new Cache('seerr', 'Seerr API', 'seerr'),
     plexcommunity: new Cache(
--- a/apps/server/src/modules/api/media-server/media-server.interface.ts
+++ b/apps/server/src/modules/api/media-server/media-server.interface.ts
@@ -160,6 +160,17 @@ export interface IMediaServerService {
   searchContent(query: string): Promise<MediaItem[]>;
 
   /**
+   * Prefetch watch history for all library items in a single bulk request,
+   * caching the result so that subsequent per-item getWatchHistory / getWatchState
+   * calls can be served from memory instead of making individual HTTP requests.
+   *
+   * This is an optional optimisation — implementations that do not support bulk
+   * fetching may omit it. If absent, rule evaluation falls back to per-item
+   * queries as before.
+   */
+  prefetchWatchHistory?(): Promise<void>;
+
+  /**
    * Get watch history for a specific item.
    * Implementation varies by server:
    * - Plex: Single API call to history endpoint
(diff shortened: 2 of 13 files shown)


enoch85:
@stormshaker Are you running this in your prod?

enoch85:
@stormshaker Just had a quick look and there are multiple issues in this PR. Sorry, this is still sloppy. Not holding the quality I'd expect.

stormshaker:
@enoch85 Yes, this is running in my prod. I've deliberately left my rules unchanged so I can get comparative timing and accuracy on matches. I only have Plex, so I can't test Jellyfin/Emby.

Leave this with me, I'll do some more code review.

enoch85:
I don't know which IDE you're using, but you're not following repo rules. Just a hint. 👍🏼

enoch85:
Just reading the commit messages here. You are changing behaviour. 

You have to think about all the thousands of users who have rules setup already, you can't just do one-offs. We prefer no migration if possible, meaning, no behaviour change if strictly not needed.

enoch85:
Also, if it's only for Plex, you need to implement it as a feature.

stormshaker:
Understood. I'll continue to use my own fork on my prod.

enoch85:
> Understood. I'll continue to use my own fork on my prod.

Sure!

Here's the AI review from yesterday if you're interested. Will remove it soon: https://gist.github.com/enoch85/1ee6a377e71983be12ab292399f185eb

Bottom line, follow the rule set out here in this repo and you will probably be better off. 👍🏼

enoch85:
Here you go: https://github.com/Maintainerr/Maintainerr/tree/development/.claude

stormshaker:
@enoch85 OK, I can see what went wrong. You've gone to the effort of adding AI instructions on the repo, which is excellent. But, because I'm working not only with a fork but with a separate project repo, they weren't sourced. Your jabs about the rules and a feature made perfect sense after reading them. I was a bit miffed after spending on and off for the last couple of weeks getting to this PR.

So - all good now. I see you even created a new PR with a sample implementation - I'll make sure that my AI reads the rules, reads your PR, and we run PR #3082 in my prod to check for bugs and performance. There's a very significant performance improvement to be had on the prefetch. It took me 6+ hours to run rules before I started this optimisation, which made iterating on rule development very tedious. It looks like we'll be able to get rule execution down to 10 minutes or less by adding the previous Sonarr/Radarr and now the Plex prefetch for watch history. Give me some time to test, and we'll get a good PR. Thanks for your patience.

enoch85:
@stormshaker While AI is a good *tool* it will never replace human finesse. You can't just blindly trust that everything AI spits out is good or even valid. You have to understand it, and put your critical glasses on - and reject suggestions it makes even if they sound good. There are *so many* times it (AI) was dead certain that something was 100% correct, just to later admit it was wrong. Without understanding the code, and just blindly trust it - you will end up in a complete mess.

What you made here felt like "I want this, please make it - good, now send a PR". I won't care much for those types of PRs. They are pure slop to be honest. 

I use AI very much myself, but in a controlled way. I review the edits, I read the code. I'm critical. This app is for thousands of users. You need to respect that. Think of everyone.

Just a few pointers. :+1:

PR #3082, opened 2026-06-10
perf: bulk-prefetch Plex watch history (leaf items) before rule evaluation

enoch85:
Builds on #3072 by @stormshaker — same bulk watch-history prefetch, scoped to leaf items only.

Replaces O(items × conditions) per-item Plex `/status/sessions/history/all` calls with a single paginated sweep at the start of each rule group, cached in a dedicated `plexwatchhistory` cache. Movies and episodes are served from the leaf map keyed by their own `ratingKey`; show/season history falls through to the per-item `metadataItemID` query, which Plex rolls up server-side.

- **Leaf-only.** The show/season rollup keyed on `grandparentKey`/`parentKey`, which are undocumented on the history endpoint and absent over some Plex connections (verified on a real server) — a missing key silently read as "never watched", risking false deletes. Own `ratingKey` is documented and connection-stable; show/season use the server-side rollup instead.
- **Dedicated cache.** The leaf map lives in its own `plexwatchhistory` cache (`useClones: false`), leaving the shared `plexguid` cache at its non-persistent defaults so general Plex reads stay cold-start per run. The snapshot is rebuilt fresh per scheduler batch (flushed at batch end) and on Plex re-init, with a 1 h TTL as a backstop; groups within a batch share one sweep. `getWatchHistory` returns copies, so callers sorting in place can't mutate the snapshot.
- **Abstraction.** `prefetchWatchHistory` is a required `IMediaServerService` method gated by `supportsFeature(CENTRAL_WATCH_HISTORY)`; Jellyfin/Emby (per-user history) keep their per-item path. The prefetch is abort-checked so a cancellation can't kick off a long history sweep.

Credit to @stormshaker for the original implementation and benchmarking in #3072.


enoch85:
@stormshaker I won't merge this anytime soon. But something like this is how I would do it.

enoch85:
This PR tested with around 13 000 items:

image-3

enoch85:
@stormshaker after I worked the whole weekend with replacing the server for *.maintainerr.info, we now also run the whole dev environment there. That means real Plex, Jellyfin, Emby and all services connected. 

With the new server it's now also possible to cross test everything, benchmark and do all sorts of things - which I did.

The result being that I feel pretty confident about this PR now, and will merge it. 🚀

enoch85:
## Quick benchmark

Right now, when rules run, we ask Plex about each item's watch history one at a time. On a big library that's thousands of separate requests — which is exactly what hammers Plex and causes the timeouts.

This PR grabs the whole history in one shot up front, keeps it in memory, and looks each item up from there.

I tested it against a real Plex server and counted the actual requests. Times below assume a remote Plex (~100ms per request):

| Library | Requests before | Requests after | Time before | Time after |
|---|---|---|---|---|
| 500 items | 500 | 7 | ~6s | ~1s |
| 2,000 items | 2,000 | 34 | ~25s | ~3s |
| 13,000 items | 13,000 | 167 | ~2.5 min | ~17s |

In my small test, 10 items went from 10 requests down to 1.

One thing left on purpose: the "is it watched?" check still asks Plex live for each item, so something you just watched can't get deleted by mistake.

PR #3170, opened 2026-06-24
fix: bulk-prefetch Seerr requests for rule evaluation (#3152)

### LINKED ISSUE #3152: Seerr getter returns null per-item under load on large libraries — Seerr-seeded rules silently match almost nothing
## Describe the bug

On a large library, rules that depend on a Seerr property (e.g. `Seerr - Requested in Seerr`) match almost nothing, because the Seerr getter returns `null` for most items during a full run. The comparator treats a `null` first value as a missing operand and skips the rule for that item, so a rule like `isRequested = false AND not watchlisted` ends up matching only the handful of items whose Seerr lookup happened to succeed that run.

## Root cause

`SeerrGetterService.get()` makes a per-item `seerrApi.getMovie()` / `getShow()` call. Under the request volume of a whole-library run these calls fail/rate-limit and the response comes back without `mediaInfo`, so the getter hits its `else` branch and returns `null` (`Couldn't fetch Seerr metadata for media '<title>' … no Seerr query could be made`).

The per-item-API-hammering problem was recognized and fixed in #2580 / #2936 by introducing bulk prefetch/caching — but that work covered the **Plex / Tautulli / Sonarr / Radarr** getters and **not** the Seerr getter. `getter.service.ts` still routes `Application.SEERR` straight to a per-item `seerrGetter.get()`, with no prefetch or cache layer.

## Evidence (v3.15.2, ~1,150-movie Plex library, Overseerr/seerr)

- A single run of a rule seeded on `Seerr: Requested = false` logged **278 unique titles** with `Couldn't fetch Seerr metadata … no Seerr query could be made` → `firstValue=null` → `Skipping rule comparison because a value is unavailable`. Only 2 items survived to the collection.
- The **same titles resolve correctly** when evaluated individually via `POST /api/rules/test` — e.g. `Shazam!` returns `isRequested: 0` in the single-item test, yet appears in the bulk-run failure log. This confirms it's load/rate-limit dependent, not missing/unmatched data.
- A functionally-equivalent rule reseeded on Plex-only properties (`Times viewed`, `Last view date`) with the *same* watchlist exclusion matched **155 items reliably** on the same library — i.e. the Plex and watchlist getters scale fine; only the Seerr getter does not.

## Expected behavior

The Seerr getter should resolve request data reliably on large libraries — e.g. fetch/cache it once per run (bulk prefetch, as the other getters now do), or at minimum add retry/back-off — so Seerr-seeded rules don't silently degrade to near-zero matches.

## Version

3.15.2 (self-hosted; Plex media server + Overseerr/seerr request service).

enoch85:
Rules seeded on a Seerr property (e.g. "Seerr - Requested") matched almost nothing on large libraries. The getter made a per-item `GET /movie|/tv` call per item; under a whole-library run those rate-limited, the getter returned a value the comparator skipped, and Seerr-seeded rules silently degraded to near-zero matches. Plex/Tautulli/Sonarr/Radarr already had run-scoped bulk prefetch/dedupe; Seerr did not.

- `SeerrApiService` now fetches all requests in one paginated `/request` sweep and builds a run-scoped index grouped by `media.tmdbId` (Seerr's canonical, non-null key), with in-flight `Promise` dedup, cached in a dedicated non-persistent cache flushed/rebuilt per rule-group run. Page size is 100 — Seerr imposes no `take` cap and uses OFFSET pagination, so a larger page than `getUsers`' 50 cuts round-trips and server re-scan.
- `SeerrGetterService` reads that index for every request-derived property (isRequested, amountRequested, requestDate, addUser, approvalDate, mediaAddedAt) at movie and season/episode level.
- A failed sweep yields `undefined` (transient — the comparator protects the item), never `null`; `releaseDate` keeps its documented per-item fallback (not available from `/request`).
- Adds `tools/dev/fake-seerr.mjs` and a shared `scale-library` fixture, and seeds Seerr rule groups, for end-to-end testing.
- Consolidates request-service naming to "Seerr" across code comments, UI strings and docs (the legacy `/overseerr` + `/jellyseerr` route aliases stay for backward compatibility).

Closes #3152


enoch85:
@area51tazz Feel free to test this on your library. In around 5 minutes there should be a PR tag you can use.

Please note that it's built on latest development and includes non reversible migrations, so if you want to revert you need to do so from a backup.

Would be nice with some confirmation though since I only tested with around 100 items, and that's not enough to really see the effects of this.

Thanks! 🙏🏼

enoch85:
Tested this back and forth now for several hours. I don't think it gets any better than this at the moment. 

I will merge it so that you can continue testing in `development` if you want @area51tazz. :+1:

PR #3200, opened 2026-06-30
feat: clean up leftover media folder after a per-file *arr delete

enoch85:
Opt-in (default off, marked BETA) cleanup of the folder Radarr/Sonarr strand when
they delete an item's files one by one, along with the sidecars left in it
(`.srt`, `.nfo`, artwork). Set **per collection** on the rule group, next to the
\*arr action it modifies, and only offered when the chosen action actually
strands a folder.

Whole-entity deletes are deliberately not cleaned: `DELETE /movie/{id}` and
`DELETE /series/{id}` already remove the folder in the *arr itself (verified
against upstream `MediaFileDeletionService` and a live Radarr/Sonarr). Only the
per-file paths leave a folder behind, and the *arr's own "Delete empty folders"
option does not cover them because a stray sidecar means the folder is not empty.

## Which collection actions clean up

| Library | Action | Folder cleaned |
| --- | --- | --- |
| Movies | Unmonitor and delete files | movie folder |
| Shows | Unmonitor show + seasons, delete all episodes | series folder |
| Shows | Unmonitor show, delete existing episodes | series folder |
| Seasons | Unmonitor and delete season | season folder |
| Seasons | Unmonitor and delete existing episodes | season folder |
| Seasons | Unmonitor and delete season + delete show if empty | season folder, unless the show was deleted too |

Everything else is out of scope and documented as such: `Delete` / `Delete entire
show` (the *arr removes the folder itself), episode-level deletes (the season
folder is shared with the episodes that are kept), Sportarr (a league carries no
folder path at all), and deletes that fall back to the media server.

## Notes

- New `LeftoverFolderCleanupService`, called inline at each per-file delete site
  (mirrors the `removeDownloads` idiom); best-effort, never throws.
- Fail-safe guardrails: a real subfolder strictly inside a known *arr root that
  held one of the just-deleted files, never a root or symlink, never at or above
  another tracked item, season strictly under its series folder, and removal only
  when every remaining entry is a recognized sidecar file. A media file, an
  unknown extension or a link keeps the folder.
- Every input those guardrails rest on is read uncached, right before the
  delete: an item added since the last cached read would otherwise be invisible
  to the fence. Sonarr resolves them from one `GET /episodefile?seriesId=` call,
  which carries each file's season and path.
- Removes only the verified files, then the directories bottom-up with a
  non-recursive `rmdir`, so a file that appears mid-run is never removed unseen.
- `leftoverCleanupScope` in `@maintainerr/contracts` is the single definition of
  which actions strand a folder: the UI uses it to decide whether to show the
  option, the handlers use it to decide whether to act, and the server clamps the
  stored flag with it so a hidden value is never left enabled.
- New `collection.cleanupLeftoverFolders` column and migration.
- Requires the media library mounted into Maintainerr at the same path the *arr
  reports; when it is not, cleanup no-ops and logs a warning saying so.
- Sidecars are deleted outright, not moved to the *arr's Recycle Bin.


enoch85:
Not sure I'm going to merge this, but keeping as draft anyway. FYI for people coming here with hopes. :)

PR #3229, opened 2026-07-05
fix(sonarr): delete episode file after a slow unmonitor PUT is confirmed

### LINKED ISSUE #3228: Sonarr episode DELETE: slow/timed-out PUT leaves episode unmonitored with file still on disk
## Describe the bug

When a Sonarr episode-file DELETE action runs against a slow-responding Sonarr instance, the action can land in a **half-applied state**: the episode is left `monitored=false` but its file is never deleted and remains on disk. The whole action is then reported as failed to the collection worker, even though Sonarr actually applied the unmonitor.

This affects **any** Sonarr episode DELETE collection action, not one specific rule.

## Root cause

`UnmonitorDeleteEpisodes` (in `apps/server/src/modules/api/servarr-api/helpers/sonarr.helper.ts`) performs two steps in order:

1. `runPut('episode/{id}', { monitored: false })`
2. `runDelete('episodefile/{id}')`

`runPut` (in `apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts`) returns `false` when the HTTP response is slow/times out. But Sonarr can have **already applied** the PUT server-side before the client deadline. When `runPut` returns `false`, the handler short-circuits and step 2 (`runDelete`) never fires.

Net result:

- Sonarr: `monitored=false` (PUT did apply)
- Sonarr: `hasFile=true` (delete never fired)
- Disk: the episode file is still present
- Maintainerr log: `[WARN] [SonarrApi] Failed to run PUT: /episode/{id}`

The action is reported as failed to the worker: *"the configured action could not be completed."*

## Observed occurrence

During a "Handle Collections" run that matched 30 episodes of a daily-broadcast series, 29 resolved cleanly (27 deleted, 2 self-healed via the "Plex item missing" path from #3094) and **1** landed in the half-applied state above — episode unmonitored, ~1 GB `.mkv` still on disk, the `Failed to run PUT` warning logged ~10s after the preceding delete (consistent with an axios timeout rather than a rejection).

Re-running "Handle Collections" clears the stuck item once Sonarr responds in time, which confirms it's a timing race, not a logic error in what to delete.

## Expected behavior

A slow/timed-out PUT should not silently skip the file delete. The unit of success should be *"file is gone,"* not *"the PUT response arrived quickly."*

## Suggested fix direction

Either of:

- **(a)** Treat a PUT timeout as *uncertain* and still attempt `runDelete`. `runDelete` is idempotent against an already-deleted file, so the worst case is a harmless no-op.
- **(b)** Re-verify the PUT outcome by refetching the episode before declaring failure, and drive the delete off the refetched `hasFile`.

Worth auditing the other `runPut` call sites before picking an approach — there are uses beyond episode-unmonitor, and the timeout-vs-applied ambiguity is general to `runPut`.

## Workaround

Re-run "Handle Collections." The stuck item retries cleanly when Sonarr responds in time.

## Additional context

Surfaced while dogfooding the `episodeFileRank` property (#3095, released in v3.17.0). This is **not** a regression from that feature — the rank change is in the rule *getter* (which items match), not the action handler. Before the property existed, the rule matched 0 items on this instance so the DELETE path never fired; now that it matches, the pre-existing timeout race in the action handler is visible.

jackemcpherson:
### Description & Design

**Problem.** When a Sonarr episode-file DELETE action runs against a slow Sonarr instance, it can leave an episode half-applied: `monitored=false` at Sonarr but the file still on disk. `runPut` returns `false` on a slow or timed-out response even when Sonarr already applied the unmonitor, so the handler skips the file delete. It also returns on the first such episode, so the remaining matched episodes in the same batch are never processed. The action is then reported as failed even though Sonarr applied the change.

**Approach.** In `UnmonitorDeleteEpisodes`, when a PUT reports failure, re-fetch the episode's live monitored state (uncached) before deciding. Delete the file only when Sonarr confirms `monitored === false`; otherwise leave it in place and log. This is deliberately fail-closed: an unconfirmed or failed lookup keeps the file, because deleting a still-monitored episode's file would trigger a re-download. The loop now accumulates a `success` flag instead of returning early, so one slow or failed episode no longer aborts the rest of the batch. Both choices mirror patterns already in this file: `getSeriesByTvdbId` reads uncached straight after a mutation, and `unmonitorSeasons` accumulates success across items.

**Trade-offs.** The fix stays in the caller; `runPut`/`runDelete` are shared by other call sites and their semantics are unchanged. The extra GET happens only on the failure path, so the happy path adds no requests. The same timeout-vs-applied ambiguity exists at other `runPut` call sites (e.g. `updateSeries`) and is out of scope here.

### Related issue

Fixes #3228

### AI-Assisted Development

- **AI-assisted:** drafting the fix and the five regression tests, following the existing conventions in the file.
- **Personally reviewed / validated:** read the full diff; ran the targeted `sonarr.helper` suite (32 pass, 5 new) plus the downstream `sonarr-action-handler` and `servarr-api` suites (124 pass); confirmed lint and typecheck clean; verified by test that the file is kept on an unconfirmed/failed verification (fail-closed) and that a failed PUT or delete no longer aborts the batch.
- **Fit with standards:** reuses the repo's uncached-after-mutation and accumulate-success patterns; no change to public types or shared helpers.

### Checklist

- [x] I have read the CONTRIBUTING.md document.

### How to test

1. Configure a Sonarr collection with a DELETE action matching on-disk episodes.
2. Make Sonarr respond slower than the client timeout so the unmonitor PUT times out at the client but still applies server-side (e.g. a proxy delay).
3. Run Handle Collections. Before this change the episode is left unmonitored with its file on disk and the action reports failure; after, the file is deleted once the unmonitor is confirmed and the remaining episodes still process.

Unit: `cd apps/server && yarn jest sonarr.helper`

### Additional context

Surfaced while dogfooding a separate rule-property feature; filed as #3228.

### DIFF
--- a/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
+++ b/apps/server/src/modules/api/servarr-api/common/servarr-api.service.ts
@@ -13,6 +13,12 @@ import {
   Tag,
 } from '../interfaces/servarr.interface';
 
+// Slow/underpowered *arr instances can take >10s to answer even simple reads;
+// allow more headroom than the shared 10s axios default before aborting
+// (#3181). Used for uncached reads whose failure would change action or rule
+// behavior.
+export const SLOW_INSTANCE_TIMEOUT_MS = 20000;
+
 export abstract class ServarrApi<QueueItemAppendT> extends ExternalApiService {
   static buildUrl(settings: DVRSettings, path?: string): string {
     return `${settings.useSsl ? 'https' : 'http'}://${settings.hostname}:${settings.port}${settings.baseUrl ?? ''}${path}`;
--- a/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.spec.ts
+++ b/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.spec.ts
@@ -1,5 +1,8 @@
 import { Mocked } from '@suites/doubles.jest';
-import { createRadarrMovie } from '../../../../../test/utils/data';
+import {
+  createRadarrMovie,
+  createRadarrMovieFile,
+} from '../../../../../test/utils/data';
 import { MaintainerrLogger } from '../../../logging/logs.service';
 import { RadarrImportListExclusion } from '../interfaces/radarr.interface';
 import { RadarrApi } from './radarr.helper';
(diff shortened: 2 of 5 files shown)


jackemcpherson:
For context: I hit this while working on the `episodeFileRank` rule property that shipped in v3.17.0 (#3095). It is not a critical bug. It only appears when Sonarr responds slower than the client timeout, and re-running Handle Collections clears the stuck item. I wanted to close it out before moving on to the next piece of work rather than leave it open. Open to adjusting the approach if you would prefer the verification handled differently.

mmazur519:
Will this fix prevent my issue from yesterday it is very similar there was a permission issue everywhere except for plex the show said it was deleted but all of the files were still on the server because it was created at a different time under a different permission model so sonnar was unable to delete the episode but nothing told me that maintainer failed to delete the files.

[https://discord.com/channels/1152219249549512724/1522920006273339442](url)

enoch85:
Thanks!

> let success = true;

Why do we need this? Why can't we just keep using true and false with `return`?

jackemcpherson:
The `let success` accumulator is what lets the loop finish the whole batch instead of bailing on the first problem episode.

With `return false` on the first failed PUT (or delete), the remaining episodes in `matchedEpisodes` never get processed, so one slow or failing episode strands the rest of the batch. That was the second half of #3228: a single slow unmonitor PUT aborted the entire run rather than just its own episode. Accumulating into `success` and `continue`-ing handles each episode independently while still returning an overall pass/fail to the collection worker. It matches `unmonitorSeasons` a few methods down, which already does `success = (await ...) && success` for the same reason.

If the inline flag reads as noisy, one alternative is to lift the per-episode work into a small boolean-returning helper and let the loop stay a one-liner:

```ts
let success = true;
for (const e of matchedEpisodes) {
  success = (await this.unmonitorAndDeleteEpisode(e, deleteFiles)) && success;
}
return success;
```

where `unmonitorAndDeleteEpisode` holds the verify-then-delete logic and returns whether that one episode fully succeeded. Same behavior (still per-episode, still no early abort), but the batch loop reads clearly and the accumulator shrinks to one line. I lean slightly toward this since it separates the per-episode logic from the iteration, though the inline version is fine too. I can push that refactor if you'd like it.

enoch85:
> The `let success` accumulator is what lets the loop finish the whole batch instead of bailing on the first problem episode.
> 
> With `return false` on the first failed PUT (or delete), the remaining episodes in `matchedEpisodes` never get processed, so one slow or failing episode strands the rest of the batch. That was the second half of #3228: a single slow unmonitor PUT aborted the entire run rather than just its own episode. Accumulating into `success` and `continue`-ing handles each episode independently while still returning an overall pass/fail to the collection worker. It matches `unmonitorSeasons` a few methods down, which already does `success = (await ...) && success` for the same reason.
> 
> If the inline flag reads as noisy, one alternative is to lift the per-episode work into a small boolean-returning helper and let the loop stay a one-liner:
> 
> ```ts
> let success = true;
> for (const e of matchedEpisodes) {
>   success = (await this.unmonitorAndDeleteEpisode(e, deleteFiles)) && success;
> }
> return success;
> ```
> 
> where `unmonitorAndDeleteEpisode` holds the verify-then-delete logic and returns whether that one episode fully succeeded. Same behavior (still per-episode, still no early abort), but the batch loop reads clearly and the accumulator shrinks to one line. I lean slightly toward this since it separates the per-episode logic from the iteration, though the inline version is fine too. I can push that refactor if you'd like it.

If you think that's the tighter and most trimmed version, then yes. :+1:

jackemcpherson:
Extracted the per-episode logic into `unmonitorAndDeleteEpisode`, so the batch loop is now the one-liner above. Behavior is unchanged (same unmonitor → verify → delete order, same warning), and the existing tests pass untouched.

enoch85 (approved):

PR #3256, opened 2026-07-12
feat(rules): add Tautulli played-by-user and last-played-date properties

pezzy-o:
### Description & Design

Problem: Maintainerr users need "played" metrics (user clicked play) distinct from "viewed" metrics (met Tautulli's watched threshold, default 85%, minimum 50%). Currently, "played" metrics are only available to Jellyfin users via e.g. playCount/sw_playCount. Plex users relying on Tautulli have no way to distinguish "user pressed play" from "user finished watching" - existing Tautulli properties (seenBy, lastViewedAt, sw_watchers, sw_lastWatched) all filter by the watched threshold. This PR adds equivalent "played" metrics for Tautulli users, making them available to Plex users.

Why this approach: Follows the existing pattern established by Jellyfin's playCount/sw_playCount (IDs 30/31) — append new properties at the end of the TAUTULLI application. Reuses the existing getHistoryForMetadata helper and getPlexUsernamesForIds resolver. Mirrors sw_watchers (UNION) and sw_lastWatched (latest season → latest episode) semantics so users can compare "played" vs "viewed" apples-to-apples.

How it works: Adds 4 new TAUTULLI properties (IDs 9–12):

playedBy (TEXT_LIST, MOVIE) - usernames with any play event
lastPlayedAt (DATE, BOTH) - most recent stopped timestamp
sw_playedBy (TEXT_LIST, SHOW) - UNION of users who played ≥1 episode
sw_lastPlayedAt (DATE, SHOW) - newest episode play date in latest season
Getter logic uses raw history entries without watched_status/percent_complete filtering. sw_playedBy collects all unique user_id across episodes (UNION). sw_lastPlayedAt sorts by parent_media_index desc then media_index desc (matching sw_lastWatched).

### Trade-offs / limitations:

- No per-user API filtering (fetches all history, filters in-memory) - acceptable as single-item history is small.
- No UI tooltips explaining "played" vs "viewed" distinction - relies on property naming and docs.
- sw_playedBy does not support "played ALL episodes" (intersection) - only UNION like sw_watchers. Could be added later if needed.

### Related issue

N/A

### AI-Assisted Development

AI-assisted: All code generation (constants, getter logic, unit tests, type exports) was AI-assisted based on detailed design discussions and codebase exploration.
Human review: I reviewed all logic, verified the "played vs viewed" distinction is correctly implemented (no threshold filtering), validated the UNION vs existing INTERSECT patterns, confirmed test coverage (10 new tests), and ensured lint/format/type checks pass.
Standards compliance: Follows existing getter patterns (constants structure, error handling returning undefined/[]/null), Jellyfin playCount precedent for property placement, conventional commit format, and project test patterns (streamystats-getter style unit tests + getter-property-coverage).

### Checklist

### How to test

Configure Tautulli in Maintainerr Settings (URL + API key)
Create a movie rule: First Value → Tautulli → select Played by (username) or Last played date
Create a show rule: First Value → Tautulli → select Users that played at least one episode or Newest episode played date
Evaluate rule against media items that were partially watched (< 85% or your threshold)
Compare results in Tautulli web UI: playedBy/sw_playedBy should return users who pressed play; seenBy/sw_watchers should be empty for those same items
Verify timestamps: lastPlayedAt/sw_lastPlayedAt reflect actual play events, not threshold crossings

### Additional context

Briefly discussed with devs in Discord

### DIFF
--- a/apps/server/src/modules/api/tautulli-api/tautulli-api.service.ts
+++ b/apps/server/src/modules/api/tautulli-api/tautulli-api.service.ts
@@ -46,7 +46,7 @@ interface TautulliHistory {
   total_duration: string;
 }
 
-interface TautulliHistoryItem {
+export interface TautulliHistoryItem {
   user_id: number;
   user: string;
   watched_status: number;
--- a/apps/server/src/modules/rules/constants/rules.constants.ts
+++ b/apps/server/src/modules/rules/constants/rules.constants.ts
@@ -1106,6 +1106,36 @@ export class RuleConstants {
           type: RuleType.TEXT_LIST, // return usernames []
           showType: ['show', 'season', 'episode'],
         },
+        {
+          id: 9,
+          name: 'playedBy',
+          humanName: '[list] Played by (username)',
+          mediaType: MediaType.MOVIE,
+          type: RuleType.TEXT_LIST, // returns usernames []
+        },
+        {
+          id: 10,
+          name: 'lastPlayedAt',
+          humanName: 'Last played date',
+          mediaType: MediaType.BOTH,
+          type: RuleType.DATE,
+        },
+        {
+          id: 11,
+          name: 'sw_playedBy',
+          humanName: '[list] Users that played at least one episode',
+          mediaType: MediaType.SHOW,
+          type: RuleType.TEXT_LIST, // return usernames []
+          showType: ['show', 'season', 'episode'],
+        },
+        {
+          id: 12,
+          name: 'sw_lastPlayedAt',
+          humanName: 'Newest episode played date',
+          mediaType: MediaType.SHOW,
+          type: RuleType.DATE,
+          showType: ['show', 'season'],
+        },
       ],
     },
     {
--- a/apps/server/src/modules/rules/getter/tautulli-getter.service.ts
+++ b/apps/server/src/modules/rules/getter/tautulli-getter.service.ts
@@ -43,6 +43,9 @@ export class TautulliGetterService {
   ) {
     try {
       const prop = this.appProperties.find((el) => el.id === id);
+      if (!prop) {
+        return null;
+      }
       const metadata = await this.tautulliApi.getMetadata(libItem.id);
       const collection = await this.collectionRepository.findOne({
         where: { id: ruleGroup.collection.id },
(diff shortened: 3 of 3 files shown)


enoch85:
Thanks for the write-up, and for disclosing the AI assistance. I tested this against a real Plex + real Tautulli: generated actual playback history and ran every getter through `POST /api/rules/test`.

I have to ask for changes.

> **Note:** I rebased your branch onto current `development` (your spec collided with the one now in `development`). Your original commit was `3b12ff38` if you need anything back.

## 1. We already support this

**"Tautulli watched percent override"**, in the same New Rule form. Set it to `0` and every Tautulli property stops using the watched threshold and counts every play instead - which is exactly "the user pressed play".

A movie played to 20%, and a show with two episodes played to 30% and 40%. Nothing crossed the 85% threshold:

| Existing property | default | override = 0 | Your new property | value |
| --- | --- | --- | --- | --- |
| `seenBy` | `[]` | `["user"]` | `playedBy` | `["user"]` |
| `lastViewedAt` | `null` | `16:38:27` | `lastPlayedAt` | `16:38:27` |
| `viewCount` | `0` | `2` | (none added) | - |
| `sw_watchers` | `[]` | `["user"]` | `sw_playedBy` | `["user"]` |
| `sw_lastWatched` | `null` | `16:38:57` | `sw_lastPlayedAt` | `16:38:57` |
| `sw_viewedEpisodes` | `0` | `2` | (none added) | - |

Every new property returns exactly what an existing property already returns with the override set.

The one thing the override cannot do is mix "played" and "viewed" in the **same** rule group, because it applies to the whole collection. That is a fair gap - but a much smaller one than 4 new properties, and the PR does not mention the override at all.

## 2. `sw_playedBy` is broken for episodes

You set `showType: ['show', 'season', 'episode']`, but the code walks *children*. An episode has no children, so the loop never runs and it returns an empty list without reading any history.

On a real episode played to 30%:

```
sw_playedBy  ->  []            (never calls get_history)
sw_watchers  ->  ["user"]      (correct)
```

An empty result here can wrongly qualify media for a delete collection.

## 3. `sw_playedBy` makes far too many API calls

`sw_watchers` gets the same answer with **one** call, because Tautulli returns a show's whole episode history at once (`grandparent_rating_key`). That is what `getHistoryForMetadata` is for.

Yours loops over every season and every episode. Counted through a proxy in front of Tautulli:

```
sw_watchers   -> 2 calls
sw_playedBy   -> 7 calls   (show with 1 season, 3 episodes)
```

That is `3 + seasons + episodes`. A 10 season / 200 episode show costs **213 calls instead of 2**, per item, per rule run.

The per-episode loop only exists in `sw_allEpisodesSeenBy` because "watched EVERY episode" genuinely needs per-episode data. A union does not. The whole case is:

```ts
case 'sw_playedBy': {
  const history = await this.getHistoryForMetadata(metadata);
  return this.getPlexUsernamesForIds([...new Set(history.map((el) => el.user_id))]);
}
```

That fixes the episode bug too.

`getUsers()` is also not needed - `getPlexUsernamesForIds` already filters against Plex's user list, so `users.filter(...)` is an extra call that can only lose users.

## 4. `sw_lastPlayedAt` carries an outdated `sw_lastWatched`

You copied `sw_lastWatched` before #3264 fixed it, so your version still has the old shape (the threshold filter ran on a copy and the result was discarded). Post-rebase it now sits directly under the corrected `sw_lastWatched` in the same file, same idea, two shapes, one wrong.

It is also redundant: per the table, it returns `sw_lastWatched` with the override at 0.

## 5. Remove the `if (!prop) return null` guard

`undefined` and `null` mean different things here. `undefined` from a getter means "the lookup failed, skip this item"; `null` means "no value, compare it". The comparator says so in a comment.

Previously an unknown property id threw, got caught, and returned `undefined` - safe. Now it returns `null`, so a `NOT_EXISTS` rule would evaluate to **true** and add the item to a delete collection.

It only exists to make the "returns null for unknown property id" test pass. Please drop both.

## Small stuff

- `lastPlayedAt` uses `.sort()` with no compare function on numbers - that is a text sort. Use `.sort((a, b) => b - a)`.
- `added_at: '1_700_000_000'` is a string with underscores; `+'1_700_000_000'` is `NaN`.
- `type TautulliHistoryItem = TautulliHistoryItemType` renames an import back to its own name.
- Your spec only covered `sw_playedBy` at show level, which is why the episode bug slipped through.
- `tools/docs-drift.mjs` watches `rules.constants.ts`. New keys need a matching PR to `Maintainerr_docs` (`Glossary.md` and `Rules.mdx`).

## Suggested way forward

Tell us what the override does **not** give you. If the answer is "I need played and viewed in the same rule group", the small version of this is **2 properties, not 4**:

- **Keep** `playedBy` and `sw_playedBy`, both one-liners over `getHistoryForMetadata` with no threshold filter. Correct at episode level, 1 API call.
- **Drop** `lastPlayedAt` and `sw_lastPlayedAt`. Both are existing properties with the override at 0.
- Add your tests to the existing `tautulli-getter.service.spec.ts` rather than replacing it.

Happy to help shape it. The idea is reasonable, it just needs to be smaller and reuse what is already there.

pezzy-o:
First off, thank you very much for the code review: I've learned a couple of things here, not least of all to thoroughly check existing functionality first (see below), but also around the codebase and how it works. Thank you for taking the time to explain.

In answer to your comments:

### 1. We already support this
You're right on this. When I said that I couldn't set the Tautulli override below 50%, I was talking about _in the Tautulli app itself_ - I hadn't seen the checkbox in Maintainerr rules for override (which does not have this restriction). I've set this to 0% as suggested and it does cover much of the use case. As you mentioned, the gap is that the override cannot mix "played" and "viewed" in the same rule group.

### 2. Bug: sw_playedBy is broken for episodes
I don't see the code mismatch you're talking about, but evidently sw_playedBy is flawed anyway (see #3).

### 3. sw_playedBy makes far too many API calls
Unnecessary loops / API calls noted.

### 4. sw_lastPlayedAt copies a property that was broken (now fixed)
Noted, and glad you picked up on the bug.

### 5. Please remove the if (!prop) return null guard
Noted.

### Small stuff
All noted.

### Docs
Thanks for pointing that out - happy to contribute to Docs too.

### Suggested way forward

For now, I'm happy to try using the Maintainerr-configured Tautulli override as a way of measuring who clicked play (and when - see below), at the expense of a customisable watched threshold in the same rule group. If I understand correctly:

- The intent of *playedBy* is covered by *seenBy* with the override
- The intent of *sw_playedBy* is covered by *sw_watchers* with the override
- The intent of *lastPlayedAt* is covered by *lastViewedAt* with the override
- The intent of *sw_lastPlayedAt* is covered by *sw_lastWatched* with the override

enoch85:
Confirmed - all four mappings are correct:

| Your property | Equivalent to |
| --- | --- |
| `playedBy` | `seenBy` + override 0 |
| `sw_playedBy` | `sw_watchers` + override 0 |
| `lastPlayedAt` | `lastViewedAt` + override 0 |
| `sw_lastPlayedAt` | `sw_lastWatched` + override 0 |

You also get two you didn't add: `viewCount` + override 0 is "times played", and `sw_viewedEpisodes` + override 0 is "episodes played".

The override is `percent_complete >= 0`, and Tautulli writes a history row for every playback session, so it does catch a play that was stopped immediately. That is genuinely "clicked play".

Closing this one. :+1:

PR #3281, opened 2026-07-18
fix(media-server): reflect a manually marked Jellyfin season in rule getters (#3274)

### LINKED ISSUE #3274: Jellyfin season manually marked watched is missing from “Users that watched every episode”
# Jellyfin season manually marked watched is missing from “Users that watched every episode”

### Description

I am creating a season-level cleanup rule that should match only when every Jellyfin user who started watching a season has watched every episode.

The affected season was manually marked as watched in Jellyfin using the season-level watched checkmark. Jellyfin shows the season as fully watched for the user.

Maintainerr correctly detects that the user watched at least one episode, but the same user is not returned by **Jellyfin - Users that watched every episode**.

### Environment

- Maintainerr version: 3.17.1
- Media server: Jellyfin
- Rule media type: Seasons

### Steps to reproduce

1. Open a season in Jellyfin.
2. Mark the entire season as watched using the season-level watched checkmark.
3. Confirm that Jellyfin displays the season as fully watched.
4. In Maintainerr, create a season rule comparing:
   - `Jellyfin - Users that watched every episode`
   - `Jellyfin - Users that watched at least one episode`
5. Run **Test Media** for the season.

### Sanitized Test Media output

```yaml
mediaServerId: <redacted>
result: false
sectionResults:
  - id: 0
    result: false
    ruleResults:
      - action: before
        firstValueName: Jellyfin - Last episode added at
        firstValue: <date older than configured threshold>
        secondValueName: custom_days
        secondValue: <calculated threshold date>
        result: true

      - operator: AND
        action: not_contains_all
        firstValueName: Jellyfin - [list] Collections media is present in (titles) (incl. parents)
        firstValue: []
        secondValueName: text
        secondValue: Protected Collection
        result: true

      - operator: AND
        action: contains_all
        firstValueName: Jellyfin - [list] Users that watched every episode
        firstValue: []
        secondValueName: Jellyfin - [list] Users that watched at least one episode
        secondValue:
          - user1
        result: false
```

### Expected behavior

Because Jellyfin shows the entire season as watched for `user1`, I expected Maintainerr to return:

```yaml
firstValue:
  - user1
result: true
```

The list returned by **Users that watched every episode** should contain every user listed under **Users that watched at least one episode** when those users have completed the season.

### Actual behavior

Maintainerr returns:

```yaml
firstValue: []
result: false
```

At the same time, it correctly reports:

```yaml
secondValue:
  - user1
```

Maintainerr therefore recognizes that `user1` has watch activity for the season, but does not recognize the user as having watched every episode after the season was manually marked watched in Jellyfin.

### Additional information

The season was marked watched manually using Jellyfin’s season-level watched checkmark rather than being played through episode by episode.

enoch85:
Fixes #3274.

A Jellyfin season marked watched via the season-level checkmark stayed missing from **Users that watched every episode** for hours - Test Media returned an empty list while **Users that watched at least one episode** was correct. Verified against a real Jellyfin that the mark propagates to every episode instantly, so the staleness was Maintainerr's watch cache, not Jellyfin.

- `resetMetadataCache` scoped the watch-history clear to the item id, so a season/show never invalidated its **descendant episodes'** watch caches (they carry the episode id). It now clears the whole watch-history namespace; the item's favourite/play-count entries and the server-wide aggregate caches (users/libraries/status/collections) are invalidated exactly as before.
- Jellyfin/Emby cache TTLs were written in milliseconds, but `node-cache` takes seconds, so watch history lived ~83h instead of 5min (status ~16.7h, users ~20d). Corrected to the intended seconds.

Before: a freshly marked season returned `every episode = []`, `result: false`. After: `[user]`, `result: true`.

### DIFF
--- a/apps/server/src/modules/api/media-server/emby/emby-adapter.service.ts
+++ b/apps/server/src/modules/api/media-server/emby/emby-adapter.service.ts
@@ -1325,9 +1325,9 @@ export class EmbyAdapterService implements IMediaServerService {
   // ============================================================================
 
   resetMetadataCache(_itemId?: string): void {
-    // The Emby cache only stores library/server-wide aggregates, never
-    // per-item entries, so per-item invalidation collapses to a full flush.
-    // Mirrors the Jellyfin adapter, which keeps the same cache shape.
+    // The Emby cache only stores server-wide aggregates (users/libraries/status/
+    // collections), never per-item watch entries (getWatchHistory hits the API
+    // fresh), so a full flush is the simplest correct reset.
     void _itemId;
     this.cache.flush();
   }
--- a/apps/server/src/modules/api/media-server/emby/emby.constants.ts
+++ b/apps/server/src/modules/api/media-server/emby/emby.constants.ts
@@ -1,14 +1,15 @@
 // Emby cache and batching tuning. Values mirror Jellyfin defaults until tuned
-// against a live Emby server.
+// against a live Emby server. TTLs are in SECONDS - node-cache takes ttl in
+// seconds, not milliseconds (see api/lib/cache.ts and #3274).
 
 export const EMBY_CACHE_TTL = {
-  WATCH_HISTORY: 300000,
-  USER_DATA: 300000,
-  PLAYED_THRESHOLD: 300000,
-  USERS: 1800000,
-  LIBRARIES: 1800000,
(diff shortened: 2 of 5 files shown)


SykApps:
Thanks for fixing this so quickly! <3

Issue #3285, opened 2026-07-18
Metadata ID-resolution runs once per rule condition instead of once per item (redundant CPU + duplicate logs)


nickwolf:
**Describe the bug**

During rule evaluation, the metadata ID-resolution step (`MetadataService.resolveLookupCandidatesFromMediaItemForService`, via the getters' `findLookupCandidatesFromMediaItem`) runs once per rule *condition* per item, rather than once per item. It is not memoized: `ArrLookupCache` (added in #2897) dedupes the arr movie/series lookup that follows, but not the candidate-resolution that precedes it.

So a rule group with N conditions on the same service re-resolves and re-validates every item N times per run. For each redundant call, `validateDirectIds` deep-clones the (cache-hit) provider response (`useClones: true`), re-runs the year-drift validation, and re-emits its INFO log line. The provider *response* is served from cache, so this is not extra network — it is redundant CPU and duplicate logging that scales with (conditions × library size).

It affects the Radarr and Sonarr getters (both call `resolveLookupCandidatesFromMediaItemForService`).

**To Reproduce**

1. Create a rule group with several conditions on the same service (e.g. multiple Radarr `original language` equals conditions).
2. Run it on a large library at `LOG_LEVEL=debug`.
3. Observe the `Accepted direct provider IDs for "..."` log line (and the underlying resolution) repeat for the same item — once per condition, back to back.

**Expected behavior**

The candidate IDs for an item are resolved (and logged) once per item per run, then reused across that item's conditions — the same run-scoped dedup `ArrLookupCache` already provides for the adjacent arr lookup.

**Device (please complete the following information):**
 - OS: Windows 11 (Docker Desktop / WSL2)
 - Version: 3.17.1 (official image `ghcr.io/maintainerr/maintainerr:latest`)
 - Server: Plex

**Debug logs**

The same item resolving twice back-to-back within one run (two conditions), abbreviated:

```
[MetadataService] Accepted direct provider IDs for "<a movie>" (YYYY) with a one-year drift from <provider> id NNNN ...
[MetadataService] Accepted direct provider IDs for "<a movie>" (YYYY) with a one-year drift from <provider> id NNNN ...
```

**Additional context**

[...]

PR #3286, opened 2026-07-18
fix: bound external API response cache key counts (#3284)

### LINKED ISSUE #3284: Rule evaluation heap grows unbounded with library size — external API response caches are not count-bounded
**Describe the bug**

On a large movie library, a single rule-group evaluation grows the server heap roughly linearly with the number of items processed, and can exhaust memory before the run finishes. The shared external-API response caches (`apps/server/src/modules/api/lib/cache.ts`) are bounded only by TTL, not by key count. During a bulk sweep the executor inserts one cached response per library item far faster than the TTL expires them, so a cache grows to library size and holds a full API response per item until the run ends.

Two caches grow one entry per item processed:

- **`tmdb`** — `stdTtl: 21600` (6 h), `persistent: true`, no key limit. The TTL far exceeds a run's duration, so it never prunes during the run and grows without bound. This is the dominant retainer.
- **`plexguid`** — default `stdTtl: 300` (5 min). It self-limits (its TTL prunes mid-run), so it plateaus, but still holds a large working set of full metadata responses.

**To Reproduce**

1. Point Maintainerr at a large movie library (measured on a 10k+ movie library).
2. Enable a rule group that evaluates a per-item property requiring a metadata lookup.
3. Run the rule group with a modest heap (e.g. `NODE_OPTIONS=--max-old-space-size=536`, the default for a ~1 GB container) and watch memory.
4. The `tmdb` cache key count and heap climb linearly with items processed; on a constrained heap the run OOMs (`FATAL ERROR: ... JavaScript heap out of memory`) before completing.

**Expected behavior**

Peak memory during a rule run is bounded independent of library size; a large library does not OOM the process mid-run.

 - OS: Windows 11 (Docker Desktop / WSL2)
 - Version: 3.17.1 (official image `ghcr.io/maintainerr/maintainerr:latest`, Node 26.3.0)
 - Server: Plex

**Debug logs**

Per-page instrumentation (each cache's key count + `process.memoryUsage().heapUsed`), one evaluation on a 10k+ movie library:

```
page=1    tmdb=50     plexguid=51    heapMB=133
page=51   tmdb=2550   plexguid=2601  heapMB=710
page=121  tmdb=6048   plexguid=4742  heapMB=1130
page=191  tmdb=9546   plexguid=4640  heapMB=1356
```

`tmdb` climbs at exactly +50/page and never drops, continuing to the end of the library; `plexguid` oscillates ~4,300–6,300 (its TTL pruning). Heap tracks `tmdb` and peaks over 2 GB by the end of a large run. At `--max-old-space-size=536` the same run instead OOMs around page 57.

**Additional context**

Root cause was isolated with two-snapshot heap diffs (`v8.writeHeapSnapshot` at page 5 vs page 15, which run a full mark-sweep first, so only reachable objects are counted):

- Grouping every retained node under its owning class+property put **+68.7 MB of the +74.4 MB page-5→page-15 delta in `NodeCache.data`** (next owner, `RuleExecutorService.statisticsData`, was +1.4 MB).
- Enumerating the `NodeCache` instances showed the `tmdb` and `plexguid` response caches each grew from 250 keys (page 5) to 750 keys (page 15) — one entry per item — keyed by request URL. Neutering `ArrLookupCache` to a passthrough did not change the delta, ruling it out.

Proposed fix: give the shared caches a key-count ceiling with FIFO eviction, so peak memory is bounded independent of library size. `node-cache`'s own `maxKeys` throws `ECACHEFULL` on overflow rather than evicting (and callers `set()` directly), so the bound is enforced in a thin `NodeCache` subclass that evicts the oldest key before inserting a new one. Verified on the same library at `--max-old-space-size=536`: with the ceiling the caches hold flat at the cap and heap stays flat (a couple hundred MB) through a full sweep that previously OOM'd. Happy to open a PR (one is prepared) if the approach sounds right.

enoch85:
Fixes #3284.

The external API response caches in `apps/server/src/modules/api/lib/cache.ts` were bounded only by TTL, not by key count. During a bulk rule sweep the executor caches one response per library item far faster than the TTL (up to 6h for `tmdb`/`tvdb`) can expire them, so a cache grows to library size and can exhaust a small container's heap before the run finishes.

- `BoundedNodeCache` evicts the oldest key (FIFO) before inserting once the key ceiling is reached. `node-cache`'s own `maxKeys` throws `ECACHEFULL` on overflow instead of evicting, and callers `set()` directly, so the bound is enforced by overriding `set()`.
- The ceiling (`DEFAULT_MAX_KEYS = 1000`) applies to the per-item response caches. The single-aggregate prefetch Maps (`plexwatchhistory`, `seerrrequests`) opt out with `maxKeys: 0`; they hold one large value, not one entry per item, and must never be evicted mid-run.

Peak memory during a rule run is now bounded independent of library size. Verified against a real Plex + Radarr stack: with a low ceiling, a sweep that previously grew the caches one key per movie held flat at the cap, with unchanged rule output. Complements #3285.

### DIFF
--- a/apps/server/src/modules/api/lib/cache.spec.ts
+++ b/apps/server/src/modules/api/lib/cache.spec.ts
@@ -1,0 +1,109 @@
+import cacheManager, { Cache, DEFAULT_MAX_KEYS } from './cache';
+
+describe('Cache key-count bound (#3284)', () => {
+  const fill = (cache: Cache, count: number, prefix = 'k') => {
+    for (let i = 0; i < count; i++) {
+      cache.data.set(`${prefix}${i}`, { i });
+    }
+  };
+
+  describe('bounded cache', () => {
+    it('holds at the ceiling and evicts the oldest key first (FIFO)', () => {
+      const cache = new Cache('t', 't', 'tmdb', { maxKeys: 3 });
+
+      cache.data.set('a', 1);
+      cache.data.set('b', 2);
+      cache.data.set('c', 3);
+      expect(cache.data.keys().sort()).toEqual(['a', 'b', 'c']);
+
+      // Inserting a 4th key evicts the oldest ('a'), not a newer one.
+      cache.data.set('d', 4);
+      expect(cache.data.keys()).toHaveLength(3);
+      expect(cache.data.has('a')).toBe(false);
+      expect(cache.data.keys().sort()).toEqual(['b', 'c', 'd']);
+      // The just-inserted value is retrievable.
+      expect(cache.data.get('d')).toBe(4);
+    });
+
+    it('stays bounded no matter how many distinct keys are inserted', () => {
+      const cache = new Cache('t', 't', 'tmdb', { maxKeys: 10 });
+      fill(cache, 1000);
+      expect(cache.data.keys()).toHaveLength(10);
+      // FIFO: only the most-recent 10 survive.
+      expect(cache.data.has('k999')).toBe(true);
+      expect(cache.data.has('k989')).toBe(false);
+    });
+
+    it('overwriting an existing key reuses its slot (no eviction)', () => {
+      const cache = new Cache('t', 't', 'tmdb', { maxKeys: 2 });
+      cache.data.set('a', 1);
+      cache.data.set('b', 2);
+      cache.data.set('a', 99); // overwrite, not a new key
+
+      expect(cache.data.keys()).toHaveLength(2);
+      expect(cache.data.get('a')).toBe(99);
+      expect(cache.data.has('b')).toBe(true);
+    });
+
+    it('preserves the ttl argument on set()', () => {
+      const cache = new Cache('t', 't', 'tmdb', { maxKeys: 5 });
+      cache.data.set('a', 1, 100);
+      const ttl = cache.data.getTtl('a');
+      expect(ttl).toBeGreaterThan(Date.now());
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


nickwolf:
Ran this branch against a ~15k-movie library to confirm the memory behaviour at scale (the reproduction from #3284).

**Method:** applied this PR's `cache.ts` to a 3.17.1 build and instrumented the rule executor to log each cache's key count and `heapUsed` per page, then ran a full sweep at `--max-old-space-size=536` (the ~512 MB heap a 1 GB container gets — the config that OOM'd before).

**Result** — the default `maxKeys: 1000` holds:
```
page=1    tmdb=50    plexguid=51    heapMB=149
page=25   tmdb=1000  plexguid=1000  heapMB=360   <- caps reached
page=101  tmdb=1000  plexguid=1000  heapMB=384
page=201  tmdb=1000  plexguid=1000  heapMB=355
page=297  tmdb=1000  plexguid=1000  heapMB=426   <- sweep complete
```
Both caches plateau at 1000 by ~page 25 and stay there; heap oscillates ~285–431 MB (peak 431) and never approaches the limit; the full sweep completes with unchanged rule output. On the same build without the bound, the identical run OOM'd around page 57.

The default ceiling is comfortable for a 512 MB-class heap on a large library. Thanks for turning this around so fast.

enoch85:
Thanks for testing! 👍🏼

This will go into development.

Issue #3125, opened 2026-06-18
Collection items flip in/out when Sonarr API lookups fail transiently during rule runs


d3vr:
**Describe the bug**

Shows in my **Leaving Soon** collection keep getting **added and removed** across rule runs, even though the series are tracked in Sonarr and the TVDB IDs match on both Jellyfin and Sonarr.

When it happens, Maintainerr logs warnings like:

```
Could not retrieve show by tvdb ID <id>
```

followed by:

```
None of the resolved external IDs [TVDB:<id>] for '<title>' matched a series in Sonarr. Is the series tracked in Sonarr?
```

A manual `curl` to `GET /api/v3/series?tvdbId=<id>` works for the same ID. Sonarr logs also show those series being refreshed/scanned around the same time, so they clearly exist in Sonarr.

The worst churn tends to coincide with **heavy Sonarr activity** (e.g. the scheduled **Refresh Series** task), but it also happens on runs where I can't see an obvious Sonarr refresh overlap — so load seems to make it worse, not be the sole trigger.

Example from my logs (same day):
- **Morning rule run** — no Sonarr lookup failures → items **added** to Leaving Soon
- **Afternoon rule run** — many Sonarr lookup failures → items **removed** from Leaving Soon

Same shows, same IDs, opposite outcome depending on the run.

───

**To Reproduce**

This is intermittent, so I don't have a 100% reliable repro yet. Best effort:

1. Configure a TV rule group using **Sonarr → Date added** (my Leaving Soon rules do this).
2. Ensure affected shows are in both Jellyfin and Sonarr with matching TVDB IDs.
3. Schedule rule runs at 00:00 / 08:00 / 16:00 (or trigger manually via **Rules → Execute**).
4. Optionally overlap with Sonarr load: **System → Tasks → Refresh Series → Run**, while Maintainerr runs a full-library rule pass.
5. Enable debug logging (`LOG_LEVEL=debug` or UI log settings), restart Maintainerr, then inspect logs.
6. Observe Leaving Soon membership flip between runs, with `Could not retrieve show by tvdb ID …` warnings for IDs that `curl` confirms are in Sonarr.

I haven't completed a controlled repro yet, but my production logs over several days show the pattern clearly.

───

**Expected behavior**

If a show is tracked in Sonarr and the TVDB ID resolves correctly, **Sonarr → Date added** should return a value and the show should stay in (or out of) Leaving Soon consistently across runs.

[...]

PR #3287, opened 2026-07-18
perf: memoize metadata id-resolution per item during rule evaluation (#3285)

### LINKED ISSUE #3285: Metadata ID-resolution runs once per rule condition instead of once per item (redundant CPU + duplicate logs)

enoch85:
Fixes #3285.

During rule evaluation the Radarr/Sonarr getters resolved an item's metadata lookup candidates (`MetadataService.resolveLookupCandidatesFromMediaItemForService`, via `findLookupCandidatesFromMediaItem`) once per rule *condition* instead of once per item. The provider response is served from cache, so this is not extra network, but each redundant call still deep-clones the response, re-runs year-drift validation and re-emits its log line - scaling with (conditions x library size). `ArrLookupCache` (#2897) already dedupes the arr identity lookup that follows, but not the candidate resolution that precedes it.

- Route the candidate resolution through the same run-scoped `ArrLookupCache`, keyed `metadata:${service}:${libItem.id}`. For Sonarr the key is the parent show, so every season/episode of a show collapses onto one entry.
- Empty results are evicted so a transient metadata-provider failure retries on the next condition instead of sticking for the run, preserving the #3125 fail-closed behaviour (mirrors the existing `resolveMovie`/`resolveSeries` eviction).

The memo is the eval-only instance created in the executor and never handed to the collection/action phase, so cleanup still resolves fresh (same lifecycle as #2897).

Verified on a real Plex + Radarr stack: a 2-condition Radarr group over a 5-movie library ran the resolution 10 times before and 5 times after, with identical rule output. Complements #3284.

### DIFF
--- a/apps/server/src/modules/rules/getter/radarr-getter.service.spec.ts
+++ b/apps/server/src/modules/rules/getter/radarr-getter.service.spec.ts
@@ -16,7 +16,11 @@ import { ServarrService } from '../../api/servarr-api/servarr.service';
 import { CollectionMedia } from '../../collections/entities/collection_media.entities';
 import { MaintainerrLogger } from '../../logging/logs.service';
 import { MetadataService } from '../../metadata/metadata.service';
+import { ArrLookupCache } from '../helpers/arr-lookup-cache';
 import { RadarrGetterService } from './radarr-getter.service';
+
+// Let the memo's eviction callback (chained on the resolved promise) run.
+const flushMicrotasks = () => new Promise((resolve) => setImmediate(resolve));
 
 describe('RadarrGetterService', () => {
   let radarrGetterService: RadarrGetterService;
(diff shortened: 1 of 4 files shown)


enoch85:
@nickwolf Please test this PR as well. Will release a test PR soon.

nickwolf:
Tested this on a ~15k-movie library. The rule group I used has 12 conditions, 9 of them Radarr properties (8 x original language, plus tags), so it's a good stress case for the per-condition resolution.

**Method:** applied this PR's memoization to a 3.17.1 build, ran the same page window with and without it, and counted `MetadataService` resolutions per item via the "Accepted direct provider IDs" log.

**Result:** resolutions per item dropped from 10 to 2.

```
before:  800 log lines / 80 distinct titles = 10 per item
after:   178 log lines / 89 distinct titles =  2 per item
```

(the second run covered a few more pages before I stopped it, so the per-item ratio is the comparable number, not the totals)

That breaks down exactly as expected: the 9 Radarr resolutions, one per Radarr condition, collapse to 1. The remaining one is the Seerr getter's own `resolveIdsFromMediaItemForService`, which this PR doesn't cover. So the Radarr candidate resolution went from 9x to 1x per item.

No behaviour change I could detect: every title that resolved in the baseline also resolved with the patch (0 missing), and the distinct WARN/ERROR message types were identical across both runs.

PR #3290, opened 2026-07-19
perf(rules): memoize metadata id-resolution in the Seerr and Sonarr-fallback getters

enoch85:
Follow-up to #3287.

#3287 memoized the Radarr/Sonarr candidate resolution per item. Two callers of the same MetadataService id-resolution still ran once per rule condition; this routes them through the same run-scoped `ArrLookupCache`:

- **Seerr getter** - `resolveIdsFromMediaItemForService` (media item -> tmdb), run before its already-cached request lookup. Key `metadata:seerr:${libItem.id}` (the parent show for season/episode, so seasons/episodes collapse onto one entry); evicts a no-tmdb result so a transient metadata-provider failure retries next condition.
- **Sonarr tv-detail fallback** - `tryMetadataFallback` -> `resolveIdsFromMediaItem`, for shows not tracked in Sonarr using `ended` / `firstAirDate` / `seasons`. Distinct key `metadata:sonarr:details:${libItem.id}` because it resolves under the default (all-provider) policy, not the Sonarr `{tvdb}` policy the candidate memo uses.

The memo is the eval-only cache the executor creates for the rule loop and never hands to the collection/action phase, so cleanup still resolves fresh (same lifecycle as #2897). Also corrects the `ArrLookupCache` docstring: its "Seerr already caches at the API layer" note refers to Seerr's request lookups - the id-resolution that feeds them is a separate layer and wasn't deduped.

No behaviour change: the memo wraps the same resolve call and returns the same value; evict-on-failure preserves the per-condition retry. Unit tests assert once-per-item with a shared cache, re-resolve without one, and evict-and-retry on a transient failure.

### DIFF
--- a/apps/server/src/modules/rules/getter/getter.service.ts
+++ b/apps/server/src/modules/rules/getter/getter.service.ts
@@ -83,7 +83,12 @@ export class ValueGetterService {
         );
       }
       case Application.SEERR: {
-        return await this.seerrGetter.get(val2, libItem, dataType);
+        return await this.seerrGetter.get(
+          val2,
+          libItem,
+          dataType,
+          arrLookupCache,
+        );
       }
       case Application.TAUTULLI: {
         return await this.tautulliGetter.get(
--- a/apps/server/src/modules/rules/getter/seerr-getter.service.spec.ts
+++ b/apps/server/src/modules/rules/getter/seerr-getter.service.spec.ts
@@ -10,6 +10,10 @@ import {
 } from '../../api/seerr-api/seerr-api.service';
 import { MetadataService } from '../../metadata/metadata.service';
 import { SeerrGetterService } from './seerr-getter.service';
+import { ArrLookupCache } from '../helpers/arr-lookup-cache';
+
+// Let the memo's eviction callback (chained on the resolved promise) run.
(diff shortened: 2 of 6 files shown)


enoch85:
@nickwolf Please test this as well. Thanks! :pray:

enoch85:
Note on Test-Media-vs-real-run parity (the #3152 invariant): verified this holds.

`ArrLookupCache` is only wired on full runs (`executeForRuleGroups`), not on Test Media (`testRuleGroupWithData` passes no cache), so a natural question is whether Test Media diverges from a run. It doesn't: this cache is a per-run memoization (created fresh each run, never persisted, evicts failures), so the memoized value equals what the un-memoized Test Media path computes for the same item in the same run. Unlike #3152's persistent `seerrrequests` index, there's no cross-run state to go stale.

Checked on a real Jellyfin + Seerr stack (Seerr rule over the movie library): Test Media predicted the exact match set the memoized full run collected (`10378, 133701, 358332, 45745`), and the before/after full runs matched the identical set - the memo changed only the resolution count (13 -> 5, i.e. 3x -> 1x per item), never the values.

enoch85:
Tested and works!

Issue #3065, opened 2026-06-08
Emby movie rule actions fail: "Couldn't resolve any supported external IDs" — resolution reads the movie's ID-less parent


jaerchom:
**Describe the bug**
On an Emby media server, **every** movie rule action (Radarr) fails with `Couldn't resolve any supported external IDs for movie with media server ID <id>`, even though the Emby item clearly has TMDB and IMDb IDs. The rule *matching* works (the collection fills with the expected movies), but the **action** can never resolve a TMDB id, so nothing can be unmonitored/deleted/etc.

Root cause appears to be in `apps/server/src/modules/metadata/metadata.service.ts` → `getHierarchyResolutionItem()`:

```ts
const hierarchyTargetId = item.grandparentId ?? item.parentId;
if (!hierarchyTargetId) return item;
const hierarchyItem = await mediaServer.getMetadata(hierarchyTargetId);
return hierarchyItem;   // IDs are resolved from the PARENT, not the item
```

This is correct for episodes (resolve from the series), but for an **Emby movie** `parentId` points at an ID-less container, so the movie's own provider IDs are discarded → 0 lookup candidates → `RadarrActionHandler.handleAction` hits the `lookupCandidates.length === 0` branch and aborts. It fails silently (no `MetadataService` debug/warn) even with `LOG_LEVEL=debug`.

**To Reproduce**
1. Connect an Emby server as the media server.
2. Create a movie rule that builds a collection (Radarr action; mine is "Unmonitor", but any servarr action reproduces it).
3. Open the collection, pick a movie, and click "Trigger Rule Action".
4. See the error in the logs: `Couldn't resolve any supported external IDs ... Failed to trigger the collection action for this item.`

**Expected behavior**
Maintainerr should resolve the movie's TMDB id (which is present on the Emby item) and perform the Radarr action against the matching movie.

**Screenshots**
Emby's "External Ids" panel for the affected movie shows valid IMDb (`tt0109040`) and TheMovieDb (`3049`) IDs, so the data exists on the item. (Screenshot can be attached.)

**Device (please complete the following information):**
 - OS: Unraid 7.1.4 (official image `ghcr.io/maintainerr/maintainerr:latest`, deployed via Docker Compose). Note: not host-specific — this is deterministic application logic and reproduces independent of host.
 - Version: 3.14.0
 - Server: Emby 4.10.0.14

**Debug logs**

[...]

PR #3306, opened 2026-07-22
feat: add Sportarr as a native application connection

Sportarr:
Adds Sportarr as its own native application in Maintainerr, alongside Sonarr and Radarr. It talks to Sportarr's own API directly (not the Sonarr-v3 compatibility shim), lives entirely in its own module so none of the Sonarr/Radarr logic is touched, and is fully opt-in.

Resolution is exact-id: Sportarr stamps a numeric alias in the tvdb namespace on its Plex items, which the existing Plex mapper already pulls into `providerIds.tvdb`, and the connection reverses that back to the league id and looks it up. No title or year fuzzing.

Includes the settings side (server CRUD, entity, migration), a native API client, the rule getter + properties, and the action handler across show / season / episode scope (delete, unmonitor, change quality profile), with coverage-aware download-client cleanup. Full parity with the Sonarr connection.

See the comment below for the full design write-up, rationale, AI-assisted-development note, and how to test.


enoch85:
Thanks for the PR! 

While I understand why, I still think this is risky. I haven't had a proper look yet, but reading the description and looking quickly through the code makes me wonder if this is strict enough? You mention fuzzy matching, but isn't this solution very close to that?

Elaborate further on that please.

enoch85:
Also, we already do several tricks with the metadata to get it right. Year matching for example. 

Would a better solution to implement a strict Sportarr metadata fetch which only is enabled if you enable Sportarr?

Sportarr:
> Also, we already do several tricks with the metadata to get it right. Year matching for example.
> 
> Would a better solution to implement a strict Sportarr metadata fetch which only is enabled if you enable Sportarr?

Want to make sure I read you right first. Do you mean Sportarr as its own app connection in Maintainerr, sitting next to Sonarr/Radarr and talking to my API directly instead of going through the Sonarr endpoints? Or something lighter, like keeping the Sonarr-style connection but with a Sportarr flag that switches it to strict id resolution? I'm happy to build either, but if you mean the full connection, I'm all in! That's the cleanest version of this by far imo.

Either way, most of the pieces already exist. Plex items from Sportarr already carry real external ids. If you pull the XML for one of the shows you'll see two GUID entries, sportarr://lg-000278 and tvdb://900000278. The tvdb one is just the sportarr id plus a 900 million offset so it stays way out of range of any real TVDB id, and my Sonarr-compatible API returns that exact number as tvdbId. So strict resolution is already possible today with zero guessing: read the GUID, look up by id, matched. The id scheme is written up here: https://github.com/Sportarr/Sportarr/blob/dev/docs/EXTERNAL_IDS.md

For the full-connection version, Sportarr has its own complete API too, documented here: https://github.com/Sportarr/Sportarr/blob/main/docs/api.md with the longer version at https://sportarr.net/docs/api. Everything a Maintainerr integration would need is in there: lookups by id, monitor/unmonitor, delete with files, grab history for the download client cleanup. And whatever endpoints Maintainerr ends up using, I'll keep them stable and versioned on my side.

I'll gladly drop the title fallback for any of this if you want. This was the workaround for direct connection anyway. Let me know which shape you had in mind, and whether I should rework this PR or close it and open a fresh one.

Also, FYSA...the Plex GUID part is already live (existing shows pick it up after a metadata refresh), but the tvdbId reporting on my API is in dev builds now. Next stable release is likely this weekend.

enoch85:
> Do you mean Sportarr as its own app connection in Maintainerr

Yes, that's what I mean. That would be easier for me to merge as it wouldn't affect Sonarr/Radarr logic.

> If you pull the XML for one of the shows you'll see two GUID entries, sportarr://lg-000278 and tvdb://900000278.

Hmm, but why isn't that enough? Why do you need the fallback - in the current implementation? EDIT: Aah your ID contains letters, not only numbers.

Maybe this should go in as a Metadata Provider instead? 

What I want: as few moving parts as possible, as tight and trimmed as possible. We already have super good Metadata logic (IMHO) so maybe just adding support for your tag would be the smallest fix?

enoch85:
We also have this: https://features.maintainerr.info/posts/152/support-for-sportarr

Sportarr:
Yeah, let's do the app connection. As you said, that version is easiest to merge since it stays out of the Sonarr/Radarr logic, and it's what I want long-term too, so I'll rework this PR into that.

On the metadata provider idea, I think a provider only gets you halfway. It'd solve the identity side (reading sportarr:// to know which show an item is), but the actions still have to go somewhere, and right now that somewhere is the Sonarr-shaped API. A real connection handles both the lookup and the delete/unmonitor/history calls against my own endpoints, and the nice part is it all lives in its own module, so your existing metadata and arr logic doesn't get touched.

And yeah, the id has letters (lg-000278), but the number is just the digits plus a 900,000,000 offset, which is what that second tvdb://900000278 GUID already is. So the tag support folds in.

The title fallback drops out entirely in this design; it was only ever the workaround for not having a direct connection. (Truly didn't expect the willingness to add it in, but I am quite grateful)

Been working on a rebuild all day today, so I'll follow up here with a full rundown of what actually changed once it's ready for a proper look. Also saw your message about the feature board post; proud to know there's demand for it beyond just me.

Sportarr:
Reworked this into the native Sportarr connection, and it's now full parity with the Sonarr side, including download-client cleanup.

### Description & Design

**What this does.** Adds Sportarr as its own application in Maintainerr, next to Sonarr and Radarr. It talks to Sportarr's own API directly instead of the Sonarr-v3 shim, and everything lives in its own module, so none of your Sonarr/Radarr logic is touched.

**Why this shape.** You said an app connection would be easiest to merge because it stays out of the existing arr logic, and I agree, so that's how it's built. It's also opt-in: for a "Shows" library (which in Plex can be either a TV library or a sports library) there's a single "Managed by: Sonarr / Sportarr" selector that defaults to Sonarr and only appears once a Sportarr server is configured. If you don't run Sportarr, the collection form is exactly what it is today. Pick Sportarr and you get the one Sportarr server + action block in place, never a second stray picker.

**How resolution works.** Exact id, never fuzzy. Sportarr stamps a numeric alias in the tvdb namespace on its Plex items (tvdb://900000278 for league lg-000278), which your Plex mapper already pulls into providerIds.tvdb. The connection reverses that back to the league id and looks it up by id. No title matching, no year inference. The offset is a frozen part of my published id contract, and if the rule executor hands over a lightweight item without providerIds, it re-fetches the show's metadata before giving up.

**What's included.** The settings side (server CRUD, entity, migration), a native API client, the rule getter and props, and the action handler across show/season/episode scope: delete, unmonitor, change quality profile. Delete also cleans the download client the same way Sonarr does, and it's coverage-aware, a season or weekend pack that still backs an event you're keeping is left in the client. That last part needed one small native endpoint on the Sportarr side to expose per-league download ids, which I've shipped.

**Trade-offs.** The action set is intentionally focused on what makes sense for sports rather than mirroring every Sonarr action verb. Grabs with no captured download-client id can't be auto-removed, but that's identical to Sonarr's own history behavior, not a gap here.

### Related issue

No open issue covers this directly; it's the follow-through on our discussion in this PR.

### AI-Assisted Development

I used an AI coding assistant to help draft the implementation and the tests. The design came from reading how your Sonarr getter and action handler are structured and mirroring that exactly, so the Sportarr equivalents read as siblings, down to the null/undefined fail-closed lookup contract, the run-scoped ArrLookupCache memoization, uncached reads on the action paths, and the coverage rule the download cleanup uses.

I reviewed and tested everything, and I ran the type-checker and lint as well as the suite before pushing. Beyond the unit tests I ran the whole thing end to end against a real Plex library and a live Sportarr instance (details below). Happy to walk through any part of it.

### How to test

1. Add a Sportarr server under Settings, Sportarr.
2. Create a collection over a Shows library. You'll see a "Managed by" selector (defaults to Sonarr); switch it to Sportarr and pick your Sportarr server.
3. Build a rule from the Sportarr properties and run it. Items resolve by exact id against Sportarr, and a delete cleans the download client.
4. Unit tests: `yarn workspace @maintainerr/server test sportarr`.

### Additional context

Verified end to end on a real setup: a rule over a sports library resolved and collected the right leagues by exact id, shows not tracked in Sportarr were skipped with no false matches, and the download-cleanup path was proven against real grab history. Full server suite passing, plus type-check and lint clean.

One note for you: the "Managed by" selector is my answer to the fact that Plex reports both TV and sports libraries as the same type, so a Shows collection has to let the user say which arr owns it. Defaulting to Sonarr keeps the existing flow unchanged. Happy to adjust the UX if you'd rather it work differently, and to stack this into smaller PRs if that's easier to review.

enoch85:
@Sportarr Can you please resolve the conflict?

Sportarr:
> @Sportarr Can you please resolve the conflict?

Rebased onto the latest dev, conflict resolved. It was just the migration test file, we both added a migration and updated the same assertions. Your rule-removal migration runs first in the chain, mine after, and the tests now cover both. Everything should be green.

Sportarr:
Spoke too soon on the green part, sorry about that. Fixed the two red checks from before: a formatting nit in the new settings page and the settings tab test that didn't know about the new tab yet. That one bit me because I'd only been running the server suite locally, so the frontend suite is part of my checks now too.

After that, I did a full review pass over my own PR against how your Sonarr paths behave, and found a few things worth fixing before anyone spends review time on it, so the latest push includes those as well.

The big one was failure semantics. A failed Sportarr fetch could read as an empty events list, meaning "0 downloaded events", which is exactly what removal rules match on. During an outage that would have failed open. It now fails closed everywhere the same way your Sonarr getter does, and the delete paths unmonitor first and check every write before touching files, same reasoning as your #3228 fix.

Also tightened the opt-in story. The app is filtered out of the rule constants when no Sportarr server is configured, and the rule builder only offers Sportarr properties when the collection actually has a Sportarr server selected, so users without Sportarr never see it anywhere. The server also rejects a payload that binds a collection to both Sonarr and Sportarr.

Change quality profile is now fully reachable from the UI (profiles endpoint plus selector), and it's refused for season and episode collections since the profile lives on the league. Season matching keys on seasonNumber with the label derived from events, since sports season labels can be cross-year strings like 2025-2026.

All of it is covered by tests on both sides, including wiring Sportarr into your per-property getter coverage matrix. Full suites green, still a single commit.

enoch85:
Great PR! High quality and following the repo rules. "27 agents, 0 errors, ~35 min, 1.33M subagent tokens." didn't find anything substantial. Nice work! 

Though there are a few things:

1. One functional bug to fix first. The whole-league delete/unmonitor/quality action silently no-ops on Jellyfin/Emby. In sportarr-action-handler.ts the league-id fallback walks the item up to grandparentId ?? parentId unconditionally, but on JF/Emby a show's parentId is the library folder, so it re-fetches the folder (which has no tvdb alias) and resolves nothing. It only works on Plex because a show has no parent there. Same trap as #3065 - please switch on item.type and only walk up for season/episode scope, reading providerIds.tvdb directly for a show. It fails closed so there's no data loss, but it breaks the primary action on 2 of the 3 servers.

A few small ones worth folding in while you're in there, none blocking: resolveEvent requires both index and parentIndex, so a date-only event can't be actioned; the no-providerIds getter fallback skips ArrLookupCache (the #3285 redundancy); and getLeagueByExternalId re-pulls the whole /leagues list once per league instead of memoizing it.

2. The bigger one - the API this leans on isn't in a stable Sportarr release yet. I went through your source to confirm: the tvdb-alias contract (EXTERNAL_IDS.md) and the endpoints the action side calls (/events/{id}/files, /leagues/{id}/seasons/{season}/toggle, /leagues/{id}/download-history) are on your dev branch, not the current stable. The offset (900,000,000) and the reversal match NumericIdAlias.cs exactly, and the native /api/* auth accepts ?apikey=, so the wiring is right - but the action path won't work for anyone until that ships in a tagged release. The read/identity side only needs /api/leagues + /events, which are already stable.

Given that, plus the size (~4k lines, roughly half of it the action surface), I'd like to land this in two steps:

v1 now: identity + rules read only (resolve leagues, build and run Sportarr rules). Depends only on stable endpoints, about a third of the diff, and it's exactly the tight, minimal version I was after.
v2: delete/unmonitor/quality + download cleanup, once those endpoints are in a tagged Sportarr release so we can pin a minimum version and test against it.

Everything else is genuinely clean: fail-closed handling on every getter branch and every write, a faithful mirror of the Sonarr paths, a safe migration (I tested a 3.19 -> 3.20 upgrade with real data, no loss and a clean rollback), good reuse of the shared settings modal / getter / cache, and the opt-in gating is exactly right.

Let me know which way you want to go and I'll help get it over the line.

enoch85 (changes requested):
See comment above.

enoch85:
Thanks!

Another standard in this repo: Keep enums and variables in one single file. So everything that's Sportarr specific (like version), should go into that file. Look for other example files in this repo. :+1:

Sportarr:
> Thanks!
> 
> Another standard in this repo: Keep enums and variables in one single file. So everything that's Sportarr specific (like version), should go into that file. Look for other example files in this repo. 👍

Since your last review I cut Sportarr 4.0.1022, and it ships everything this connection calls. Most of the endpoints were already in the previous stable release, the only missing pieces were the download history endpoint and the Jellyfin/Emby id emission, and both are in the tagged release now. That removes the reason for the two step landing, so this can go in as one piece. The connection also now checks the Sportarr version when you test the server in settings and refuses anything older than 4.0.1022 with a clear message to update.

All four review items are fixed and pushed on top of your merges. Shows resolve from their own provider ids now and only seasons and episodes walk up to the parent, with a regression test for the library-folder case you caught. Date-only events resolve by air date with ambiguous days failing closed, the metadata fallback and the /leagues list are memoized per run, and the league resolver takes a pre-fetched list.

And since the folder-parent bug proved Plex-only testing wasn't enough, I stood up throwaway Jellyfin and Emby servers and ran the whole thing end to end on both. The plugin stamps the id alias, a rule resolves and collects the right episode, and the delete unmonitors first then removes exactly the one file. Results matched the Plex run on both servers. One side note from that testing, an Emby setup needs the user id saved with the connection or metadata reads fail. When that happens resolution fails closed with a warning instead of acting, which is the designed behavior.

Also moved the Sportarr specific constants (the id alias offset and range, the minimum version) into a single file in contracts per the repo standard, and the UI imports them from there too instead of repeating the numbers.

enoch85:
Nit: Can you fix the naming of the consts file?

Example: /workspace/apps/server/src/modules/api/seerr-api/seerr-api.constants.ts

enoch85 (approved):
Let's do this!

Sportarr:
Merged plus follow-ups?! Thanks for the smooth review! The upcoming events property is a great addition too. Looking forward to the next release. Excited to tell my users.

enoch85:
> Merged plus follow-ups?! Thanks for the smooth review! The upcoming events property is a great addition too. Looking forward to the next release. Excited to tell my users.

Well, when such a great PR arrives, it's hard to say no. :smile:  Really good job, nice working with you!

PR #3335, opened 2026-07-27
feat(sportarr): follow-up nits + "has upcoming events" rule property

enoch85:
Follow-ups from the review of #3306.

- **fix:** `seasonIndex` reads the season item's `index` directly instead of inferring scope from `grandparentId` presence (#3065); the getter's outer-catch keeps `warn` plain with the throwable on `logger.debug`; the version parser strips the `v` prefix without a regex.
- **feat:** adds a `hasFutureEvents` ("Has upcoming events") BOOL rule property at show/season scope, computed from the events already fetched for the count properties. Lets retention rules keep in-progress leagues and only act on ended ones. Fails closed (undefined) when the events fetch fails.
- **docs:** adds Sportarr to the supported-services badges in the README.

Note: `hasFutureEvents` currently counts any scheduled future event, monitored or not. Sportarr's own "continuing" status keys on monitored future events - happy to switch to that if preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ spec:
 
 # Features
 
-- Build rules from properties across Plex, Jellyfin, Emby, Radarr, Sonarr, Seerr, Tautulli and Streamystats, combined with AND/OR logic.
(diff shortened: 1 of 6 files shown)


enoch85:
cc @Sportarr

PR #3345, opened 2026-07-28
perf(media-server): resolve show watch state in one sweep per user (#3337)

### LINKED ISSUE #3337: Jellyfin: Shows rule runs scale with episodes × users, no bulk prefetch
**Describe the bug**

Rule evaluation for Shows rule groups on Jellyfin scales with `episodes × users`, which makes runs impractically long on a medium to large sized libraries.

Every show-level `sw_*` property (`sw_lastWatched`, `sw_watchers`, `sw_viewedEpisodes`, `sw_allEpisodesSeenBy`) walks all seasons and episodes of a series. For each episode, `JellyfinAdapterService.getWatchHistory()` calls `getAllUserItemData()`, which issues **one `/Items` request per Jellyfin user**. For a library of roughly 5000 episodes and ~20 users, that is on the order of 100k HTTP requests for a single full sweep.

Two things appear to make this worse than it needs to be:

1. `JellyfinAdapterService.prefetchWatchHistory()` throws `Bulk watch-history prefetch is not supported on Jellyfin (per-user history)` and is gated off via `supportsFeature(MediaServerFeature.CENTRAL_WATCH_HISTORY)`. The Plex adapter has `fetchWatchHistoryMap()` and does get a bulk prefetch. Jellyfin genuinely has no *central* history endpoint, but it does support bulk queries *per user* (see Expected behavior).

2. The shared `jellyfin` cache is registered with the defaults (`stdTtl: 300`, `maxKeys: 1200`), while a sweep of a library this size needs one watch-history key per episode. With FIFO eviction the working set never fits, so every section that touches watch data re-fetches nearly all of it. By comparison, `plexwatchhistory` is deliberately exempted with `maxKeys: 0` so its prefetched map is "never evicted mid-run" — Jellyfin has no equivalent.

**To Reproduce**

1. Connect Maintainerr to a Jellyfin server with a few thousand episodes and a double-digit user count
2. Create a rule group with media type *Shows*
3. Add any show-level watch property, e.g. `Jellyfin - Newest episode view date` (`sw_lastWatched`)
4. Run the rule group, or use *Test media* on a single series
5. Observe the runtime and the request volume hitting Jellyfin — roughly `episodes × users` per property sweep

**Expected behavior**

Watch state for a Shows rule group should be retrievable in bulk rather than one request per (episode, user) pair.

Jellyfin's `/Items` endpoint already supports this: it accepts `ids` as a comma-delimited list and additionally `isPlayed`, `recursive`, `includeItemTypes`, `startIndex`/`limit`, `userId` and `enableUserData`. A single paginated query per user — "all played episodes in this library" — would return the same information the current code assembles item by item. For a library of this size that is roughly 20 users × a handful of pages, i.e. a few hundred requests instead of ~100k.

That would let Jellyfin participate in the existing prefetch flow, following the pattern `PlexApiService.fetchWatchHistoryMap()` already establishes, including its strictness about completeness: Plex discards the whole map if the collected count does not match the reported total, and Jellyfin's `/Items` returns `TotalRecordCount` for the same check. Given the destructive nature of the tool, a partially populated map must never be treated as authoritative — an item missing from the map would read as "never watched" and become a deletion candidate.

If a full prefetch is out of scope, a smaller improvement would be giving Jellyfin watch-history its own cache with a raised key ceiling, so at least repeated sections within one run can hit it.

**Screenshots**

n/a

 - OS: Docker on Debian Trixie  
 - Version: 3.19.0
 - Server: Jellyfin 10.11.11

**Debug logs**

n/a

**Additional context**

Things already tried, which help but do not remove the rollups:
- Ordering each section so cheap properties run first. `Jellyfin.addDate` comes from already-loaded metadata and costs no call, and `Sonarr.tags` / `Seerr.isRequested` are memoized, so the expensive `sw_*` rules only see the survivors. This gave a noticeable improvement, since the first rule of a section is evaluated against the whole library and every following AND rule only against the remaining set.
- Dropping `sw_allEpisodesSeenBy` from the ruleset. Worth flagging separately: it is the only getter that fires `Promise.all` across *every* episode of a series simultaneously, so a 200-episode show issues several thousand concurrent requests in one burst. The other getters iterate sequentially.

Happy to test a patch and to provide timings before/after.

enoch85:
Closes #3337.

Every show-level watch property walked seasons then episodes and called `getWatchHistory` per episode, and each of those issues one `/Items` request **per Jellyfin/Emby user**. A Shows rule group therefore cost `O(episodes x users)` requests.

`getDescendantEpisodeWatchers` already had the cheap shape - one recursive `/Items` per user, scoped to the show. This generalises it into `getDescendantEpisodeWatchHistory`, which returns watch records keyed by episode id, and serves `sw_lastWatched`, `sw_viewedEpisodes`, `sw_amountOfViews`, `sw_allEpisodesSeenBy` and show/season `lastViewedAt` from it. Cost becomes `O(users)` per show.

- **No semantic change.** The sweep hits the same endpoint with the same `enableUserData` payload the per-item path reads; I verified the `UserData` objects are byte-identical between the bulk and single-id forms on Jellyfin 10.11.11 (unplayed, played, and mid-progress).
- **No `isPlayed` filter.** It tests the `Played` flag alone, so it would silently drop episodes watched only by crossing the `PlayedPercentage` threshold (#2466) - a false "never watched" on a tool that deletes media.
- **All-or-nothing.** A failed user sweep or a short page throws instead of caching a partial map, so a transient failure skips the item rather than reading as never watched (#2744, #3318). `getDescendantEpisodeWatchers` no longer swallows errors into an empty watcher list for the same reason.
- **Sweeps get their own small uncloned cache.** They are much larger than a normal cached response (cloning one measured ~48ms per read) and every consumer only reads them; the key ceiling is sized to the executor's 50-item chunk so #3284's memory bound still holds.
- Emby mirrors Jellyfin throughout - it has the identical per-user model and had the identical defect.

Measured against a real Jellyfin (Blender Open Movies, 5 episodes, 5 users), before vs after: all six property values identical, **154 requests -> 46**. Per property the watch cost drops from `episodes x users` to `users`, so the ratio grows with episode count.

Not addressed here: movie and episode-level rule groups still pay `O(users)` per item, which is inherent to Jellyfin's per-user watch model without a library-wide prefetch.

### DIFF
--- a/apps/server/src/modules/api/lib/cache.spec.ts
+++ b/apps/server/src/modules/api/lib/cache.spec.ts
@@ -90,6 +90,24 @@ describe('Cache key-count bound (#3284)', () => {
         cache.data.flushAll();
       }
     });
+
+    // Each entry covers a whole show's episodes across every user, so these
+    // hold far fewer keys than the shared per-item caches and are never cloned.
+    it('bounds the watch-sweep caches well below DEFAULT_MAX_KEYS and skips cloning', () => {
+      for (const id of ['jellyfinwatchsweep', 'embywatchsweep'] as const) {
+        const cache = cacheManager.getCache(id);
+        cache.data.flushAll();
+
+        const value = { 'ep-1': [] };
+        cache.data.set('sweep', value);
+        expect(cache.data.get('sweep')).toBe(value);
+
+        fill(cache, DEFAULT_MAX_KEYS);
+        expect(cache.data.keys().length).toBeLessThan(DEFAULT_MAX_KEYS);
+        expect(cache.persistent).toBe(false);
+        cache.data.flushAll();
+      }
+    });
   });
 
   describe('existing behaviour is preserved', () => {
--- a/apps/server/src/modules/api/lib/cache.ts
+++ b/apps/server/src/modules/api/lib/cache.ts
@@ -13,7 +13,9 @@ type AvailableCacheIds =
   | 'streamystats'
   | 'github'
   | 'jellyfin'
-  | 'emby';
+  | 'jellyfinwatchsweep'
+  | 'emby'
+  | 'embywatchsweep';
 
 type CacheType = AvailableCacheIds | 'radarr' | 'sonarr' | 'sportarr';
 
@@ -29,6 +31,10 @@ const DEFAULT_CHECK_PERIOD = 120; // 2 min
 // window. It stays far above any paginated/working-set flow, so normal use never
 // evicts.
 export const DEFAULT_MAX_KEYS = 1200;
+// Key ceiling for the Jellyfin/Emby per-show watch sweeps. Sized to the rule
+// executor's 50-item chunk (rule evaluation is rule-major, so only the current
+// chunk's sweeps need to survive from one rule to the next) with headroom.
+const WATCH_SWEEP_MAX_KEYS = 100;
 
 type CacheOptions = {
   stdTtl?: number;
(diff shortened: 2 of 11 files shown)


enoch85:
@bin101 Please test the PR release (when it arrives).

bin101:
@enoch85 works and pushed the runtime from 34 minutes down to 6 minutes, ty and nice solution.

PR #3352, opened 2026-07-29
fix(plex): use native watched state for rules

noahphex:
### Description & Design

This change fixes two Plex-related rule accuracy issues.

First, Plex watched-state rules could treat an item as watched when Plex currently reports it as unwatched. This happened because Maintainerr used Plex watch-history rows as the primary watched signal. A historical play row can remain even after an item is manually marked unwatched, so cleanup rules could match items that were not currently watched in Plex.

This PR changes Plex watch-state evaluation to use Plex’s native `viewCount` from the media item metadata as the source of truth for `viewCount` and `isWatched`. That matches Plex’s current watched state and avoids stale history rows causing false positives.

Second, Plex last-view-date rules now prefer native Plex `lastViewedAt` values from the library item or metadata before falling back to watch history. This supports items that Plex reports as viewed through current metadata even when a separate watch-history query does not return usable rows.

This PR also improves provider ID extraction for older Plex metadata by parsing legacy top-level GUIDs such as `com.plexapp.agents.imdb://...`, `com.plexapp.agents.themoviedb://...`, and `com.plexapp.agents.thetvdb://...`. This allows Maintainerr to resolve IMDb/TMDB/TVDB IDs for older library items where Plex does not provide modern `Guid[]` children.

Trade-offs:
- For Plex watched-state rules, this intentionally favors Plex’s current watched state over historical play events. This is more accurate for cleanup rules that are asking “is this watched now?”
- Historical watch records are still used as a fallback for `lastViewedAt` when native Plex fields are unavailable.
- Legacy GUID parsing ignores unknown providers and continues to ignore `plex://` GUIDs.

### Related issue

No linked issue. This came from testing against a real Plex library where cleanup rules selected movies that were currently unwatched in Plex because they had old watch-history rows.

### AI-Assisted Development

AI assistance was used to help inspect the affected Maintainerr code paths, translate a locally tested Docker patch into source-level TypeScript changes, and draft/update focused unit tests.

I personally reviewed the resulting changes, checked that they match the observed Plex behavior, and validated them locally. The implementation follows the existing project structure by keeping the behavior in the Plex adapter, Plex mapper, and Plex getter service, with tests added to the existing spec files for those modules.

Local validation completed:
- `yarn workspace @maintainerr/server test plex.mapper.spec.ts plex-adapter.service.spec.ts plex-getter.service.spec.ts`
- `yarn format:changed`
- `yarn lint`
- `yarn check-types`

### Checklist

### How to test

1. Run the focused Plex server tests:
   ```bash
   yarn workspace @maintainerr/server test plex.mapper.spec.ts plex-adapter.service.spec.ts plex-getter.service.spec.ts

### DIFF
--- a/apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
+++ b/apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
@@ -235,17 +235,19 @@ export class PlexAdapterService implements IMediaServerService {
     itemId: string,
     nativeViewCount?: number,
   ): Promise<MediaWatchState> {
-    const metadata =
-      nativeViewCount === undefined
-        ? await this.plexApi.getMetadata(itemId)
-        : undefined;
+    // Read live: something watched moments ago must not be judged from a
+    // stale snapshot, and a failed read has to throw rather than pass for a
+    // confirmed never-watched.
+    const history = await this.plexApi.getWatchHistory(itemId, false);
 
-    const nativeCount = Number(nativeViewCount ?? metadata?.viewCount ?? 0);
+    // Plex writes no history row when an item is marked watched without a
+    // play event (a "mark as played", a Trakt scrobble), so the item's own
+    // count is the only record of those views. History stays the floor: it
+    // covers every account on the server, while the item's count only covers
+    // the account whose token Maintainerr holds.
+    const viewCount = Math.max(history.length, nativeViewCount ?? 0);
 
-    return {
-      viewCount: nativeCount > 0 ? nativeCount : 0,
-      isWatched: nativeCount > 0,
-    };
+    return { viewCount, isWatched: viewCount > 0 };
   }
 
   async getItemSeenBy(itemId: string): Promise<string[]> {
--- a/apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
+++ b/apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
@@ -234,37 +234,17 @@ export class PlexAdapterService implements IMediaServerService {
   async getWatchState(
     itemId: string,
     nativeViewCount?: number,
-    itemTitle?: string,
-    itemType?: PlexLibraryItem['type'],
   ): Promise<MediaWatchState> {
-    const history = await this.plexApi.getWatchHistory(itemId, false, itemType);
+    const metadata =
+      nativeViewCount === undefined
+        ? await this.plexApi.getMetadata(itemId)
+        : undefined;
 
-    if (history.length > 0) {
-      return {
-        viewCount: history.length,
-        isWatched: true,
-      };
-    }
-
-    // When watch history is empty (purged or item was marked watched without
-    // a play event), fall back to the native Plex viewCount for the boolean
-    // only.  This value is per-user (admin token) so we do not use it for
-    // the numeric viewCount to avoid misrepresenting server-wide counts.
-    const watchedByNative =
-      nativeViewCount !== undefined && nativeViewCount > 0;
-
-    if (watchedByNative) {
-      this.logger.log(
-        `Media '${itemTitle ?? 'unknown'}' (ratingKey=${itemId}) is marked watched in Plex ` +
-          `but has no watch history. viewCount will be 0. This can happen when ` +
-          `history is purged or the item was marked watched without a play event ` +
-          `(e.g. Trakt/API scrobble).`,
-      );
-    }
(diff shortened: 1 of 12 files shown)


apps/server/src/modules/api/media-server/plex/plex-adapter.service.ts
```diff
@@ -232,39 +232,19 @@ export class PlexAdapterService implements IMediaServerService {
...
-    }
-
-    // When watch history is empty (purged or item was marked watched without
-    // a play event), fall back to the native Plex viewCount for the boolean
-    // only.  This value is per-user (admin token) so we do not use it for
-    // the numeric viewCount to avoid misrepresenting server-wide counts.
-    const watchedByNative =
-      nativeViewCount !== undefined && nativeViewCount > 0;
+    void _itemTitle;
+    void _itemType;
```
enoch85: Why do you void these, and why do you use underscore for enums?

enoch85:
Related: https://features.maintainerr.info/posts/219/mark-as-watched-rule-parameter

noahphex:
Thanks, that feature request does look related.

I think this PR is the narrower fix for the existing Plex watched-state behavior: rules using watched state/view count should follow Plex’s current native watched state, not stale historical play rows. The feature request seems like it could build on this by exposing more explicit rule semantics, for example current watched state vs historical playback activity, but I kept this PR scoped to correcting the current false-positive behavior.

I also cleaned up the unused parameter issue. `itemTitle` and `itemType` are no longer passed through `getWatchState`, and `itemId` is now used to fetch Plex metadata when `nativeViewCount` is not already available. I reran format, lint, type checks, and the focused Plex tests.

enoch85:
Thanks for this, the underlying problem is real. I have taken it over and reworked it. Here is what changed and why.

**Kept, and now covered by tests.** Marking something played in Plex, or a scrobble from an external tracker, writes no watch-history row at all. I confirmed that against a real server. So an item watched that way used to report "never viewed", and its last view date came back empty. Both now count, which is the good half of this PR.

**Changed: history is the floor again, not something the item's own count replaces.** Plex's per-item view count only reflects the account whose token Maintainerr holds, while watch history covers everyone on the server. On my test server a movie has two history rows and a native count of one, so this PR reported "Times viewed = 1" where it is really 2. On a shared server the plays being dropped belong to other people, and a rule like "times viewed = 0, then delete" would delete what someone else watched. The count is now whichever of the two is higher, so a scrobble still counts and nobody's views go missing.

**Changed: a failed Plex read no longer passes for "never watched".** The metadata call this used swallows errors and returns nothing, which came out the other end as not watched. The history call it replaced propagates them on purpose, because Maintainerr must never treat "could not find out" as a confirmed answer. That also removes one HTTP request per item, on the cache key the getter had already fetched.

**Changed: guid parsing.** Plain string operations instead of regular expressions, which is what the repo asks for, and the season and episode a legacy agent appends to the series id (`com.plexapp.agents.thetvdb://73141/1/1?lang=en`) are no longer read as part of the id. The provider name mapping moved up to the media-server layer, so Plex, Jellyfin and Emby share one copy instead of three.

**Still open.** The case you originally hit, an item marked unwatched that still matches because old history rows remain, is not fixed here. Telling that apart from "someone else watched it" needs to know which account a history row belongs to, and Plex does not hand that over cheaply. That is the explicit rule semantics https://features.maintainerr.info/posts/219/mark-as-watched-rule-parameter describes, and it belongs there rather than in a change to what the existing rules mean.

Checked against a real Plex server, comparing what each rule resolves on `development`, on this PR, and after the rework. The Jellyfin and Emby side is unit tests only, since their behaviour there is unchanged.

enoch85:
@noahphex PLease confirm it still works on `development`. Comment here otherwise.

noahphex:
Confirmed working in development. 

Thanks, your changes makes sense.

For my setup, the important scrobble/Letterboxd case should be covered by your rework: my Letterboxd sync marks movies watched in Plex without necessarily creating a Plex watch-history row, so using the native item view count as additional evidence should let Maintainerr still pick those up for cleanup.

I also understand why history remains the floor for the existing rules. On a shared server, replacing history with the token account’s native view count could drop other users’ views, which would make rules like “times viewed = 0” unsafe.

The remaining gap is the one I originally hit: items currently marked unwatched in Plex can still match because old history rows remain. Based on your explanation, I agree that probably needs separate explicit rule semantics rather than changing the meaning of the existing watched/view-count rules. Something like “currently marked watched by the configured Plex account” vs “has historical playback activity” would cover that distinction.

enoch85:
Thanks for confirming!

PR #3387, opened 2026-08-02
feat(rules): add Tracearr watch history application

enoch85:
Adds Tracearr as a rule Application, so watch-history rules work against Plex, Jellyfin and Emby from one source. Tautulli covers Plex only.

The history index is swept once per rule run and indexed locally by rating key rather than queried per item. Show and season scoping resolves from the grandparent rating key already present on every history row, so no TMDB/TVDB resolution is involved.

Properties mirror Tautulli, minus date added which the media server already answers: viewed by, users that watched every episode, users that watched at least one episode, times viewed, total views, watched episode count, last view date, and newest episode view date. The per-collection watched-percent override applies, using the completion percent Tracearr reports per play.

Rules fail closed. If the sweep or the user mapping does not complete, every property returns undefined so items are preserved rather than read as unwatched. Items added before Tracearr's earliest recorded play are treated as unknown for the same reason, since Tracearr records only what it observes and does not backfill from the media server.

Requires Tracearr 2.0.0-beta.1 or later, a public API v2 token, and the Tracearr server ID.


enoch85:
image

PR #3393, opened 2026-08-03
feat: add studio rules, sorting, and bulk exclusions

enoch85:
### Description & Design

This PR adds studio-aware rules, Jellyfin studio sorting, and multi-select bulk global exclusions on the Overview.

**Problem**

- Rules cannot compare a media item studio list (Fider #172).
- Browsing cannot order Jellyfin media by native studio metadata.
- Excluding several items requires opening an exclusion dialog per item (Fider #94).

**Design**

- `[list] Studios` is a metadata-backed rule value: one canonical TEXT_LIST from the metadata layer (TMDB production companies for movies, networks for TV), identical on Plex, Jellyfin and Emby, so rules migrate cleanly across servers. Resolution is shared by all three getters and memoized per run; a failed lookup stays `undefined` so the comparator protects the item. Seasons and episodes roll up to their show.
- Media-server `Studios` map onto the server-agnostic media item where Maintainerr-side sorting needs them (collection grids, exclusions, global search).
- Native `SortBy=Studio` is a capability (`LIBRARY_STUDIO_SORT`) held by Jellyfin and Emby (verified against live 10.11 / 4.9 servers), gated client- and server-side; Plex rejects the sort with a 400 and never shows the option.
- `POST /rules/exclusions/bulk` (max 250 ids per request) runs the existing `setExclusion` in bounded batches, collapses ids nested under another selected id (a show plus its own episode cannot race the same exclusion row), and reports per-item results. The client chunks large selections into smaller requests. Hierarchy cascade, global-over-scoped precedence and *arr tag sync are preserved.
- Overview multi-select: selection count on the action button, selection cleared on every content replacement (sort, library switch, search transitions), toast feedback, failed items stay selected.

**Trade-offs and limitations**

- Studio rule values are metadata-provider data, not the media server's own field; provider gaps leave the value absent (item protected) rather than fabricating an empty list.
- All three servers were live-tested: the studios rule resolves on Plex, Jellyfin and Emby (Emby verified against a live 4.9 server), the studio-sort gate rejects Plex/Emby server-side, and the bulk overlap collapse was verified on all three.

### Related issue

- [Fider #172](https://features.maintainerr.info/posts/172)
- [Fider #94](https://features.maintainerr.info/posts/94)


enoch85:
**Bulk exclusion**
image

**Sorting**
image

**Rules**
image

enoch85:
@SmolSoftBoi I will review this tomorrow evening or tonight, but if you have time it would be nice with an extra pair of eyes. :+1:

apps/server/src/modules/metadata/providers/tmdb-metadata.provider.ts
```diff
@@ -102,6 +103,20 @@ export class TmdbMetadataProvider implements IMetadataProvider {
...
   }
 
+  private getStudios(
+    record: TmdbMovieDetails | TmdbTvDetails,
+    type: 'movie' | 'tv',
+  ): string[] | undefined {
+    const values =
+      type === 'movie'
+        ? (record as TmdbMovieDetails).production_companies
+        : (record as TmdbTvDetails).networks;
```
SmolSoftBoi, line 118: The Studios rule returns networks for TV records even though TMDB exposes production companies separately. This makes the rule match broadcasters instead of production studios. Use `production_companies` here, or expose networks as a distinct property.

apps/ui/src/components/Overview/index.tsx
```diff
@@ -368,6 +406,69 @@ const Overview = () => {
...
+    if (!nextSelectionMode) {
+      setSelectedMediaIds(new Set())
+      setBulkExclusionFeedback(undefined)
+    }
+  }
+
+  const handleBulkExclusion = async () => {
+    const response: BulkExclusionResponse = await bulkExcludeMedia([
+      ...selectedMediaIds,
+    ])
```
SmolSoftBoi: Infinite scrolling can place more than 250 IDs in `selectedMediaIds`, but this call sends the whole set while the request schema rejects arrays over 250 before processing starts. Cap selection with `BULK_EXCLUSION_MAX_ITEMS` or chunk the requests.

apps/ui/src/components/Overview/index.tsx
SmolSoftBoi, line 336: When the search query becomes empty, this branch replaces the search results without clearing `selectedMediaIds`. The enabled bulk action can then exclude search items that are no longer visible. Clear selection as part of every replacement sync.

apps/server/src/modules/rules/rules.service.ts
```diff
@@ -1173,6 +1177,41 @@ export class RulesService {
     }
   }
 
+  async setBulkExclusions(mediaIds: string[]): Promise<BulkExclusionResponse> {
+    const results: BulkExclusionItemResult[] = [];
+    const uniqueMediaIds = [...new Set(mediaIds)];
+
+    for (const batch of _.chunk(uniqueMediaIds, BULK_EXCLUSION_CONCURRENCY)) {
+      const batchResults = await Promise.all(
+        batch.map(async (mediaId): Promise<BulkExclusionItemResult> => {
```
SmolSoftBoi: Exact root-ID deduplication does not cover a selected show and one of its episodes. Parallel `setExclusion` calls can expand to the same episode, both miss the existing row, and insert duplicates because the table has no matching unique constraint. Collapse overlapping roots or make the write atomic.

SmolSoftBoi (changes requested):

enoch85:
Rebased onto latest development (merge commit dropped) and pushed fixes for every review point:

- **TV networks vs production companies** - kept networks, now documented in `tmdb-metadata.provider.ts`. This mirrors Sonarr's `network` field, and both the Jellyfin and Plex TMDB agents surface TMDB networks as a series' studio, so a networks-backed rule matches what users see in their server UI. TMDB production companies on series are co-producers and would diverge from the visible value; if broadcaster-vs-producer filtering is wanted later it should be a separate property rather than a silent union.
- **250-id cap / oversized selections** - the client now chunks every selection into requests of 25 via a `useBulkExcludeMedia` mutation and aggregates per-item results, so selections above the request cap succeed and no single request carries an unbounded show cascade. The action button shows the selected count.
- **Stale selection after clearing search** - selection is now cleared at the top of every replacement sync (search entry, search exit, library switch, sort change), verified in the browser against dev Jellyfin.
- **Overlapping roots race** - `setBulkExclusions` now resolves each id's parent/grandparent first and collapses ids nested under another selected id into their ancestor (the child reports the ancestor's outcome), so a show and its own episode can no longer race the same exclusion row. Unit-covered and verified live on both Jellyfin and Plex (no duplicate rows, correct 7-row cascade).

Also fixed while re-verifying end-to-end: collection-grid hydration now requests `Studios` (studio sort on the collection media tab was a silent no-op), unneeded `Studios` payload additions were trimmed from the library list/search/children queries, the exclusion DTO now reflects its real optional fields, and the dev seed covers the new property. Full suite green; real-stack evidence (Jellyfin + Plex + Radarr tag sync) in the PR conversation is available on request.

apps/ui/src/api/rules.ts
```diff
@@ -190,6 +192,31 @@ export const useRuleConstants = (options?: UseRuleConstantsOptions) => {
...
+    mutationKey: ['rules', 'exclusions', 'bulk'],
+    mutationFn: async (mediaIds) => {
+      const results: BulkExclusionResponse['results'] = []
+
+      for (const batch of chunk(mediaIds, BULK_EXCLUSION_REQUEST_CHUNK)) {
+        const response = await PostApiHandler<BulkExclusionResponse>(
+          '/rules/exclusions/bulk',
+          { mediaIds: batch },
+        )
+        results.push(...response.results)
```
SmolSoftBoi: If a later `POST` throws, the mutation rejects and discards results from earlier chunks even though those exclusions have already been persisted. Overview then cannot update successful cards or remove them from the selection. Return accumulated results and per-item failures for failed or unattempted batches.

apps/ui/src/components/Overview/index.tsx
```diff
@@ -263,6 +286,10 @@ const Overview = () => {
   const performOverviewSync = useCallback(
     async (libraryId?: string, nextSortParams = sortParams) => {
       invalidateFetches()
+      // Every sync replaces the visible item set (search results, a library
+      // switch, a sort change, or leaving search), so a selection made against
+      // the previous set must never survive into the next one.
+      setSelectedMediaIds(new Set())
```
SmolSoftBoi, line 298: Clearing here happens before the request, while the existing cards remain rendered and selectable during loading. A user can reselect an old card and run the action before or after the replacement hides it. Disable selection and confirmation while replacing content, or clear again when the response is applied.

apps/ui/src/components/Overview/index.tsx
```diff
@@ -368,6 +395,70 @@ const Overview = () => {
...
+    setSelectedMediaIds(failedIds)
+    for (const mediaId of succeededIds) {
+      invalidateMaintainerrStatusDetails(mediaId)
+    }
+
+    if (succeededIds.size > 0) {
+      const nextItems = dataRef.current.map((item) =>
+        succeededIds.has(item.id)
+          ? { ...item, maintainerrExclusionType: 'global' as const }
+          : item,
```
SmolSoftBoi: The server cascades a selected show exclusion to its seasons and episodes, but this only marks and invalidates exact submitted IDs. In global search, visible child cards remain unmarked and their cached status may remain scoped. Refresh after the mutation or reconcile items whose `parentId` or `grandparentId` is a successful root.

SmolSoftBoi (changes requested):

enoch85:
All three points verified against the code and fixed in 9c00ef08:

- **Chunk failure discarding persisted results** - the chunk loop now catches a failed request and reports that batch plus every unattempted batch as per-item failures instead of rejecting the aggregate, so the Overview still reconciles the chunks that were persisted. Extracted as `postBulkExclusions` with a focused spec covering the mid-sequence throw.
- **Selection window during content replacement** - kept the immediate clear for feedback, and the selection is now cleared again when the replacement item set is applied (both the library and search paths), so nothing selected against outgoing cards survives into the new set.
- **Cascaded children in mixed search results** - successful roots now reconcile any visible item whose `parentId` or `grandparentId` matches: badge patched and status cache invalidated. Verified live against dev Jellyfin: searching returns a series plus its five episodes; submitting only the series id marks all six cards.

Full suite green; branch is four commits on latest development.

Issue #3344, opened 2026-07-28
Show rule "leaks" into movie section


olli991:
**Describe the bug**
Somehow a collection, which is aimed a shows is shown in movies, with a movie in it.
The one on the right:
<img width="467" height="389" alt="Image" src="https://github.com/user-attachments/assets/28be2ab6-759b-4fc7-a3c7-308cc56df4d8" />

**To Reproduce**
1. Open Movies and navigate to collections
2. Scroll down to my normal collection and see the wrong one besides it

**Expected behavior**
Collection not present in Movies because it's not meant to be.

**Screenshots**
Present rules in Maintainerr (Filme = Movies / Serien = Shows)
<img width="1213" height="551" alt="Image" src="https://github.com/user-attachments/assets/548b246a-fde3-4085-b5d8-dcd761c8215a" />

Corresponding Collections:
<img width="1221" height="531" alt="Image" src="https://github.com/user-attachments/assets/d3b5988c-2c5d-4aab-98b8-6a532951b988" />

The Collection / Rule that is somehow "leaking" into movies:
<img width="1690" height="954" alt="Image" src="https://github.com/user-attachments/assets/a2a26a84-c09f-4441-9bd6-747a1682bc71" />
```yaml
mediaType: SEASONS
rules:
  - "0":
      - firstValue: Plex.sw_allEpisodesSeenBy
        action: CONTAINS
        lastValue: Seerr.addUser
      - operator: AND
        firstValue: Sonarr.tags
        action: NOT_CONTAINS_PARTIAL
        customValue:
          type: text
          value: keep
      - operator: AND
        firstValue: Plex.watchlist_isWatchlisted
        action: NOT_EQUALS
        customValue:
          type: boolean
          value: "true"
      - operator: AND
        firstValue: Sonarr.unaired_episodes
        action: NOT_EQUALS
        customValue:
          type: boolean
          value: "true"
```

**Device (please complete the following information):**
 - OS: Linux x64 (UGREEN NAS)
 - Version 6.12.30+
 - Server: Plex

**Debug logs**
```
Add your debug logs here
```

**Additional context**
Could it be a match if there were two movie rules in the past and i changed the second one to a show rule and not deleting it and some old ID is still linked in the back?
I can't completly confirm or deny if I did this to the "leaking" rule.

enoch85:
Hey! Thanks for posting!

Do you have a debug log of when this is happening, or maybe a complete debug log?

olli991:
[...]

PR #3411, opened 2026-08-05
feat: bulk media actions from one shared modal

enoch85:
### Description & Design

A media selection now drives every collection and exclusion action through one shared "Add / Remove Media" modal, on the Overview and on both collection detail tabs. The per-item `AddModal` is deleted and the Overview card's `Add` / `Excl` hover buttons with it, so an Overview poster only carries badges. The collection pages keep their own per-card button, which is unchanged.

A one-item selection keeps the season/episode narrowing and the global-exclusion warning the old per-item modal had.

### Where each action lives

| Surface | Control | Actions | Scope |
| --- | --- | --- | --- |
| Overview, card hover | *(none, removed)* | - | - |
| Overview, selection | `Add/Exclude selected` -> shared modal | Add to collection / Remove from collection / Remove from all collections / Add exclusion / Remove exclusion | Collection picker: any collection in the library, or `All collections` |
| Collection > Media, card hover | `Exclude` (unchanged) | Removes the item from this collection **and** adds a rule-group exclusion | This collection, one item |
| Collection > Media, selection | `Add/Exclude selected` -> shared modal | Remove from collection / Remove from all collections / Add exclusion / Remove exclusion | Pinned to this collection. `Add to collection` is hidden: the items are already in it |
| Collection > Exclusions, card hover | `Remove` (unchanged) | Deletes that exclusion row; warns first when it is global | One exclusion row |
| Collection > Exclusions, selection | `Add/Exclude selected` -> shared modal | Add to collection / Remove from collection / Remove from all collections / Remove exclusion | Pinned to this collection. `Add exclusion` is hidden: the items are already excluded |
| Any card, click through | Media detail modal (unchanged) | Postpone, Trigger rule action (collection pages only) | One item |

Selection mode closes once nothing is left to act on, and stays open with only the failed items selected otherwise. Same contract in all three places.

### Endpoints

- `POST /rules/exclusions/bulk` gains `collectionId` and `action`. A collection scopes the exclusion to its rule group and drops the items from that collection, matching the per-card `Exclude` above; no collection means global, or a removal of every exclusion the items carry.
- `POST /collections/media/bulk` adds or removes a selection. It resolves each item's hierarchy in bounded-parallel batches but writes **once**: parallel per-item writes race the collection find-or-create and leave duplicate collections beside the linked one, which is #3344's failure mode.

### Fixes found while building this

- An id the media server does not hold became a membership row for media that does not exist. The bulk path refuses it; an inconclusive lookup is still taken as present so a blip never blocks a real add.
- Emby resolved a single id through the unscoped `/Items/{id}` route when `emby_user_id` is unset. Emby answers 404 there for items that exist, so `itemExists` reported every live item as deleted and `getMetadata` found nothing. Both now use the list form, matched on the exact id.
- `setExclusion` and `removeExclusionWitData` threw on a collection with no rule group.

### Also

- Library cards carry a badge naming the collections an item belongs to.
- The provider id badges link out to TMDB, IMDb and TheTVDB.
- lodash default imports replaced with named ones across the server. `notifications.service.ts` had a local `isEqual` helper that only reached lodash through the `_` namespace; the sweep would have made it shadow the new named import, so the local is renamed to `sameAgents`.

### Review round

Two proven bugs and the secondary findings are fixed in `6b18892`:

- `removeBulkExclusions` left every cascaded season/episode row behind while reporting success. It matches `parent` too now.
- "Remove from all collections" left ghost cards on the Media tab.
- A scoped exclusion-add now invalidates the collection caches it changes.
- Media type is read off the picked items instead of the page, so an exclusion list's parent-type rows cannot be handed to the wrong collection type (#3382's shape).
- A narrowed season/episode travels as a `context` again, keeping the show as the media id so `exclusion.parent` still points at it. That is what the parent card's exclusion badge, `getExclusions` and the *arr tag sync read.
- The nesting collapse now runs for scoped requests too: a season collection expands a selected show to its seasons, colliding with a separately selected season.
- `removeFromAllCollections` reports its own failures; the bulk branch was discarding them.
- One provider-url builder, so the badges get the Sportarr tvdb-alias guard and stop rendering dead links.
- The per-card `Exclude` uses the same endpoint as the modal, so both use the server's exclude-then-drop order.

Verified against the real dev stack on Jellyfin, Emby 4.9.5 and Plex, reading state back from each media server, plus Playwright for the UI.

CodeQL flags 3 SSRF alerts on the new Emby `fetchItem` lines. Same rule and same cause as the sibling alerts already dismissed as "won't fix" in that file (the media server URL is user-configured by design).

### Second review round

Verifying that round found four more, fixed in `d646012`. Every one is the same mistake: acting on the id the selection entered through instead of the ids it resolves to.

- A narrowed un-exclude dropped its `context` at the controller. Un-excluding one season deleted every row the show had cascaded, including the show's own, and reported success. Proven on Jellyfin: 7 rows in, 7 gone; now 6 go and the show stays excluded.
- The scoped add paired the collection drop with the entry point. A season or episode collection holds the resolved children, so the drop matched nothing and the excluded item stayed in the collection. `setExclusion` now reports the ids it wrote and the drop uses those.
- Both exclusion paths named the entry point after the collection's own `dataType`, which stopped the traversal before it started: a show reaching a season collection wrote a single row for the show id, typed `season`, in a season rule group. They use the item's own type now, as the collection bulk path already did.
- The collection picker read its library off the media item. No media server fills that field, so it was always empty and every library's collections were offered, which is the rejected-add shape #3383 fixed. The page supplies the library again, and supplies none while a search spans them. This one was mine, from the first review round.

A pinned show collection no longer offers season narrowing: a show collection acts on the show itself, so the choice promised a reach the action never had. The unpinned picker already handled it by dropping show collections once a season is picked.

The external review landed against the pre-fix commit. Its two P1s are the same two bugs above, found independently. Its third P1 was half right: the caller propagation was already in, but the helper it names was swallowing failures of its own, so `removeFromAllCollections` reported a partial failure as a complete success. Fixed. Its remaining findings were already fixed or no longer apply, except the optimistic Overview patch, which stays open.


apps/server/src/modules/rules/rules.service.ts
```diff
@@ -1205,6 +1245,100 @@ export class RulesService {
...
+            code: 0 as const,
+            message: 'Failed - no rule group',
+          })),
+        };
+      }
+      ruleGroupId = group.id;
+    }
+
+    const exclusions = await this.exclusionRepo.find({
+      where: { mediaServerId: In(uniqueMediaIds) },
```
SmolSoftBoi: [P1] Remove descendant exclusion rows too

`setExclusion()` expands a show or season into child rows and stores the submitted ID in each row's `parent` field. This query only loads rows whose `mediaServerId` exactly matches a submitted ID, so removing the top-level exclusion leaves its descendant exclusions active while returning success. Include rows whose `parent` matches a submitted ID, apply the same scope filtering, and map their removal outcome back to the submitted root.

apps/server/src/modules/rules/rules.service.ts
```diff
@@ -1194,7 +1209,32 @@ export class RulesService {
...
+    if (collectionId !== undefined) {
+      const excludedIds = rootIds.filter(
+        (mediaId) => resultById.get(mediaId)?.code === 1,
+      );
+
+      if (
+        excludedIds.length > 0 &&
+        !(await this.collectionService.removeFromCollection(
+          collectionId,
+          excludedIds.map((mediaServerId) => ({ mediaServerId })),
```
SmolSoftBoi: [P1] Remove the resolved collection members

For a show submitted to a season or episode collection, `setExclusion()` expands the show to the child IDs that the collection contains, but this follow-up sends only the original show ID to `removeFromCollection()`. Nothing is removed from that collection and the bulk response still reports success. Preserve or re-resolve the child IDs used by `setExclusion()`, remove those IDs, and map any failure back to the submitted show.

SmolSoftBoi (commented):
Reviewed current head ed8439a. I found six actionable issues, including three correctness problems that can report success without reaching the requested state. Details are inline.

apps/server/src/modules/collections/collections.service.ts
```diff
@@ -2545,6 +2552,133 @@ export class CollectionsService {
...
+          } else if (result.serverRejectedIds.length > 0) {
+            const rejected = new Set(result.serverRejectedIds);
+            for (const [mediaId, ids] of resolvedByMediaId) {
+              if (ids.some((id) => rejected.has(id))) {
+                fail(mediaId, 'Failed - refused by the media server');
+              }
+            }
+          }
+        } else if (collectionId === undefined) {
+          await this.removeFromAllCollections(media);
```
SmolSoftBoi: [P1] Propagate remove-all failures

`removeFromAllCollections()` returns `{ code: 0 }` on its own failure, but this caller ignores that result. The helper also ignores `undefined` returned by an individual `removeFromCollection()` call, so a partial media-server or database failure can still make every item here fall through as `code: 1`. Return explicit removed and failed IDs from the helper, or at minimum propagate its failure into `failAllResolved()`, so the interactive bulk response reflects the actual outcome.

apps/ui/src/components/Common/MediaActionModal.tsx
```diff
@@ -0,0 +1,533 @@
...
+        .map((result) => result.mediaId)
+
+      if (isCollectionAction && succeededIds.length > 0) {
+        await invalidateCollectionQueries(queryClient)
+      }
+
+      onSubmitted({
+        action: selectedAction,
+        collectionId,
+        succeededIds,
```
SmolSoftBoi, line 315: [P2] Keep narrowed failures retryable

When one selected show is narrowed to a season or episode, `submittedIds` contains the child ID, so a failed response replaces the visible show's selection with an off-screen child ID. The modal then closes with no visible selected card, and Overview derives no `mediaType` for that hidden selection, which prevents retrying collection actions. Keep the source selection ID separate from the submitted narrowed ID, or keep this modal open on failure, and add a narrowed-failure retry test.

apps/ui/src/pages/CollectionMediaPage.tsx
```diff
@@ -124,13 +135,56 @@ const CollectionMediaPage = () => {
...
+
+  const handleBulkOutcome = ({
+    action,
+    succeededIds,
+    failedIds,
+  }: MediaActionOutcome) => {
+    applyBulkOutcome(new Set(failedIds))
+
+    // Only these two drop the item from this collection.
+    if (action === 'exclusion-add' || action === 'collection-remove') {
```
SmolSoftBoi: [P2] Prune successful remove-all items

`collection-remove-all` also removes each successful item from the collection displayed by this page, but this branch only prunes `exclusion-add` and `collection-remove`. After a successful remove-all action, the cards remain visible until a reload. Include `collection-remove-all` here and cover it in the page test.

apps/ui/src/components/Overview/index.tsx
```diff
@@ -452,34 +426,40 @@ const Overview = () => {
...
+      // Only a global exclusion changes what a library card shows; a scoped one
+      // and the collection actions are invisible here.
+      const nextExclusionType =
+        collectionId === undefined && action === 'exclusion-add'
+          ? ('global' as const)
+          : action === 'exclusion-remove'
+            ? undefined
+            : null
+
+      if (nextExclusionType !== null) {
```
SmolSoftBoi: [P2] Reconcile authoritative Overview status

This path only patches `maintainerrExclusionType`. Collection badges render from `maintainerrCollections`, so add, remove, and remove-all actions leave those badges stale until a full reload. For `exclusion-remove`, `isCovered` also clears every visible descendant of a successful parent even when a descendant has an independent exclusion. Refetch or re-enrich affected items and replace the authoritative collection and exclusion fields, then test collection badge updates and independent child exclusions.

apps/server/src/modules/rules/rules.service.ts
enoch85: Confirmed and fixed in 6b18892: the query now matches `[{mediaServerId: In(ids)}, {parent: In(ids)}]` and maps each cascaded row back to the selected root. Proven on Jellyfin, where a global show exclusion writes 7 rows and un-excluding cleared 1 of them before the fix and all 7 after.

apps/server/src/modules/rules/rules.service.ts
enoch85: Confirmed and fixed in d646012: `setExclusion` now returns `handledIds` (the ids it resolved to) and the drop uses those instead of the submitted show. Proven live on a season collection, where the membership row was the season id and the old drop matched nothing.

apps/server/src/modules/collections/collections.service.ts
enoch85: Both halves fixed. The caller propagation landed in 6b18892; the sub-point about the helper swallowing `undefined` from each `removeFromCollection` was real and is fixed in 8504fe99 - `removeFromAllCollections` now returns NOK if any collection failed.

apps/ui/src/components/Common/MediaActionModal.tsx
enoch85: This one no longer applies: since 6b18892 a narrowed season travels as a `context` and the show stays the submitted media id, so the response is keyed by the visible show and the failed selection stays on-screen. Verified against the live response.

apps/ui/src/pages/CollectionMediaPage.tsx
enoch85: Fixed in 6b18892: `collection-remove-all` is included in the prune condition, covered by a page test and confirmed in the browser.

apps/ui/src/components/Overview/index.tsx
enoch85: Half of this is fixed: collection badges are patched for add/remove/remove-all via `maintainerrCollections`. The descendant over-clear is real and left open, along with the stale badge after a scoped exclusion-add - both are optimistic-patch precision issues that self-correct on refetch, and narrowing the patch to exact ids would make the genuine cascade case stale instead. Tracked for a follow-up that re-enriches the affected items.

PR #3426, opened 2026-08-06
feat(rules): scope watch statistics to one user

enoch85:
Rules can only ask what the whole household did. This adds three properties to each watch-history integration, so a rule can ask about one person: "keep anything alice watched more than twice". [Feature request 224](https://features.maintainerr.info/posts/224/streamystats-per-user-rules) asked for it on Streamystats; the other two answer the same question from their own history, so all three use the same names.

| Property | Streamystats | Tautulli | Tracearr |
| --- | --- | --- | --- |
| Times viewed by user | yes | yes | yes |
| Watch time by user (minutes) | yes | yes | yes |
| Last view date by user | yes | yes | yes |

The user is picked on the rule itself, next to the property, the same way ARR disk targets are. Rules are stored as JSON, so there is no database migration.

### Why a username and not a user id

Tautulli reports the plex.tv account id, the media server reports its own local one, and nothing maps cleanly between all three. Every other user-facing rule property already compares usernames. The picker offers the same list the rules resolve against (the plex.tv-corrected one on Plex) and filters as you type, so a server with a thousand users does not put a thousand options in every rule card.

### What happens when a user cannot be resolved

| Situation | Behaviour |
| --- | --- |
| Name is not one the media server reports | Save is rejected, with the name in the message |
| User exists but never watched the item | 0 plays, no date |
| User is renamed or deleted after the rule is saved | Rule pauses on those items, never reads as "watched nothing" |
| Rule is uploaded to the community list | Username is stripped, since it names someone's household |

### What counts

| Property | Counts |
| --- | --- |
| Times viewed, last view date | Watched plays only, honouring the collection's watched-percent override |
| Watch time | Every play, since an abandoned one still ran for its minutes |

Streamystats has no completion filter to apply, and keeps no session against a season, so seasons answer no value rather than a zero that would read as never watched.

Verified against the real dev stack rather than mocks: all three integrations with real playback, a full rule run, and the editor at 1003 users.

Also here: the rules regression matrix (`yarn workspace @maintainerr/server test:e2e`) could not start any more, because it builds `RulesService` and never gained the dependencies that service picked up since. Fixed, and all 107 scenarios produce results again.

### DIFF
--- a/apps/server/src/modules/api/tracearr-api/tracearr-api.service.ts
+++ b/apps/server/src/modules/api/tracearr-api/tracearr-api.service.ts
@@ -245,7 +245,7 @@ export class TracearrApiService {
     this.logger.log(
       `Tracearr history starts at ${new Date(historyIndex.earliestStartedAt).toISOString()}. Earlier unobserved media is skipped.`,
     );
-    const usernames = await this.fetchUsernamesByTracearrUserId(historyIndex);
+    const usernames = await this.fetchUsernamesByTracearrUserId();
     if (!usernames) {
       this.logger.warn(
         'Tracearr media-server user lookup did not complete. Tracearr username rule values are unavailable for this run.',
@@ -256,6 +256,11 @@ export class TracearrApiService {
     this.activeUsernamesByTracearrUserId = usernames;
   }
 
+  /**
+   * Tracearr's public history only returns a play once one of its segments ran
+   * for 2 minutes or more, so every Tracearr rule property counts sustained
+   * plays only.
+   */
   private async refreshHistoryIndex(): Promise<
     TracearrHistoryIndex | undefined
   > {
@@ -455,9 +460,9 @@ export class TracearrApiService {
     return name || id;
   }
 
-  private async fetchUsernamesByTracearrUserId(
-    historyIndex: TracearrHistoryIndex,
-  ): Promise<Map<string, string[]> | undefined> {
+  private async fetchUsernamesByTracearrUserId(): Promise<
+    Map<string, string[]> | undefined
+  > {
     const api = this.api;
     const serverId = this.settings.tracearr_server_id;
     if (!api || !serverId) {
@@ -472,9 +477,6 @@ export class TracearrApiService {
 
     const usernamesByAccountId = new Map(
       mediaUsers.map((user) => [user.id, user.name]),
-    );
-    const historyUserIds = new Set(
-      [...historyIndex.rowsById.values()].map((row) => row.user.id),
     );
     const usernamesByTracearrUserId = new Map<string, string[]>();
     const cursors = new Set<string>();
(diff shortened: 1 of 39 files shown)


apps/server/src/modules/api/tracearr-api/tracearr-api.service.ts
```diff
@@ -499,14 +501,22 @@ export class TracearrApiService {
...
+        // fails closed). Both spellings are kept because the media server and
+        // Tracearr can name the same Plex account differently.
+        const usernames = [
+          ...new Set(
+            user.accounts
+              .filter((account) => account.server_id === serverId)
+              .flatMap((account) => [
+                usernamesByAccountId.get(account.external_user_id),
+                account.username,
+              ])
```
SmolSoftBoi, line 519: Tracearr includes removed accounts in each identity's account block via `removed_at`. Because this filter checks only `server_id` and then falls back to `account.username`, a departed account can remain resolvable and continue driving existing and per-user rules. Parse `removed_at` and require it to be `null` before mapping the account.

apps/server/src/modules/rules/rules.service.ts
```diff
@@ -1887,6 +1904,75 @@ export class RulesService {
...
+    // block a save; a populated one that lacks the user means the rule would
+    // skip every item, so reject it here instead.
+    if (
+      rule.username &&
+      knownUsernames.length > 0 &&
+      !knownUsernames.includes(rule.username)
+    ) {
+      return this.createReturnStatus(
+        false,
+        `The media server has no user named '${rule.username}'`,
```
SmolSoftBoi, line 1984: This rejects a saved username as soon as the account is renamed or deleted, even though the editor intentionally preserves that value and execution fails closed. Any later edit to the rule group is therefore blocked. Allow usernames already persisted in this group on the update path while still rejecting newly introduced unknown names.

apps/server/src/modules/rules/getter/tautulli-getter.service.ts
```diff
@@ -221,6 +236,68 @@ export class TautulliGetterService {
...
+    }
+
+    const history = (await this.getHistoryForMetadata(metadata)).filter(
+      (el) => el.user_id === plexUser.plexId,
+    );
+
+    if (propName === 'watchTimeByUser') {
+      const seconds = history.reduce(
+        (total, el) => total + (el.play_duration ?? 0),
+        0,
```
SmolSoftBoi, line 279: Tautulli versions before 2.12.3 return `duration` rather than `play_duration`, and Maintainerr does not enforce a newer minimum. These installations therefore calculate every user's watch time as zero. Type both fields and use `play_duration ?? duration ?? 0`.

SmolSoftBoi (changes requested):

apps/server/src/modules/api/tracearr-api/tracearr-api.service.ts
enoch85: Good catch, and it is worse than it looks: before this PR the map was built only from media-server users, so a departed account was already unresolvable. The username union I added is what brought it back, and a per-user rule would then read "watched nothing" for someone the server no longer has, instead of skipping the item.

Fixed in f487ba8ae: `removed_at` is now parsed and removed accounts are left out of the map, which restores the earlier behaviour rather than changing it. Covered by a test with a `removed_at` account.

apps/server/src/modules/rules/rules.service.ts
enoch85: Agreed, that was too blunt. I had accepted the trade deliberately, on the grounds that nothing is lost silently, but blocking every later edit to the group is a real regression.

Fixed in f487ba8ae: the update path now allows usernames the group already has saved, while a newly introduced unknown name is still rejected. So an import cannot bring a foreign user in, and a renamed or deleted account no longer freezes editing - the rule still pauses on the items it cannot resolve, which is the intended behaviour. Two tests cover saved versus new.

apps/server/src/modules/rules/getter/tautulli-getter.service.ts
enoch85: Confirmed against Tautulli's source, not just the changelog: v2.12.2 emits only `'duration': item['play_duration']` (datafactory.py:311), and 2.12.3 onwards emits both keys from the same expression - played seconds with the paused counter subtracted. So it was a response-key rename with the old name kept, and the values are identical.

Fixed in f487ba8ae with `play_duration ?? duration ?? 0`, both typed as optional, plus a test using the pre-2.12.3 shape.

I did not add a minimum version for it: the data is already there under its old name, and every other Tautulli property works on older versions, so a gate would reject installs to fix something that is not missing. `MINIMUM_TRACEARR_VERSION` is a different case - Tracearr's whole v2 public API did not exist before 2.0.0-beta.1.

PR #3464, opened 2026-08-11
feat(notifications): include type, title and provider ids in webhook media items

jamcalli:
### Description & Design

The webhook's `mediaItems` carries only the `mediaServerId`. For *Media Handled* events that id points at something that was removed and thus downstream external tools have no way to reliably validate what was just actioned. 

**EDIT** - for context: I maintain Pulsarr and having it play nicely with Maintainerr is a frequently requested feature. The current problem is that if users never prune their watchlists, and then a Maintainerr rule actions those items, without them being added to Pulsarr's exclusions list, they would be re-added. If the webhook payload contained `providerIds` and `type`, they could be matched and added to the exclusions list automatically. 

It would be helpful if some of the metadata snapshot was also included in the payload, since it's already fully available at the call site:

`type` - needed for provider id collisions between types
`title` - currently exists only inside the `message` but would be helpful
`providerIds` - `{ imdb: string[], tmdb: string[], tvdb: string[] }` which are already existing, normalized from the mappers

The snapshot is best-effort, so the new fields appear only when it resolved. Failed lookup falls back to the current bare shape, and events that never carry metadata (added/removed/failed/overlay) are unchanged. 

An edge case: unmatched/personal media resolves with empty provider id arrays, so consumers need to handle `{ imdb: [], tmdb: [], tvdb: [] }`.

If this PR is accepted, I will submit a follow-up to add the updated payload to the docs. 

### Related issue

NA

### AI-Assisted Development

I used Claude to investigate, sanity check, write the test, and ensure everything matched the codebase conventions exactly. I reviewed and tested everything myself, including capturing the real payload from a live setup to confirm the shape. 

### Checklist

### How to test

1. `cd apps/server && yarn test` - suite is green, including the new spec case
2. Live: subscribed a webhook agent to Media Handled and let a rule group handle a single movie. The received `mediaItems` carry `type`, `title` and `providerIds`; Media Added/Removed payloads are unchanged.

### Additional context

Captured from a live run (Plex + Radarr, unmonitor action):

```json
{
  "notification_type": "MEDIA_HANDLED",
  "collectionName": "test-collection",
  "mediaItems": "[{\"mediaServerId\":\"116021\",\"type\":\"movie\",\"title\":\"The Rain People\",\"providerIds\":{\"imdb\":[\"tt0064873\"],\"tmdb\":[\"59231\"],\"tvdb\":[\"69565\"]}}]"
}
```

### DIFF
--- a/apps/server/src/modules/notifications/notifications.service.spec.ts
+++ b/apps/server/src/modules/notifications/notifications.service.spec.ts
@@ -233,6 +233,47 @@ describe('NotificationService', () => {
     });
   });
 
+  describe('media identity on the wire', () => {
+    it('puts type, title and provider ids on the wire when the snapshot resolved', async () => {
+      const { service } = createService();
+      const sendNotification = jest
+        .spyOn(service, 'sendNotification')
+        .mockResolvedValue(undefined);
+
+      await service.handleNotification(
+        NotificationType.MEDIA_ABOUT_TO_BE_HANDLED,
+        [
+          {
+            mediaServerId: '1',
+            metadata: {
+              title: 'A Sample Movie',
+              type: 'movie',
+              providerIds: { imdb: ['tt0111161'], tmdb: ['278'], tvdb: [] },
+            } as any,
+          },
+          { mediaServerId: '2' },
+        ],
+        undefined,
+        3,
+      );
+
+      const payload = sendNotification.mock.calls[0][1];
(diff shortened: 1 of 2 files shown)


enoch85:
Thanks for this PR, looks good! :+1: 

I will just fix some nits before merge.

PR #3471, opened 2026-08-11
feat(overlays): draw the countdown on everything that leaves with a collection

enoch85:
Implements [Overlay inheritance for shows](https://features.maintainerr.info/posts/174/overlay-inheritance-for-shows).

An overlay-enabled collection also draws its countdown on everything that leaves the library with its members. No new setting - the collection's action decides. Inherited items are only drawn on, never added to the collection.

| Collection + action | Also drawn on |
|---|---|
| Show + delete entire show / unmonitor + delete all or existing episodes | its seasons and their episodes |
| Season + delete season (with or without "delete show if empty") / delete existing episodes | that season's episodes, and the show once no season of it is left outside |
| Episode + delete episode | the season once no episode of it is left outside, then the show once no season is left |
| Any + unmonitor / unmonitor show if empty / do nothing / change quality profile | nothing |

| | |
|---|---|
| Date drawn | The last child to go, when the parent empties |
| Specials | A leftover season 0 does not disqualify a show; a season with no index does |
| Already a member | An item in an overlay collection keeps its own countdown |
| Unreadable hierarchy | What was drawn is kept, and one unreadable branch does not cancel the others |

Fixed along the way:

| | |
|---|---|
| Deleting a collection stranded its overlays | The state rows cascaded away with it, so posters stayed overlaid. Now reverted before the row goes |
| `OverlaysModule` -> `CollectionsModule` | Dropped: the overlays module owns its three collection reads (two repositories, read-only) |
| `CollectionsModule` <-> `RulesModule` | Dropped: the controller only needed `isProcessing()`, now a flag on `ExecutionLockService` |

`forwardRef` count across the server: 0.


enoch85:
@MrLinford @SmolSoftBoi FYI for now (it's not done yet), but wouldn't hurt with some more eyes and testers on this. :pray:

PR #3510, opened 2026-08-15
feat(i18n): make the remaining UI strings translatable

enoch85:
Weblate was wired up, but only 38 messages ever reached it - the rest of the UI was hard-coded English. This wraps every user-visible string in `apps/ui` in the Lingui macros, then hardens the pipeline so a translation can only ever change display text.

| | Before | After |
| --- | --- | --- |
| Messages in the catalog | 38 | 1090 |
| UI modules using Lingui | 6 | 102 |
| Locales | 13 | 13 |

## What changed

| Area | Change |
| --- | --- |
| Components | `<Trans>` for JSX text, `` t` ` `` from `useLingui()` for attributes, toasts and handlers - every component consumes the i18n context, so a live language switch relabels mounted screens |
| Plain modules (`utils/`, `api/`) | `@lingui/core/macro`'s `t`, resolved when the function runs; effects and async callbacks use it too, so `t` never sits in a dependency array as a refetch trigger |
| Module-scope constants | `msg` descriptors or builder functions - option lists, sort labels, arr actions, log types. A constant would freeze in whichever locale loaded first |
| Counted copy | `plural()` / `<Plural>` instead of a suffixed `s`; every placeholder is named |
| Assembled sentences | Whole messages per branch with named placeholders, never a translated noun or verb spliced into another message |
| Form validation | zod `error` callbacks, resolved per parse |
| Dates and times | `Intl` with the active locale - no hand-rolled month or weekday messages |

## Translations never change behavior

A translation may change display text and nothing else:

- Nothing translated feeds logic: no comparison, map or query key, DOM id, sort value, stored value or API payload is ever a translated string. Calendar entries group and cache on a stable action key; the persisted default collection name is a constant.
- Wrapping is outcome-identical: English output is byte-for-byte unchanged (real plurals and two label dedups are the only deliberate copy fixes).
- Tokens a translator must not touch live outside the message: units (`(MB)`), cron expressions, file extensions, example values and domain names arrive as named placeholders fed from the same constant the code uses, so message and behavior cannot disagree.

## Security: translators are untrusted input

- URLs and domains are never translatable. `href`s live in code; visible link text and domain mentions ride through as named placeholders, so no translation can display a target the link does not open.
- `yarn i18n:validate` (required check) now also rejects: the Trojan Source bidi set and invisible characters, Zalgo combining runs, oversized entries, any translation introducing a URL marker (`://`, fullwidth-colon homoglyph, `www.`, `mailto:`, `tel:`) absent from its source, and any URL marker in a source message at all. On any pull request that is not a Weblate sync it also rejects a translated `msgstr` altered or added outside Weblate. The catalog workflow runs the validator from the base branch, so a pull request cannot weaken the gate judging its own catalogs.
- The Weblate automation is pinned: the merge only lands the exact commit the code owner approved (`--match-head-commit`), refuses any auto-merged pull request touching files outside `apps/ui/src/locales/`, and the back-translation review executes the base branch's tooling rather than pull-request code.

## Deliberately left in English

Product names on their own; universal acronyms and notation (`URL`, `TLS`, `X`/`Y`/`W`/`H`); units and the sizes built from them; badge abbreviations (`BETA`, `EXCL`, `MANUAL`); example values and placeholders; CSS tokens; image `alt` text; overlay-template seed content (stored data, not chrome); and server log lines.

Conventions, the review checklist and the security rules live in [`.github/instructions/i18n.instructions.md`](.github/instructions/i18n.instructions.md).


enoch85:
@SmolSoftBoi Can you please review this?

Please note the hard (strict) rules in the instructions. We don't trust translators, and we only translate text, not logic, and values.

.github/workflows/weblate_review.yml
```diff
@@ -2,8 +2,14 @@ name: Weblate review
...
 # and posts a back-translation table so the whole review happens on GitHub.
+#
+# pull_request_target runs this workflow (and the tools/ it executes) from the
+# BASE branch, and the back-translation step holds the AI key and a PR-write
+# token. A pull request must therefore never have its code executed here: the
+# steps below check out the base, then overlay ONLY the pull request's *.po
+# files as data. Nothing from the pull request head is ever run.
 on:
-  pull_request:
+  pull_request_target:
```
SmolSoftBoi, line 12: [P1] Gate the secret-backed AI job to Weblate authors

`pull_request_target` exposes the base workflow's secrets to every PR that changes a locale. The author test below only chooses labels; it does not prevent `Back-translate changed strings` from receiving `AI_MODEL_API_KEY`. Because `back-translate.mjs` batches every changed non-empty `msgstr` without a total entry, byte, or call cap, any contributor can submit a locale-only PR that spends repository AI quota before the separate catalogue integrity check rejects it. Gate the AI step or job on `github.event.pull_request.user.login` starting with `hosted-weblate` (do not trust the branch name), and enforce a total input or call limit.

tools/i18n/validate-catalogs.mjs
```diff
@@ -5,42 +5,328 @@
...
+          `    source:      ${entry.msgid}\n` +
+          `    translation: ${entry.msgstr}\n` +
+          `    introduces "${introducedMarker}" - a translation must never add a URL`,
+      );
+      continue;
+    }
+
+    const expected =
+      sourceArguments.get(entryKey(entry)) ?? icuArguments(entry.msgid);
     const actual = icuArguments(entry.msgstr);
```
SmolSoftBoi, line 360: [P2] Validate Lingui rich-text slot parity

This compares only ICU argument names. Numeric element slots such as `<0>...</0>` are part of Lingui's runtime contract, but `icuArguments()` ignores them, so source `Click <0>here</0>` and translation `Click <1>here</1>` pass while `<Trans>` cannot find component 1 and renders the text without the intended link or component. The runtime spec also supplies slots 0 through 9 generically and only selects messages with ICU arguments, so it misses markup-only messages and can mask invented slots. Validate that each translation has the same balanced numeric slot set as its source, then have the spec provide only source-declared slots.

apps/ui/src/components/Settings/useSettingsFeedback.tsx
```diff
@@ -25,21 +37,20 @@ export const useSettingsFeedback = (scope = 'Settings') => {
...
+  // active locale would keep serving the language that was loaded when it
+  // first ran. The callbacks below stay stable anyway - their dependency is
+  // the string's value, not an object identity.
+  const updated = messages?.updated ?? t`Settings updated`
+  const updateError = messages?.updateError ?? t`Settings could not be updated`
 
   const showUpdated = useCallback(() => {
-    showFeedback('success', scopedMessages.updated)
-  }, [scopedMessages.updated, showFeedback])
+    showFeedback('success', updated)
```
SmolSoftBoi: [P2] Resolve feedback after locale activation

`showUpdated` captures the translated string from the render that created the submit handler. `Main` awaits `setLocale(locale)` and then calls that pre-switch `onUpdated`, so the stored feedback text is still from the previous locale; later rerenders cannot translate the already stored string. Store a descriptor or status and translate while rendering, or use a late-resolving builder or global `t` after activation instead of capturing the rendered string.

apps/ui/src/components/StorageMetrics/index.tsx
```diff
@@ -146,131 +164,153 @@ const StorageMetrics: React.FC = () => {
   const hasCleanupActivity = cleanupTotals.itemsHandled > 0
   const hasAnyTotal = totals.totalSpace > 0
   const hasAnyFree = totals.freeSpace > 0 || totals.mountCount > 0
-  const mountLabel = (count: number) =>
-    `${count} mount${count === 1 ? '' : 's'}`
+  const mountCount = totals.mountCount
+  const generatedAt = new Date(metrics.generatedAt).toLocaleString()
```
SmolSoftBoi: [P2] Format with the selected app locale

Calling `toLocaleString()` without a locale uses the browser or OS preference, not `i18n.locale`. The app can be switched independently, so a Swedish UI in an English browser still shows English date and time formatting. Pass the active Lingui locale here. The same issue appears in `CollectionLogsTable`, `CollectionInfo`, `StreamystatsStatsPanel`, and `Settings/Logs`; the helper used by these views should accept the app locale too.

tools/i18n/verify-locale-switch.mjs
```diff
@@ -0,0 +1,211 @@
...
+    )
+  }
+  const browser = await chromium.launch()
+  const failures = []
+
+  try {
+    const context = await browser.newContext()
+    // Seed the stored preference before the app boots, so the first render is
+    // already Swedish rather than switching mid-session.
+    await context.addInitScript(() => {
```
SmolSoftBoi: [P2] Exercise an in-session locale switch

Seeding Swedish before boot only tests initial locale loading; it never drives the language picker or changes locale while the component tree remains mounted, which is the behaviour this verifier claims to cover. This cannot catch callbacks or rendered state frozen in the old locale. Start in English, switch to Swedish in-session, and assert the probes after the mounted tree updates. Also track the no-library alias probe as skipped: the final loop currently prints it as `ok` even after the skip warning.

tools/i18n/verify-locale-switch.mjs
```diff
@@ -0,0 +1,211 @@
...
+
+  return (
+    contents.slice(0, lineStart) +
+    `msgstr "${sentinel}"` +
+    (lineEnd === -1 ? '' : contents.slice(lineEnd))
+  )
+}
+
+const patch = () => {
+  copyFileSync(catalog, backup)
```
SmolSoftBoi, line 103: [P2] Recover an existing probe backup before overwriting it

`patch()` always copies the current catalogue over `.verify-backup`. If the process is interrupted after patching, the backup contains the clean file and `sv.po` contains sentinels; the next run overwrites that clean backup with the sentinel file, then restores and deletes it, losing the original catalogue, including uncommitted changes. Treat an existing backup as recovery state before creating a new one, and add signal cleanup or another atomic restoration strategy.

SmolSoftBoi (changes requested):
Requesting changes for six concrete issues in the new i18n pipeline. The inline comments cover unrestricted secret-backed AI usage, missing rich-text slot validation, stale-language settings feedback, date/time formatting that ignores the selected app locale, a verifier that does not perform a live switch and misreports a skipped probe, and unsafe interrupted-run backup handling.

I validated these findings against `fc467d30ff6fe41cd1ac78d42b71c6419acab117`. Local lint, typecheck, 337 UI tests, catalogue checks, and the UI production build passed; the findings are behavioural and security gaps not covered by those checks.

.github/workflows/weblate_review.yml
enoch85: Fixed in db2fc635. The AI-adjacent steps now run only when the PR author is hosted-weblate (branch names are deliberately not trusted; labelling still runs for every l10n PR), and back-translate caps a run at 500 strings with a 1000-char entry limit - anything cut is reported visibly, with a distinct summary for capped-out vs provider failure.

tools/i18n/validate-catalogs.mjs
enoch85: Fixed in db2fc635. richTextSlots() in po.mjs feeds a per-translation slot-set parity + balance check in validate-catalogs (red-tested: a renumbered <1> and an unbalanced <0> both fail, real catalogs pass), and catalog-icu.spec now supplies only the slots each message declares instead of the generic 0-9 spread.

apps/ui/src/components/Settings/useSettingsFeedback.tsx
enoch85: Fixed in db2fc635. showUpdated/showUpdateError store which message to show and the title resolves at render against the caller's freshly translated strings, so a save right after switching confirms in the new language. A spec covers the pre-switch-closure case.

apps/ui/src/components/StorageMetrics/index.tsx
enoch85: Fixed in db2fc635. All five files now format with i18n.locale (the panel's formatDate helper takes it as a parameter), plus the adjacent count formatting in StorageMetrics for consistent grouping. StreamystatsStatsPanel also gained the useLingui subscription it needed to follow a live switch at all.

tools/i18n/verify-locale-switch.mjs
enoch85: Fixed in db2fc635. The verifier now boots in English (asserting no sentinel leaked), drives the real picker + Save, and requires the sentinel to appear in the still-mounted navigation without any navigation - then re-checks after a fresh load. A skipped probe prints skip, never ok.

tools/i18n/verify-locale-switch.mjs
enoch85: Fixed in db2fc635. patch() restores a leftover backup before writing a new one (verified by simulating an interrupted run - the recovered sv.po came back byte-identical), and SIGINT/SIGTERM handlers restore the catalog since finally does not run on a signal.

enoch85:
@SmolSoftBoi when you have time 🙏🏼

tools/i18n/po.mjs
```diff
@@ -89,12 +108,132 @@ export const parsePo = (text) => {
...
+ * this parser. Two shapes proved to slip a translation past every content
+ * check while Lingui still bound it to a live message:
+ *
+ *   - an obsolete `#~` block. Comments are skipped here, so the live entry
+ *     reads as untranslated - but Lingui restores the commented msgstr onto
+ *     the live message id.
+ *   - `msgstr` written above its `msgid`. This parser starts an entry at
+ *     msgid and never sees the stray translation; Lingui binds it.
+ *
+ * Rather than chase Lingui's leniency, reject anything outside the canonical
```
SmolSoftBoi, line 76: [P1] Reject PO escapes that decode differently at runtime

`unescape()` drops the backslash for unrecognised escapes, while the `pofile-ts` parser used by Lingui decodes PO octal escapes. A minimal source-locale entry with `msgstr "Add/Exclude <\0>selected</\0>"` passes `i18n:validate`: this parser reconstructs the original source and reports a balanced `<0>` slot, but Lingui produces U+0000 characters and no rich-text slot. This bypasses the English `msgstr === msgid` integrity check as well as the character and slot checks; `i18n:check` only compares message IDs, so it does not catch the differing runtime text. Reject noncanonical escapes before validation, or match Lingui's decoding exactly, and add this case as a regression test.

SmolSoftBoi (changes requested):
Requesting changes for one remaining catalogue-integrity bypass. The six earlier review findings are fixed; this review concerns the custom PO escape decoder on the current head.

A minimal catalogue containing an octal escape passed `i18n:validate` while Lingui decoded it into different runtime characters and lost the declared rich-text slot. Local lint, typecheck, 339 UI tests, catalogue checks, locale-verifier dry-run, and the UI production build all passed; current GitHub checks are also green.

PR #3521, opened 2026-08-17
feat: support Jellyfin Leaving Soon plugin for scheduled-deletion collections

ramonskie:
## Summary

Adds an alternative to native Jellyfin BoxSets for collections that
schedule deletion: the maintained [jellyfin-plugin-leaving-soon]
project mirrors them into a symlink-backed "Leaving Soon" library
instead. New Jellyfin setting `leaving_soon_method` (`collection` |
`plugin`, default `collection`), stored in the settings DB with a
migration.

## What changed

- **API**: new `GET /api/collections/leaving-soon` endpoint consumed by
  the plugin's provider (per-collection member media with precomputed
  `deletionDate`). Only served while the method is `plugin`, and only for
  active, deletion-scheduling collections with a non-DO_NOTHING action.
- **Gating**: in plugin mode, scheduled-deletion collections create no
  Jellyfin BoxSet; membership persists locally without a media-server add,
  the relink path is skipped, and add-rollback is guarded accordingly.
- **Reconcile**: switching the method removes existing BoxSets for these
  collections (plugin) or recreates them with current members (collection),
  run under the shared execution lock and never throwing.
- **UI**: "Leaving Soon collections" select in Settings -> Jellyfin,
  gated behind the new Jellyfin-only `SYMLINK_LIBRARY` feature flag.
- **Docs**: README section covering install, plugin config, and behavior.

## Testing

- 234 server specs + 3 UI specs, check-types, lint and format all green.
- Migration verified on a copy of an existing DB (schema rebuild, data
  preserved, new column defaulted to `collection`).

## Notes

The feature targets the Jellyfin-only leaving-soon plugin; other media
servers keep native BoxSet behavior unchanged. Marked as `feat:` (minor
release).


enoch85:
`Run integration tests` is red on this branch.

`apps/server/src/database/migration-tests/migrations.spec.ts:184`, "revert the newest migration cleanly (symmetric down)", hardcodes `tracearr_server_id` as the marker column for whatever the newest migration is. This PR adds a newer migration (`1786912383502-Add-leaving-soon-method-setting`), so `undoLastMigration()` correctly leaves `tracearr_server_id` in place and the assertion fails.

The fix is to point the marker at `leaving_soon_method`. That test needs updating whenever a migration is added.

enoch85:
Thanks for this, and for the care in it: a properly generated migration, real
test coverage, docs, translations, and best-effort error handling throughout.
The lock ordering and the rollback guard are clearly deliberate, not accidental.

I checked the branch out and ran it. Type checking is green and the collections
specs pass, so nothing below is "it doesn't work" - it's about where the cost
lands.

Reviewing this convinced me the underlying capability belongs in Maintainerr
properly, so I built it: **#3579 adds a per-collection "Keep in Maintainerr
only" option**, which is [feature request #11](https://features.maintainerr.info/posts/11/option-to-not-sync-collection-with-media-server)
(40 votes, open since May 2024). It's off by default, works on every media
server, and applies per collection rather than globally. Your PR is what made
me go looking, and the gating points it identified fed straight into it.

That changes what I need from this PR, and shrinks it a lot.

## 1. Drop the new endpoint

`getLeavingSoonCollections` (`collections.service.ts:1969`) is a near-copy of
`getCollectionsForOverlayData`, 42 lines above it in the same file. Both call
`findCollections()` and `getCollectionMediaByCollection()`.

`GET /api/collections/overlay-data` is already documented as "Get collections
with full media membership for overlay consumers ... for overlay and helper
integrations". Same `libraryId` / `typeId` params. It returns the whole
collection row (`deleteAfterDays`, `isActive`, `arrAction`, `libraryId`,
`type`) plus every member with `mediaServerId`, `addDate` and `tmdbId`.

That's a superset of the new payload. The only additions are three filters and
`deletionDate = addDate + deleteAfterDays`, the same arithmetic the overlay
processor already does internally.

The plugin can poll that endpoint today, against released Maintainerr, with no
server change at all. If server-side filtering is worth having, a query
parameter on `/overlay-data` is far better than a second method that has to be
kept in sync with the first one forever.

## 2. Drop the BoxSet gating

Once #3579 lands, the user ticks "Keep in Maintainerr only" on their
leaving-soon collections and no collection is created on Jellyfin. Same
outcome, with no `leaving_soon_method` setting, no `SYMLINK_LIBRARY` feature
flag, no reconcile handler, and no branch in the collection lifecycle tied to
one media server plus one plugin.

That last part was my real hesitation. Five conditionals in `collections.service`
plus a settings-change handler means every future change to collection handling
has to be reasoned about twice, and the second branch outlives the plugin.

## 3. The trap worth knowing about, whichever way this goes

You guarded the title-based relink on the add path with exactly the right
reasoning:

> In plugin mode a scheduled-deletion collection deliberately has no BoxSet;
> the title-based relink must not resurrect one that happened to keep the same
> name on the server.

`checkAutomaticMediaServerLink` does `if (foundColl) { collection.mediaServerId
= foundColl.id; save }`, and there are **four** call sites, not one:

| Site | Reached by |
| --- | --- |
| `collections.service.ts:3252` | the add path - guarded here |
| `collections.service.ts:3689` | `removeFromCollectionInternal`, hit by `collection-handler.ts:251` for every item the worker handles |
| `collections.service.ts:4060` | the delete path, which relinks and then deletes |
| `rule-executor.service.ts:805` | `getCollectionForMediaServerSync`, on every rule run |

So a same-titled BoxSet on Jellyfin would silently relink the collection and
revert it to BoxSet mode, and on the delete path it would delete a collection
that was never ours. Same class as #3344. #3579 honours the flag at all four,
and I tested it by planting a same-titled BoxSet and then running a full rule
execution plus a member removal against it: zero relink attempts, decoy
untouched.

The rest of my notes - the reconcile counting a failed `deleteCollection`
without clearing `mediaServerId`, the reconcile proceeding without the
execution lock after 30s, `SYMLINK_LIBRARY` reading as a server capability when
it's really "is this Jellyfin plus a plugin", and nothing verifying the plugin
is actually installed before the BoxSets are deleted - all disappear with the
gating, so I won't belabour them.

## What I'd take

1. Drop the endpoint; point the plugin at `/api/collections/overlay-data`.
2. Drop the gating; "Keep in Maintainerr only" covers it once #3579 merges.
3. Keep the README section, adjusted to describe the plugin plus that checkbox.

That leaves this PR as documentation plus, if you want it, a filter parameter
on the existing endpoint. Your plugin then works against released Maintainerr
instead of waiting on core, and it stops being coupled to a Jellyfin-only mode
flag - the same setup would work for a Plex or Emby equivalent later.

enoch85:
@ramonskie The PR is now merged to `development`.

If you think that's enough, then close this one. If not, please let me know! :+1:

enoch85:
@ramonskie Closing this as it's no longer relevant.

Please open a new issue/PR for the smaller change if needed.

ramonskie:
Do still want the readme?

enoch85:
> Do still want the readme?

Naah it's fine as is. 👍🏼

Issue #751, opened 2024-01-09
[Feature Request] Add season/episode rule value for media that's still being watched (previous seasons/episodes)


jorenn92:
Explore the possibiltiy to create a rule value that returns true when previous seasons are still being watched. 

Discord request https://discord.com/channels/1152219249549512724/1194061855426097314

Content: 

> Hoping to get some advice on creating a rule. 
I currently have a rule that deletes any TV show that hasn't been watched once in the past year. To make it work I had to add two sections to capture all cases.
It works well, but I have it running against "shows" and I'd like to run it against "seasons". The problem being when I use the same logic for seasons it selects seasons that I haven't got to watching yet, but I will once I've finished the previous ones. There's nothing wrong with this logic obviously, it's doing what it's told, but I'd like to exclude seasons that aren't the earliest season available.
For example if a show has 5 seasons and I only have seasons 3, 4 and 5 downloaded, and I'm halfway through 3, I want 4 and 5 to not be included in the rule. Once I've watched 3 and it gets deleted, 4 would then be included in the rule unless it gets watched.
Is this possible at all? If not it's no worry, just wanted to get a bit fancy.

Reply: 

>Hmm, I don't believe that's currently possible. However, I don't think the use case is too exotic; it seems logical to want to keep future seasons of a show that's still being watched.
I'll explore the possibility of adding a season/episode rule for this purpose!

jorenn92:
Perhaps just add a flag that returns true if a season is the first available season of a show? (or episode is part of the first available season)

appiekap653:
Should the already existing value: users that watch the show be empty for all the shows you want to remove?

So if User 1 is watching season 3, and season 4 and 5 are checked against the value: users that watch the show, it should give back User 1.

Hence checking for an empty users that watch the show should do the trick right?

Issue #3153, opened 2026-06-22
Possible inconsistency between rule execution and Test Media for latest aired season


dktrekkie:
**Describe the bug**
I am trying to create a season cleanup rule that keeps the latest aired season and adds only older seasons to a collection.

My understanding of:

`Sonarr - Is (part of) latest aired/airing season
`
is that it should return `true` for the latest season that has started airing, while ignoring seasons that are entirely in the future.

This understanding is based on the rule description and the previous changelog entry stating that future seasons should be excluded.

However, a normal rule run and Test Media return different values for the same season.

If I have misunderstood the intended use of this rule value, please let me know.

**To Reproduce**
My example is the series `Silo`:

Season 1 is an older season
Season 2 is the latest completed and aired season
Season 3 exists in Sonarr but is entirely in the future
Season 1 Episode 1 and all of Season 2 are present in Plex

I created a Rule Group with media type `Seasons` and Sonarr action `Do nothing`.

The rules are:

`Sonarr - [list] Tags (show)` contains `maintainerr-season-cleanup-ready`
AND `Sonarr - Is (part of) latest aired/airing season` equals `False`

The tag is added by a separate automation in Node-RED after verifying that the latest started season has finished airing. The automation itself is not involved in the rule evaluation. Any Sonarr show tag should produce the same result here, as the tag is only used to limit which shows the rule evaluates.

Steps:

1. Run the Rule Group normally.
2. Both `Silo - season 1` and `Silo - season 2` are added to the collection.
3. Open the stored rule result for Season 2.
4. It reports:

firstValueName: Sonarr - Is (part of) latest aired/airing season
firstValue: false
result: true

5. Run Test Media for the same Season 2 with mediaServerId: "60120".
6. Test Media reports:

firstValueName: Sonarr - Is (part of) latest aired/airing season
firstValue: true
result: false

The same season therefore receives:

Normal rule run: false
Test Media:       true

I also tried placing the two conditions in separate sections joined with `AND`. In that configuration, only Season 2 was added, which also appears to be the opposite of the expected result.

**Expected behavior**
Based on my understanding of the rule:

- Season 1 should return false because it is an older season
- Season 2 should return true because it is the latest aired season
- Future Season 3 should not cause Season 2 to be treated as an older season
- Only Season 1 should be added to the collection
- Test Media and the normal rule runner should return the same value

**Screenshots**

_The Rule Group configuration_
[...]

PR #3531, opened 2026-08-18
feat(rules): add episode view date through the target season

whitestrake:
### Description & Design

This adds a season-only date rule value for Plex, Jellyfin and Emby:

`Newest episode view date in this or an earlier season`

This is intended for season cleanup rules that need to protect seasons ahead of somebody who is still progressing through a show. For example, when a viewer is active in Season 3, a rule evaluating Seasons 4 and 5 can see that recent activity rather than treating those unwatched future seasons as abandoned.

The value is a date rather than a boolean so rule authors can choose their own definition of "recently watched" using the existing date comparisons.

For each target season, the provider getter:

- Considers episode views from the target season or an earlier regular season.
- Returns the newest qualifying view date, or `null` if nothing qualifies.
- Treats Season 0 separately: specials only consider other specials, while regular seasons exclude specials.
- Skips unnumbered sibling seasons.
- Treats malformed or incomplete provider data as a transient failure rather than silently reporting the season as unwatched.

Plex uses its show-level watch history, which can include retained history for episodes since removed from the library. Jellyfin and Emby derive the value from current episode children, so views of removed episodes do not count.

The implementation adds no dependencies, database migration or UI-specific code. It follows the existing rule-property and getter patterns for each provider.

One semantic point I would appreciate maintainer input on: the current implementation does not treat specials as an earlier season for Season 1 and above. My reasoning was that Season 0 is not necessarily part of normal viewing progression, but this is intentionally isolated and straightforward to change if including specials is preferred.

### Related issue

Related to #751.

### AI-Assisted Development

Codex assisted with repository analysis, implementation, tests, focused review passes and local verification.

I reviewed the resulting diff and the underlying design decisions, including provider response semantics, season boundaries, malformed-data handling and specials behaviour. The implementation uses the existing rule constants and provider getter structures rather than introducing another abstraction or dependency.

### Checklist

### How to test

1. Run the focused getter suites:

   ```console
   yarn workspace @maintainerr/server jest src/modules/rules/getter/plex-getter.service.spec.ts src/modules/rules/getter/jellyfin-getter.service.spec.ts src/modules/rules/getter/emby-getter.service.spec.ts --runInBand
   ```

2. With `yarn dev` stopped, start the matching fake media server in a separate terminal:

   ```console
   node tools/dev/fake-plex.mjs
   node tools/dev/fake-jellyfin.mjs
   node tools/dev/fake-emby.mjs
   ```

3. Run the matching seed command, then start Maintainerr:

   ```console
   MEDIA_SERVER=plex node tools/dev/seed-db.mjs
   MEDIA_SERVER=jellyfin node tools/dev/seed-db.mjs
   MEDIA_SERVER=emby node tools/dev/seed-db.mjs
   yarn dev
   ```

4. Open the seeded `Season-prefix View History` collection and use **Test Media** against its seasons. Confirm that:

   - Season 0 considers only specials.
   - Regular seasons consider their own and earlier regular seasons.
   - Activity in a later season does not affect an earlier target season.

5. Run the project checks:

   ```console
   yarn format:check
   yarn check-types
   yarn lint
   yarn test
   yarn build
   ```

### Additional context

The focused Plex, Jellyfin and Emby getter suites pass all 184 tests. The complete server suite passes all 2,954 tests. Typechecking for the server, UI and contracts packages, targeted lint and formatting, fake-server syntax checks and diff checks also pass.

The final pushed commit was additionally validated end to end with Playwright against disposable real Plex, Jellyfin and Emby servers. Seasons 1, 2 and 3 were tested for each provider, for nine cases in total. Every request returned HTTP 201 and code 1 with the selected season ID, correct provider rule name, valid frontier date, true result and matching rendered **Test Media** output. Season 2 correctly ignored newer Season 3 activity. No browser console errors, page errors or Maintainerr error-level log entries were observed.

Season 0 was covered by focused tests and fake-server fixtures; the disposable real-server corpus contained no specials.

I also intend to submit a separate follow-up draft PR containing a pair of rules that aggregate users who watched any or all episodes after the media was added. This will exclude older watch history from previous times the media was present and later removed. It is being kept separate so that it can be discussed on its own merits, independently of this addition.


enoch85:
Thanks for the PR!

Can you make this work for Emby and Jellyfin as well? You need to setup a real test environment, and run real API calls, and do Playwright testing against all three media servers.

AI must self review critically, and when you're done land it here for discussion (if any). :+1:

whitestrake:
Sure, I'll see what I can do. I don't have much expertise with Jellyfin or Emby but I'm sure they can't be too hard to include, hopefully the hardest part is actually standing up disposable servers in the first place.

I don't think I've done Playwright testing before. What kind of tests would you like to see? I will figure out how to get it for you. I'm assuming you just want evidence rather than anything committed within the PR itself.

Having a quick look over the fake-servers in the repo, I think I will need to extend the Jellyfin/Emby ones a bit as well in order to test against properly? The mock Plex already had all the season/episode/watch history fixtures we needed - I just added some entries to the history data is all - but the others don't.

Will get onto this when I can.

enoch85:
Great! Playwright testing is a way for AI to check the actual result as an end user. Everything is in the repo rules and guidelines, so if you follow that everything should be fine. 👍🏼

How tests should look like are also described. But basically, we already have too much tests, so don't add unnecessary stuff for debugging och just to prove a point. I've seen AI made the most stupid tests like if x = x then pass!

Yeah, the fake servers was added before I had a real setup (complete Docker stack with all services and media servers) so they might be incomplete. They are better than having nothing at all, but real servers always wins. Just fire up some containers, should be fairly quick, and AI can do that as well.

Looking forward to your commit! 💪🏼

whitestrake:
Done. I've now extended `Newest episode view date in this or an earlier season` to Jellyfin and Emby.

The behaviour is consistent across all three providers:

- A regular target season considers episode views from Season 1 through the target season, excluding specials and later seasons.
- Season 0 considers specials only.
- Missing history returns no date, while failed or incomplete lookups remain errors rather than being mistaken for empty history.

I've added focused adapter and getter tests, along with Jellyfin and Emby mock-server fixtures and seeded rule coverage.

I also tested the complete flow through Playwright using fresh Maintainerr instances connected to real disposable Plex, Jellyfin and Emby servers. For each provider I configured the server, created and ran a season-level rule using the new property, and checked the results through Test Media.

| Provider | Season 1 | Season 2 | Season 3 |
| --- | --- | --- | --- |
| Plex | `2026-08-28T08:10:11.000Z` | `2026-08-28T08:10:13.000Z` | `2026-08-28T08:10:15.000Z` |
| Jellyfin | `2026-08-18T01:00:00.000Z` | `2026-08-24T02:00:00.000Z` | `2026-08-27T03:00:00.000Z` |
| Emby | `2026-08-18T01:00:00.000Z` | `2026-08-24T02:00:00.000Z` | `2026-08-27T03:00:00.000Z` |

In particular, the Season 2 results ignored the newer Season 3 activity, which exercises the target-season boundary rather than merely confirming that some history exists.

Formatting, lint, typechecking, the full uncached monorepo test suite, the production build, and the server e2e matrix all pass. The source patch used for the disposable test image also matched the final local diff exactly.

enoch85 (commented):
Claude Opus 4.8:

<details>
<summary>Review of #3531 at 8a3b4166 (expand)</summary>

The Plex half of this is good work: the property is correctly gated (`showType: ['season']`), the getter reads the show rollup once, the `null` vs `undefined` contract is right, and it produces the intended answer against a real Plex server. I ran it end to end on the dev stack and the feature does what the description says on all three servers.

What I cannot sign off on is the Jellyfin/Emby half and the adapter hardening that came with it. Two of those are reproducible defects on the dev servers as they are configured today, one reverses a mitigation added in #3346, and the rest is scope that does not belong in a PR titled "add Plex view date through the target season".

## Blocking

| # | Where | Defect |
|---|---|---|
| 1 | `jellyfin-getter.service.ts:328`, `emby-getter.service.ts:585` | One season with no index number makes **every** season of that show permanently unreadable |
| 2 | `emby-adapter.service.ts:617-624` | Emby's `Limit: 500` child page now throws, and non-strict callers turn it into `[]` - existing rules answer 0 |
| 3 | `jellyfin-adapter.service.ts:1422` | Reverses the #3346 cross-page dedupe: a repeated row now discards the whole watch snapshot |

### 1. Unnumbered sibling season

Both getters validate the index of **every** season of the show before filtering to the qualifying ones, and throw on a bad one. Jellyfin and Emby both model episodes they cannot place as a season with `IndexNumber: null`, which `JellyfinMapper`/`EmbyMapper` map to `index: undefined` (`jellyfin.mapper.ts:205`, `emby.mapper.ts:160`). That is normal data, not malformed data.

`dev-jellyfin` already has one. `GET /Shows/dc8cbb6bc7f984c2ef9597069da2bca4/Seasons` on the "English Premier League" series returns:

```
b15d2d15ded3e981294a2c2686e57507  IndexNumber 2024  Season 2024
5cdbf4a480a0edd75294a36fc3cdc96c  IndexNumber 2025  Season 2025
222f9b3d37c429e4958958570598edad  IndexNumber null  Season Unknown
```

`POST /api/rules/test` against a season-scoped group holding only property 48, on that real server:

| Season | Result |
|---|---|
| Season 2024 | `could not be read for this item` |
| Season 2025 | `could not be read for this item` |
| Season Unknown | `could not be read for this item` |
| Blender Open Movies S1 (no unnumbered sibling) | `2026-08-12T19:52:54.274Z` |
| Sherlock S2 (numbered, no episodes) | `null` |

Server log for all three: `[WARN] [JellyfinGetterService] Jellyfin-Getter - Action failed for 'Season 2024' ...`. `undefined` is the transient signal, so the seasons land in `transientFailureMediaIds` and the rule can never act on that show - silently, forever, with no user-visible cause.

Unit repro, added to the PR's own suite and failing on this branch (Emby is identical with `mockShow`):

```ts
it('REPRO: unnumbered sibling season', async () => {
  const seasons = [
    createMediaItem({ id: 'season-2024', type: 'season', index: 2024 }),
    createMediaItem({ id: 'season-2025', type: 'season', index: 2025 }),
    createMediaItem({ id: 'season-unknown', type: 'season', index: undefined }),
  ];
  // ... history: season-2024-episode watched 2026-06-01
  await expect(getFrontierDate(2025)).resolves.toEqual(new Date('2026-06-01'));
});
// Expected: 2026-06-01T00:00:00.000Z
// Received: undefined
```

The neighbouring getter in the same file already has the right answer. `getNewestWatchedEpisodeDate` at `jellyfin-getter.service.ts:770` does `if (episodeOrder === undefined || episode.parentIndex === undefined) continue;`. An unnumbered season is not part of any numbered progression, so skip it - and prefer keying the qualifies test off `episode.parentIndex`, the way `sw_lastWatched` already does, which removes the season-index validation entirely.

Note the existing test `returns undefined for an invalid returned season index` only covers the case where the **target** season is unnumbered. It never covers a numbered target with an unnumbered sibling, which is the shape that actually occurs.

### 2. Emby child-page guard turns truncation into zero

`getChildrenMetadata` sends `Limit: EMBY_BATCH_SIZE.MAX_PAGE_SIZE` (500) on the `/Items` branch and does not page. Emby's `TotalRecordCount` is the full match count, independent of `Limit` - confirmed against the real server:

```
GET /Items?ParentId=5&IncludeItemTypes=Episode&Recursive=true&Limit=2
-> TotalRecordCount=5, Items.length=2
```

So the new `Items.length !== TotalRecordCount` check fires for any parent with more than 500 children. The `throwOnError = false` callers - which is most of them - then swallow it in the catch and return `[]`, which is a worse answer than the truncated list it replaces.

Reproduced on the real Emby server by setting `MAX_PAGE_SIZE: 2` against a 4-episode season, on the pre-existing rule "Amount of available episodes" (property 14), which never touches this PR's new code:

| Branch | `sw_episodes` for Emby season 1991 (4 episodes) |
|---|---|
| `development` | `2` (truncated, non-zero) |
| this PR | `0` |

A rule such as "amount of available episodes is less than 1" flips from not matching to matching. `getChildrenMetadata` is also the shared path for the overlay processor (`overlay-processor.service.ts:160`), collection size roll-up (`collections.service.ts:4927`) and the Tracearr resolver (`tracearr-api.service.ts:815`), so the blast radius is not this rule.

Silent truncation at 500 is a genuine pre-existing bug, but the fix is to page the request (`StartIndex`, the way `sweepUserItems` does), not to convert it into a throw. The Jellyfin equivalent sends no limit, so its guard is inert - confirmed live, `TotalRecordCount` equals `Items.length` there both with and without `excludeLocationTypes=Virtual`.

### 3. Jellyfin snapshot sweep

`jellyfin-adapter.service.ts:1422` throws when `seenThisUser.size !== total`. `seenThisUser` exists precisely to absorb a row repeated across pages - the comment three lines above still says so, and #3346 added it because "counting it twice would inflate playCount and duplicate watch records". The PR deletes the test that pinned that behaviour:

```
-    it('counts a row repeated across pages only once', ...)
-      expect(cached.watchHistory.get('ep-1')).toHaveLength(1);
-      expect(cached.playCount.get('ep-1')).toBe(2);
+    it('abandons a snapshot when duplicate rows hide a missing item', ...)
+      expect(snapshot()).toBeUndefined();
```

So the mitigation is now the trigger: a library changing under a multi-page sweep discards the entire prefetch and puts every watch read back per item for the whole run. `total` is also reassigned on every page (`sweepUserItems` line 1534), so the equality is compared against the **last** page's count, not the first - it is racy in both directions. Static libraries page cleanly (verified: 36 of 36 unique at `limit=1` on `dev-jellyfin`), so this only bites the large installs the prefetch exists for.

## Scope

The PR is titled and described as a Plex season date rule. It also changes shared adapter behaviour that every existing rule, overlay and collection read goes through:

| Change | Affects |
|---|---|
| `getChildrenMetadata` array/count guards (Emby + Jellyfin) | every show/season getter, overlays, collection sizing, Tracearr |
| `getWatchHistory` empty-user-list and per-item shape throws (Emby) | every Emby watch read |
| `getDescendantEpisodeWatchHistory` empty-user-list throw (Jellyfin) | every Jellyfin show/season watch read |
| `sweepUserItems` returning `total` plus the caller's equality check | the watch-state prefetch |

None of it is needed to compute a season view date, and finding 3 shows it is not a free hardening pass. Please split it into its own PR so it can be argued on its own evidence, and keep this one to the rule.

## Tightness

Three implementations of one property, none of which reuse what is already there:

| Server | Shape | Cost for one target season |
|---|---|---|
| Plex | one show history read, one pass | 1 request |
| Jellyfin | descendant history + season children + episode children per qualifying season | 1 + 1 + N(qualifying) |
| Emby | bespoke cached "frontier" over **all** seasons, one `getWatchHistory` per episode | 1 + N(all seasons) + episodes x users, uncached |

- `getCompletedSeasonFrontier` ignores the target: evaluating S1 of a ten-season show reads every episode of all ten. Jellyfin's version already skips non-qualifying seasons - do the same. The name is also misleading, nothing about it is "completed".
- Its inner loop is `getLastViewedAt` (`emby-getter.service.ts:547`) re-implemented, and the outer walk duplicates `getLastWatchedShowDate` (`:768`). The Jellyfin block re-implements `newestWatchedAt` (`:641`) inline.
- The hand-rolled `completedSeasonFrontierBuilds` in-flight map is a fourth cache layer in the getter. `this.cache.data.set(cacheKey, frontier)` is also the only `set` in that file with no explicit TTL - the sibling at `:974` passes `600`.
- Jellyfin pre-scans the **entire** descendant history for invalid dates (`:309-317`) before filtering to qualifying seasons. Fold the check into the loop that already visits the records it cares about, or drop it.
- Plex defensive clutter that cannot fire: `showRatingKey !== showRatingKey.trim()`, `Math.abs(viewedAtMs) > 8.64e15`, and `view.type !== 'episode'` - `buildWatchHistorySnapshot` already filters non-episode rows out of the show rollup (`plex-api.service.ts:1032`) and the live `metadataItemID` fallback returns leaf rows only (verified against real Plex).
- 685 lines of new spec for 241 lines of getter. The Plex block alone has ten near-identical `returns undefined when ...` permutations plus a 150,000-row "without overflowing the call stack" test that guards a spread/`Math.max` implementation this PR does not have. The behaviour worth pinning is: qualifies, specials isolation, `null` when nothing qualifies, and one transient case.

## Smaller

- `seed-db.mjs` rewrites the #3153 repro group rather than adding its own: property 48 is prepended as the first rule and the Sonarr rule's `operator` flips `null -> 0`. That group exists to isolate #3153; hosting a second fixture in it means a future regression there is ambiguous. The `if (TARGET === "plex")` also becomes a bare `{ }` block - give it a comment or drop the braces.
- No comment records that the same property means different things per server: Plex counts history rows, so a view of a since-deleted episode still counts; Jellyfin and Emby walk current children and only count completed watches. Standing rule 10 asks for exactly that to be documented in code.
- On the open question in the description: keeping specials out of the progression for S1+ is the right call, and Season 0 seeing only Season 0 is consistent with it.

## Verified

Real dev stack, no mocks: `dev-plex` (PMS, Sports library, seasons indexed by year), `dev-jellyfin` 3 users, `dev-emby`.

| Check | Result |
|---|---|
| Plex: S2024 watched via `/:/timeline`, S2025 evaluated | S2025 inherits `2026-08-28T17:37:08Z` |
| Plex: S2025 watched later, both re-evaluated | S2024 `17:37:08`, S2025 `17:48:19` - earlier season correctly does not see the later view |
| Plex: full `POST /api/rules/425/execute` with the library prefetch | rollup built ("2 entries - 2 items across 1 shows"), exactly seasons 2577 and 2582 collected, F1/NBA excluded |
| Plex: show with no history | `null`, `is not recorded for this item` |
| Emby: S2024 marked played, S2025 evaluated | S2025 inherits `2026-06-01T12:00:00Z`; after S2025 watched `2026-07-01`, S2024 still answers `2026-06-01` |
| Jellyfin: numbered-only show | `2026-08-12T19:52:54.274Z`; empty season `null` |
| Jellyfin: show with an unnumbered season | all seasons unreadable (finding 1) |
| Emby: child page shorter than `TotalRecordCount` | existing property 14 answers 0, was 2 (finding 2) |
| UI (Playwright): property visibility by data type | Shows `[]`, Episodes `[]`, Seasons `Emby - Newest episode view date in this or an earlier season` |
| `yarn format:check`, `yarn check-types` | pass |
| `jest src/modules/rules/getter`, `jest src/modules/api/media-server` | 221 + 632 pass |

Plex history rows and Emby played flags created for this were removed afterwards; the dev DB was restored from backup.

</details>

whitestrake:
Thanks for the detailed review. The unnumbered-season and Emby child-page issues look valid, and I’m happy to iterate on those if they’re still within the intended scope.

I’m not sure I understand the scope feedback, though. You asked me to make the rule work for Emby and Jellyfin and to test it against all three media servers. The review now says those implementations don’t belong in this PR. Would you like me to update the PR title and body, or is the scope being reduced back to Plex only, with the Jellyfin and Emby work removed?

I’m happy to follow whatever direction you intend, but I don’t want to infer a scope reduction from an automated review when that conflicts with your earlier request.

I agree that the shared adapter hardening expanded the scope unnecessarily, so I can remove those changes from this PR.

enoch85:
> I’m not sure I understand the scope feedback, 

That's Claude being Claude. 🙂 Let's do everything in this PR.

> I agree that the shared adapter hardening expanded the scope unnecessarily, so I can remove those changes from this PR.

Yes please. 👍🏼

whitestrake:
I've gone through each point and tightened the PR back to the rule itself, the supporting machinery it requires, and the focused testing needed to prove it.

<details>
<summary><strong>Responses to individual review items</strong></summary>

| Reviewer point | Response |
| --- | --- |
| The Plex implementation, season-only gating and `null`/`undefined` behaviour were correct. | Agreed. Those parts remain unchanged apart from removing unrelated defensive checks noted later in the review. |
| An unnumbered Jellyfin or Emby sibling season made every numbered season unreadable. | Agreed and fixed. Unnumbered sibling seasons are now skipped before the numbered progression is evaluated. Both providers have a focused regression and a `Season Unknown` fake-server fixture. |
| The Emby child-count guard converted a truncated result into `[]`, changing existing rules from a partial count to zero. | Agreed. The guard and its adapter tests were removed. The final PR has no Emby adapter diff against `development`. |
| The Jellyfin unique-count check reversed the cross-page deduplication behaviour from #3346. | Agreed. The equality check and its replacement test were removed. The final PR has no Jellyfin adapter diff against `development`. |
| Jellyfin and Emby did not belong in a Plex-titled PR. | The maintainer clarified that all three providers should remain in this PR. The Plex, Jellyfin and Emby rule implementations therefore remain together. |
| Shared Emby and Jellyfin `getChildrenMetadata` hardening affected every getter, overlays, collection sizing and Tracearr. | Agreed that this expanded the scope unnecessarily. All shared child-list hardening was removed. |
| Emby `getWatchHistory` empty-user and response-shape hardening affected every Emby watch read. | Agreed that this was shared behaviour outside this rule. It was removed. |
| Jellyfin descendant-history empty-user hardening affected every show and season watch read. | Agreed that this was shared behaviour outside this rule. It was removed. |
| Returning `total` from `sweepUserItems` and comparing it with unique rows changed the shared watch-state prefetch. | Agreed that this was shared behaviour outside this rule. It was removed. |
| Emby built a complete frontier across all seasons even when evaluating an early target season, and the helper name was misleading. | Agreed. The frontier helper was removed. Emby now filters to qualifying seasons before reading episode history. |
| The Emby implementation duplicated `getLastViewedAt` and `getLastWatchedShowDate`, while Jellyfin duplicated `newestWatchedAt`. | Agreed. Emby now reuses `getLastWatchedShowDate`, and Jellyfin reuses `newestWatchedAt`. |
| The Emby in-flight frontier map added another cache layer and stored results without an explicit TTL. | Agreed. The frontier cache and in-flight map were removed entirely. |
| Jellyfin scanned all descendant history for invalid dates before filtering to qualifying seasons. | Agreed. Date handling now occurs only while visiting episodes in qualifying seasons. |
| Plex contained unnecessary whitespace, media-type and absolute date-range checks. | Removed. I do not necessarily agree that every defensive check was intrinsically worthless, but none was required for this rule and the Plex show-history rollup already constrains the row type. The required season-index and finite-date validation remains. |
| The Jellyfin and Emby request cost was much higher than Plex. | The provider APIs require different access patterns, but both implementations now read only qualifying seasons. The speculative Emby frontier cache was not retained without a demonstrated need. |
| The new specs were too large and contained many near-identical malformed-input permutations. | Agreed with the scope concern. The suite was reduced to four focused behaviours per provider: regular-season progression, Season 0 isolation, confirmed-empty history and one transient case. |
| The 150,000-row Plex test was unnecessary and guarded an implementation the rule did not use. | I do not specifically agree that the test was meaningless, because it exercised a real large-history and call-stack characteristic. I removed it anyway because that scale proof was not required to establish this rule's behaviour and was outside the minimum requested test scope. |
| `seed-db.mjs` modified the existing #3153 fixture and changed its operator. | Agreed and fixed. The #3153 fixture is restored unchanged, and property 48 now has its own isolated seeded collection and rule group. |
| The replacement bare seed block lacked an explanation. | Fixed by giving the new property 48 seed fixture its own named and documented block. |
| The provider-specific history semantics were undocumented. | Agreed and fixed. Code comments now record that Plex can count retained history for deleted episodes, while Jellyfin and Emby derive the value from current episode children. |
| Specials should remain outside the regular progression, with Season 0 considering only Season 0. | Agreed and retained across all three providers, with focused automated coverage. |

</details>

## Verification

- The three focused getter suites pass: 184 tests across Plex, Jellyfin and Emby.
- The complete server suite passes: 2,954 tests.
- Typechecking passes for the server, UI and contracts packages.
- Targeted lint, formatting, fake-server syntax checks and `git diff --check` pass.
- The fake Plex, Jellyfin and Emby probes and the provider-specific seed variants were exercised successfully.
- I rebuilt the disposable Maintainerr test instances from the final pushed commit. Playwright then exercised Seasons 1, 2 and 3 against real Plex, Jellyfin and Emby servers, for nine end-to-end cases.
- Every request returned HTTP 201 and code 1 with the selected season ID, correct provider rule name, valid frontier date, true result and matching rendered Test Media output. Season 2 correctly ignored newer Season 3 activity.
- There were no browser console errors, page errors or Maintainerr error-level log entries during that run.
- Season 0 was covered by focused tests and fake-server fixtures; the disposable real-server corpus contained no specials.

enoch85 (commented):
Claude Opus 4.8:

<details>
<summary>Review of #3531 at d9580489 (expand)</summary>

All three blockers from the previous round are fixed, and I confirmed each against the real dev stack rather than the tests. The adapter hardening is gone, the diff is back to 7 source files and 4 dev fixtures, and the getters now reuse what was already in each file. This is close.

What is left is one should-fix that this PR did not introduce but does inherit, one measured scalability cost that is Emby-only, and a consistency question on the Plex loop.

## Previous round, re-verified

| Finding | Fix | Evidence |
|---|---|---|
| Unnumbered sibling season killed every season of the show | Both getters `continue` past a season with no index | `dev-jellyfin` "English Premier League" (which has a real `IndexNumber: null` season): S2024 `2026-08-12T21:52:53Z`, S2025 same, where all three were `could not be read` before |
| Emby `Items.length !== TotalRecordCount` guard | Reverted | `sw_episodes` on Emby season 1991 answers `4`, was `0` |
| Jellyfin sweep `seenThisUser.size !== total` | Reverted, dedupe test restored | adapter files no longer in the diff |
| Scope | Adapters and UI untouched | `git diff --name-only` is 7 getter/constant files plus `tools/dev/` |
| Tightness | Emby scopes to qualifying seasons and reuses `getLastWatchedShowDate`/`getLastViewedAt`; Jellyfin reuses `newestWatchedAt`; Plex clutter (`trim()`, `8.64e15`, `type !== 'episode'`) gone; specs 685 -> 372 lines | |
| `seed-db.mjs` rewrote the #3153 repro | Repro restored untouched, own `Season-prefix View History` fixture added | |
| Per-server semantics undocumented | Comment on both constants entries | |

The fixtures now carry a `Season Unknown` season in `fake-jellyfin.mjs` and `fake-emby.mjs` and both getter specs assert it is skipped. Good.

## Should fix

### Emby answers "never watched" when it cannot read; Plex and Jellyfin answer "unknown"

I put a fault-injecting proxy in front of each real server and failed the watch read while evaluating one season:

| Server | Injected failure | Getter answer |
|---|---|---|
| Plex | 500 on `/status/sessions/history/all` | `could not be read for this item` (transient) |
| Jellyfin | 500 on the per-user recursive episode sweep | `could not be read for this item` (transient) |
| Emby | 500 on `/Users/{u}/Items/2020` | `is not recorded for this item` (**confirmed null**) |

Same input, same property, opposite contract. The Emby result is a false "never watched": with a `NOT_EXISTS` or `BEFORE` rule the season stops being protected and gets collected for deletion, which is the exact outcome this property exists to prevent.

The cause is `emby-adapter.service.ts:940` - `catch { // Some users may not have access to this item - skip silently. }` - so when every user's read fails, `getWatchHistory` returns `[]` and `getLastViewedAt` returns `null`. That is pre-existing on `development` and it is part of the adapter work I asked to split out last round, so it is not a regression here. It matters more for this property than for `lastViewedAt` because this one is a deletion guard. Either land that split-out adapter fix before or with this, or have the getter treat "no user could read this episode" as transient.

Plex and Jellyfin need no change - both were verified fail-closed above.

## Measured

### Emby repeats a full season walk per qualifying season

Counting proxy in front of each real server, one `POST /api/rules/test` per cell, English Premier League (2 seasons, 4 episodes each, 1 Emby user / 3 Jellyfin users):

| Property | target = first season | target = second season |
|---|---|---|
| Emby `lastViewedAt` (id 7) | 10 | 10 |
| Emby `sw_lastViewedAtThroughSeason` | 11 | **20** |
| Jellyfin `lastViewedAt` (id 7) | 4 | 4 |
| Jellyfin `sw_lastViewedAtThroughSeason` | 8 | 7 |

Jellyfin is flat because `descendantWatchHistory` is one call and the season children are cached. Emby calls `getLastWatchedShowDate(season.id, 'season', true)` once per qualifying season and nothing caches `getWatchHistory`, so a season-scoped group over an S-season show performs `S(S+1)/2` season walks where every other Emby season getter performs `S`. For a 10-season show at 22 episodes and 5 users that is roughly 7,300 requests per rule per run against about 1,330 for `lastViewedAt`.

Two contributors worth separating:

- The repeat itself is this PR's. A run-scoped memo on the per-season result collapses it back to `S`; the value is identical for every later target that includes that season. The removed frontier cache was doing this, so it does not need reinventing - just key it per season rather than per show.
- The per-episode `/Users/Query` is not. `getWatchHistory` calls `fetchUsersQuery` directly (`emby-adapter.service.ts:915`) instead of the cached `getUsers`, so the user list is refetched once per episode - measured 1:1 with episode reads. Pre-existing, but this property multiplies it.

## Consistency

### Plex still fails the whole read on a row it cannot place

`plex-getter.service.ts:456-462` throws for any history row without a usable `parentIndex`, including rows for seasons that would not qualify, so one such row anywhere in the show's history makes the target season transient. The Jellyfin and Emby siblings now `continue` past a season they cannot place. Three implementations of one property should not disagree about that.

It matters more here than it looks, because the Plex behaviour this PR advertises is that retained history for removed episodes counts, and a retained row is the one most likely to have lost its parent metadata. I could not produce such a row on the dev stack - removing library media there deletes the file - so I am not calling this confirmed, only unsettled. Worth confirming against a server that has one, or just skipping the row the way the other two now skip a season.

## Trim

- `emby-getter.service.ts:743` - the `show` branch of `getLastWatchedShowDate` plumbs `throwOnError` into both `getChildrenMetadata` calls but omits the invalid-date guard the `season` branch got at `:736`. Nothing calls it with `throwOnError` for a show, so it is dead and inconsistent at the same time. Either add the guard or keep the flag on the season branch.
- The flag now controls two unrelated things (children strictness and date validation), so `throwOnError` no longer describes it.
- `jellyfin-getter.service.ts:304` fetches the descendant history before it knows any season qualifies. Harmless, but the season read is the cheaper gate.
- `tools/dev/fake-jellyfin.mjs` keeps two representations of the same query param: the raw `itemTypes` at `:386` and the parsed `requestedItemTypes` at `:387`. The old exact-match comparisons at `:435` and `:438` (`itemTypes === 'Movie'`, `=== 'Series'`) should read from the Set, otherwise the next multi-type request silently misses them.

## Verified

Real dev stack, no mocks or fakes. Each server was made active in turn and driven through `POST /api/rules/test`, plus a counting/fault-injecting HTTP proxy in front of each.

| Check | Result |
|---|---|
| Plex: S2024 watched via `/:/timeline`, then S2025 watched later | S2024 `15:46:17Z`, S2025 `15:46:19Z` - earlier season does not see the later view |
| Plex: a third view on S2024 makes it newest | S2024 `15:47:05Z`, S2025 inherits `15:47:05Z` |
| Plex: `POST /api/rules/425/execute` | rollup built ("3 entries - 3 items across 1 shows"), collection holds exactly 2577 and 2582; F1 and NBA excluded |
| Plex: shows with no history | `null`, `is not recorded for this item` |
| Jellyfin: EPL, which has a real `Season Unknown` sibling | S2024 and S2025 both resolve; only the unnumbered season itself is transient |
| Jellyfin: S2024 episode set to `2026-09-01`, S2025 re-evaluated | S2025 inherits `2026-09-01T14:00:00Z` |
| Emby: S2024 `2026-06-01`, S2025 `2026-04-01` | S2025 inherits `2026-06-01`; after S2025 moves to `2026-07-01`, S2024 still answers `2026-06-01` |
| Emby: existing `sw_episodes` after the adapter revert | `4`, was `0` on the previous head |
| Fault injection on all three | table above |
| Request counts on Emby and Jellyfin | table above |
| UI (Playwright): property visibility | Shows `[]`, Episodes `[]`, Seasons `Plex - Newest episode view date in this or an earlier season` |
| `yarn format:check`, `yarn check-types`, `yarn lint` | pass |
| `jest src/modules/rules src/modules/api/media-server` | 53 suites, 1498 tests pass |

Test residue removed afterwards: Plex history rows deleted and unscrobbled, Emby played flags cleared, the Jellyfin episode's `LastPlayedDate` restored (its `PlayCount` is 3 higher, which is dev-only noise). Dev DB restored from backup.

</details>

whitestrake:
I've gone through the remaining review points. The latest changes are confined purely to the rule and its focused tests. Some of the findings were deliberately not addressed as they're outside the agreed scope of this specific PR.

<details>
<summary>Responses to individual review items</summary>

| Reviewer point | Response |
| --- | --- |
| Emby can report `null` when every per-user watch read fails. | Not addressed in this PR. This is pre-existing shared adapter behaviour affecting every Emby watch getter. Distinguishing a genuine empty history from swallowed read failures requires changing that adapter contract, which is outside the agreed rule-and-focused-tests scope. |
| Emby repeats the same season walks for later target seasons. | Agreed and fixed. The rule now memoises the show's season list and each qualifying season's computed date through the existing run-scoped `ArrLookupCache`, including concurrent callers. A focused regression covers the repeated-walk case. |
| Emby refetches `/Users/Query` once per episode. | Agreed that this is inefficient, but it is pre-existing adapter behaviour rather than something introduced by this rule. It remains untouched for separate adapter work. |
| Plex fails the read when a retained history row has no usable `parentIndex`. | Kept intentionally. Skipping an unplaceable retained row could hide real watch evidence and under-report the frontier. Returning transient `undefined` keeps this deletion-oriented rule fail-closed. |
| The Emby show branch forwarded `throwOnError` without applying all of its strict checks, and the flag now covers more than one check. | The unreachable strict arguments were removed from the show branch. The existing private flag remains on the only reachable strict path, the new season rule; renaming or splitting it would add no behavioural value here. |
| Jellyfin reads descendant history before checking whether any season qualifies. | No change. With a normal Jellyfin response the target season itself qualifies, so both reads are required regardless of order - making this a a minor edge-case micro-optimisation rather than a correctness issue. |
| The fake Jellyfin server keeps both raw and parsed forms of `includeItemTypes`. | No change. That is generic fixture cleanup outside the rule's request paths and is not required by its focused tests. |
| A rejected memoised Emby read remains cached for the rest of that evaluation run. | Left unchanged in this PR. The run remains fail-closed and the next run receives a fresh cache. Changing within-run retry policy would expand the scope beyond this rule. |
| A raw or imported rule could apply the season property to a non-season item and receive `null`. | Agreed and fixed for Plex, Jellyfin and Emby. All three now treat invalid rule scope as transient `undefined`, with one focused regression per provider. |
| Plex accepted negative timestamps and finite timestamps outside JavaScript's valid `Date` range. | Agreed and fixed. The getter now rejects both cases, covered by one compact table-driven focused test. |
| Completed records without a usable date return `null`, while unnumbered seasons are skipped. | Kept intentionally. A newest-view-date property has no recorded value when no date exists, and an unnumbered season cannot be placed in an ordered season prefix. |
| Jellyfin and Emby accepted qualifying season or episode records with blank IDs. | Agreed and fixed. Both providers now fail closed before issuing an unscoped child lookup or silently losing a history match, with compact table-driven coverage for the season and episode cases. |

</details>

Everything was re-tested as before, including a fresh Playwright end-to-end run against the disposable Plex, Jellyfin and Emby stack.

enoch85 (dismissed):
Claude Opus 4.8:

Approved. I pushed `ef098c1b` to this branch with four fixes; the feature itself is sound and does what #751 asked for.

## Verified against real servers

Built the rule the issue describes (season group, `NOT_EXISTS OR BEFORE 30d`) on real Plex and Emby, with an in-progress season and a downloaded-but-untouched later season:

| | existing `lastViewedAt` on the untouched season | this property |
|---|---|---|
| Emby EPL S2025 | `null`, so `NOT_EXISTS` matches and it is collected | inherits S2024's date, protected |
| Plex F1 S2026 | `null`, so `NOT_EXISTS` matches and it is collected | inherits S2025's date, protected |

Full sweep over a six-season library collected the four never-watched seasons and protected both seasons of the show in progress. Moving the watch back a year released both. Specials isolation, season-0 handling and the unnumbered-season skip all behave as documented.

## What `ef098c1b` changes

| Defect | Fix |
|---|---|
| A non-season item threw, which becomes the transient signal and froze every item in a wrongly-typed group | Answer `null`, matching #3402 and every other `showType`-restricted property |
| `Season Unknown` (no `IndexNumber`) logged `Action failed` every run | Own branch plus a debug line; the item is still held |
| Plex aborted on a history row with no `parentIndex` without saying why | Throw stays, message now names the show, the field and #3082 |
| Emby's per-user watch read swallowed transport failures | 403/404 stays a visibility miss, anything else reaches the caller |

The last one was the blocker. One injected 502 made the property answer `2025-01-01` for a season watched `2026-08-28`, silently, which under this exact rule collects it for deletion. `403`/`404` still skip, so genuine per-user misses are unaffected.

Whole repo suite green (3 server shards, 2 UI shards), `format:check`, `lint` and `check-types` all pass.

enoch85 (commented):
Claude Opus 4.8:

<details>
<summary>Review of #3531 at 870b1e93 (expand)</summary>

My earlier approval was wrong and I have pushed the fix for what it missed. I verified that an earlier season protects a later one, and approved on that. I never verified the other half of #751, which is the sentence that follows: *"Once I've watched 3 and it gets deleted, 4 would then be included in the rule unless it gets watched."* Plex never released it.

## The defect

Plex keeps history rows after the media is gone, and the getter counted them by design (`plex-getter.service.ts`, and it was documented in `rules.constants.ts`). So a season whose only qualifying view belonged to a since-deleted earlier season stayed protected permanently. Jellyfin and Emby did not have this, because walking current episode children excludes deleted media for free.

Reproduced with the rule the issue describes, on a real Plex server: a season group over a 6-season library, `NOT_EXISTS OR BEFORE 30 days` on the new property, watched season since deleted.

| | rule-added items | the two seasons after the deleted one |
|---|---|---|
| before | 4 | pinned out of the collection indefinitely |
| after | 6 | released, per #751 |
| control (that season present and watched today) | 4 | still protected, no over-collection |

The `Adding N media items` log line and `collection_media.isManual = 0` were used, because deleting the DB rows between runs while the Plex collection still held the items makes the sync re-adopt them as manual members and hides the difference.

## What `870b1e93` changes

Intersect the history with the seasons the show still has. Seasons are the granularity a season rule deletes at, and it is the question Jellyfin and Emby already answer, so all three providers now agree. `plexApi.getChildrenMetadata` is cached at the API layer and already used by the neighbouring `sw_lastWatched` and `sw_episodes` paths, so the extra read costs nothing per run. It returns `undefined` on failure rather than throwing, so that is failed closed explicitly.

The test that pinned the old behaviour asserted `expect(plexApi.getChildrenMetadata).not.toHaveBeenCalled()`, which shut the door on the fix. It now states which seasons the show still has, and a new case covers the deleted season. Both divergence comments in `rules.constants.ts` are updated, since the divergence is gone.

One residual difference worth knowing: a watched episode deleted from a season that still exists continues to count on Plex, where Jellyfin and Emby would drop it. Matching that exactly would mean walking every season's episodes, which is the expensive shape Plex's single history read exists to avoid.

## On two of your responses

- **"A raw or imported rule could apply the season property to a non-season item and receive `null`" - agreed and fixed.** I reverted that in `ef098c1b`. The reviewer point it answered contradicts #3402, whose commit message says: *"Tracearr answered the transient signal for a property that does not apply to the item type. The executor sweeps a library at one dataType, so that is all-or-nothing for the group; every other getter answers null there."* Throwing froze every item in a wrongly-typed group. I checked five `showType`-restricted properties against a show on a real server and all answer a definitive value, none the transient signal.
- **"Emby can report `null` when every per-user watch read fails" - not addressed, pre-existing.** Fixed in `ef098c1b` with the maintainer's agreement, because this property amplifies it: a 502 on one episode made it answer an earlier season's older date rather than `null`, so under this rule the season was collected and deleted. `403`/`404` still skip as a per-user visibility miss.

Your other deferrals are reasonable. The Emby `S(S+1)/2` season walks and the per-episode `/Users/Query` refetch are both real but belong in separate adapter work.

## Verified

Real Plex, Jellyfin and Emby containers. Full repo suite (3 server shards: 1004 / 954 / 1010, 2 UI shards: 37 files each), `format:check`, `lint`, `check-types`. Playwright confirmed the collection renders with the expected count.

</details>

PR #3584, opened 2026-08-23
feat(collections): surface deletionDate in overlay-data media

ramonskie:
## Summary

Extends the existing `GET /api/collections/overlay-data` response with each member's `deletionDate`, derived as `addDate + deleteAfterDays` via the shared `getCollectionDeleteDate` helper.

The [jellyfin-plugin-leaving-soon](https://github.com/ramonskie/jellyfin-plugin-leaving-soon) polls this endpoint (per the direction in #3521 — use `/overlay-data` rather than a new endpoint). It currently has to re-derive the deletion date from `addDate` + `deleteAfterDays`; surfacing it server-side keeps the helper integration in sync with the exact arithmetic the overlay processor already uses internally, so a "leaving soon" countdown and the overlay countdown can never drift.

## What changed

- `collections.service.ts`: `getCollectionsForOverlayData` now maps each member to `{ ...member, deletionDate }` using `getCollectionDeleteDate(member.addDate, collection.deleteAfterDays)` — the same helper the overlay processor uses (overlay-processor.service.ts:137). A member of a collection with no deletion window gets `deletionDate: null`.
- `collections.service.spec.ts`: the overlay-data spec now pins `addDate`/`deleteAfterDays` and asserts the derived `deletionDate` (and `null` without a window).

## Notes

- Pure additive: existing consumers of `/overlay-data` (media array with `mediaServerId`, `addDate`, `tmdbId`, ...) are unchanged, each member just gains `deletionDate`.
- No new endpoint, no new setting, no feature flag — per the review direction in #3521.

### DIFF
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -3116,11 +3116,27 @@ describe('CollectionsService', () => {
   });
 
   it('returns full collection media for the explicit overlay data endpoint', async () => {
-    const firstCollection = createCollection({ id: 1, title: 'First' });
-    const secondCollection = createCollection({ id: 2, title: 'Second' });
+    const firstCollection = createCollection({
+      id: 1,
+      title: 'First',
+      deleteAfterDays: 30,
+    });
+    const secondCollection = createCollection({
+      id: 2,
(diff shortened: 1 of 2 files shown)


ramonskie:
not needed in the end

PR #3601, opened 2026-08-25
chore(l10n): translate the remaining Swedish strings

enoch85:
Swedish installs now get the telemetry settings page, the first-run consent modal and the "Keep in Maintainerr only" rule option in Swedish. The telemetry work added 29 source strings that no locale had picked up, and this fills the last of them, so sv.po is complete.

### Two renderings that are not literal

| Source | Swedish | Why |
| --- | --- | --- |
| `Every ~{SAMPLE_MONTHS} months` | Ungefär var {SAMPLE_MONTHS}:e månad | Swedish needs an ordinal here, and "var ~7:e" reads badly, so the tilde survives as the word it stands for. The `:e` suffix is correct for every value from 3 up; 1 or 2 would want `:a`. |
| `The payload below` | Exempeldatan nedan | Names the sample the panel actually shows rather than the protocol term. |

### The integrity gate will fail, and that is the point

`yarn i18n:validate` rejects all 29 entries with "was changed from the base branch - translations come from Weblate, do not hand-edit". That check is working exactly as designed. Merging this means accepting one hand-written batch; the alternative is uploading the same file through Weblate instead.

Verified against the running app in Swedish, not only in specs: `{mediaServerName}` renders "Jellyfin", `{TERM_IP}`/`{TERM_MAC}` render inside the sentence, `{SAMPLE_MONTHS}` gives "Ungefär var 7:e månad", and `(Nästa: tisdag 12 jan. 07:33)` shows `{0}` going through `Intl` with the Swedish locale. `validate-catalogs --no-base`, `check-catalog` and the 45 locale specs all pass.

### DIFF
--- a/apps/ui/src/locales/sv.po
+++ b/apps/ui/src/locales/sv.po
@@ -36,7 +36,7 @@ msgstr "(manuell)"
 #. placeholder {0}: formatWhen(nextAt, locale)
 #: src/components/Settings/Telemetry/index.tsx
 msgid "(Next: {0})"
-msgstr ""
+msgstr "(Nästa: {0})"
 
 #: src/components/OverlayEditor/OverlayCanvas.tsx
 msgid "[Image]"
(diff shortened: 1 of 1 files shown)


enoch85:
Superseded: the batch went in through Weblate instead (29 updated, 0 not found), so the translations will arrive on the normal weblate-maintainerr-app sync PR. Closing rather than merging past the msgstr integrity gate.

PR #3604, opened 2026-08-26
feat(sportarr): read the native sportarr id the media server agents stamp

Sportarr:
### Description & Design

Sportarr's Plex, Jellyfin and Emby agents stamp a `sportarr` namespace on every show (`sportarr://lg-000278` on Plex, `ProviderIds.Sportarr` on Jellyfin and Emby) next to the numeric tvdb alias the Sportarr connection resolves through today. This reads that namespace as a provider id and resolves the league from it first. The alias stays as the fallback for a library that has not been refreshed since the agents started stamping the native id, so nothing that works today changes.

- `MediaProviderIds` gains `sportarr`, filed by the shared mapper table on all three servers
- the getter and the action handler resolve from the whole provider id set, native id first
- the TMDB provider no longer asks TMDB about a namespace its find endpoint cannot resolve (it only knows imdb and tvdb, and a `sportarr` key in the id bag would have produced a `/find/NaN` request)
- the media card shows the sportarr id linked to its league on sportarr.net, and no longer shows the tvdb alias, which is not a TVDB id and linked nowhere

A second PR adds a metadata provider for the new namespace so Sportarr shows get artwork; it is built on this one and follows once this lands.

### Related issue

None. Follow-up to #3306.

### AI-Assisted Development

I used an AI coding assistant to draft the mapper, helper and test changes, modelled on the existing imdb/tmdb/tvdb handling and the Sportarr connection from #3306. I reviewed every line, traced the id flow from the Plex guid through the getter, the action handler and the metadata controller, and ran format, lint, type checking, the full jest and vitest suites and the e2e matrix locally. The design and the fallback behaviour are mine.

### Checklist

### How to test

1. Point a Plex library at the Sportarr metadata provider (https://sportarr.net/plex) and refresh a show, or use a Jellyfin/Emby library with the Sportarr plugin.
2. Open the show in Maintainerr. The media card shows a `sportarr://lg-...` badge that links to sportarr.net, and no tvdb alias badge.
3. Run a Sportarr rule on that library. The getter resolves the league from the native id; a show that only carries the tvdb alias still resolves as before.

### DIFF
--- a/apps/server/src/modules/actions/sportarr-action-handler.spec.ts
+++ b/apps/server/src/modules/actions/sportarr-action-handler.spec.ts
@@ -104,6 +104,29 @@ describe('SportarrActionHandler', () => {
     expect(mockClient.deleteLeague).not.toHaveBeenCalled();
   });
 
+  it('resolves a show from the native sportarr id its agent stamps', async () => {
+    const collection = createCollection({
+      arrAction: ServarrAction.DELETE,
+      type: 'show' as MediaItemType,
+      sportarrSettingsId: 1,
+    });
+    const media = createCollectionMediaWithMetadata(collection, {
+      tvdbId: undefined,
+    });
+    mockMediaServer.getMetadata.mockResolvedValue(
+      createMediaItem({
+        type: 'show',
+        providerIds: { sportarr: [F1_EXTERNAL_ID] },
+      }),
+    );
+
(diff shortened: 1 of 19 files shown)


enoch85:
@Sportarr Thanks! Fixed some nits, merging. :100:

PR #3606, opened 2026-08-26
feat(metadata): add a Sportarr provider for league artwork and descriptions

Sportarr:
Builds on #3604 (merged). Adds a Sportarr metadata provider so leagues get posters, backdrops and descriptions like TMDB and TVDB items do. The review commits on this branch are kept as they are; the rework sits on top.

Reworked after review:

- The client is `sportarr-metadata-api` / `SportarrMetadataApiService`. A league is read from every configured Sportarr connection first, and from sportarr.net only for a league none of them tracks (an instance answers `{"error":"Series not found"}` for the rest). The league route decides the source for the season and episode routes too.
- No settings column and no migration. `SPORTARR_NET=off` in the environment stops the sportarr.net read, the same way `TELEMETRY=off` works. Settings > Metadata gets a Sportarr card that says where the artwork comes from and has a Refresh button that clears the cache and re-resolves the aliased items.
- The provider returns no year, so `validateDirectIds` accepts its ids. The TVDB `extractId` change is reverted. Every read carries the 6 hour ttl explicitly. 429s wait for `Retry-After` on top of the standard retry policy. `findByExternalId` is a stub. `parseId` is a required `IMetadataProvider` member (TMDB and TVDB parse a number). The league id helpers live in `@maintainerr/contracts`.
- `isSportarrTvdbAlias` excludes the offset itself, which the resolver already rejected as league zero, with a boundary test.
- `IMetadataProvider.isAuthorityFor(ids)`: the Sportarr provider claims any item with a Sportarr id (native or tvdb alias) and answers ahead of the primary provider, so a league that also carries a TMDB id gets the Sportarr poster with TMDB primary. TMDB and TVDB never claim and stay as fallbacks. Regression test covers a TMDB id plus a Sportarr id.
- The media modal reads a Sportarr id back as the canonical league id, so one the metadata layer answers with as a bare number links to its league rather than a page that does not exist, and it no longer earns a second badge beside the one the media server carries. A Sportarr backdrop carries its own attribution logo like a TMDB or TVDB one does.
- The 429 branch is the only thing matching 429, so the declared wait is honoured and one past the cap gives up instead of holding a poster request open. A source that answers nothing is left alone for a minute rather than costing a connect timeout per card. The provider stands down when nothing is reachable, so a claim it cannot answer stops failing a resolution other providers could finish. A provider that dates nothing says so, which drops an accept message per league to debug. TVDB declines a Sportarr alias in its fetch path rather than 404ing per league, while extractId stays honest for the Sonarr candidate. The connection list is read once per walk.
- The Settings card reuses the strings the sibling provider cards already ship translated. Catalogs extracted, ARCHITECTURE.md and AGENTS.md updated.

Verified on a rig with Jellyfin and Emby libraries stamped by the Sportarr agents and a Sportarr instance tracking one league: that league resolves through the connection with `SPORTARR_NET=off`, a league the instance does not track resolves through sportarr.net without the variable and not at all with it.

Docs: Maintainerr/Maintainerr_docs#164.


enoch85:
Tell this to your AI. :)

<details><summary>Details</summary>
<p>

Reviewed against a real Sportarr 4.1.5 instance and the published agent
contract. First, what is right: the endpoints and payload shapes are correct,
not guessed. `/api/metadata/agents/series/{id}`, `/seasons` and
`/season/{n}/episodes` all exist and return exactly the declared fields, the
`{"error":"Series not found"}` envelope at HTTP 200 is real, and snake_case is
the right casing for this upstream. Artwork does land: EPL and NBA cards render
real league posters through the Overview path.

Four things need fixing before this can go in.

### 1. The year fallback stops the feature from working

`sportarr-metadata.provider.ts:71` fills `MetadataDetails.year` from
`firstSeasonYear`, which is `Math.min` over `season_number`. The hub returns
`year: null` for real leagues, so this always runs, and season numbers are
calendar years back to a league's founding:

| League | earliest season | media server year |
|---|---|---|
| Formula 1 | 1950 | 2019 |
| NBA | 1960 | 1992 |
| English Premier League | 1992 | 2012 |

`MetadataService.validateDirectIds` treats a delta above 1 as a disagreement and
fails closed, so id resolution returns `undefined` for every league:

[WARN] Rejected direct provider IDs for media server item "Formula 1" (2019)
because no configured metadata provider confirmed the release year.
Disagreements: Sportarr returned 1950.



That kills `CollectionsService.resolveCollectionMediaArtwork` (no poster stored,
the gap this PR exists to close) and `MetadataRuleValueService.getStudios`, and
it prints a warning per league that blames the user's media server for metadata
Sportarr supplied.

The comment says the fallback exists so the year check "has something to compare
instead of warning on every resolution". With `year` left undefined the same run
succeeds: `validateDirectIds` logs `Accepted direct provider IDs ... without a
year check` at debug and returns the ids.

**Fix: delete `firstSeasonYear` and leave `year` undefined.** It also removes the
`/seasons` fetch from the poster path, which is 8 to 18 KB and 35 to 77 entries
per league for a field that path never reads.

### 2. The TVDB `extractId` change drops a live lookup candidate

`tvdb-metadata.provider.ts:28` returns `undefined` for any tvdb value in the
alias range. `hasRequiredIds` reads `extractId`, so the Sonarr policy
(`providerKeys: ['tvdb']`) can no longer be satisfied. Same item, TVDB
unconfigured:

development : sonarr lookup candidates -> [{"providerKey":"tvdb","id":900000001}]
this branch : sonarr lookup candidates -> []



Reproduced with the Sportarr provider not registered, so the TVDB change alone
causes it. It affects the Sonarr getter, the Sonarr action handler and
`ServarrTagService` (the exclusion-tag path from #3578) for anyone whose Sportarr
instance is wired in as a Sonarr connection, which `EXTERNAL_IDS.md` documents:

> The Sonarr v3 compatibility API on a Sportarr install reports the same number
> in its `tvdbId` fields, so tools that read an id from the media server and look
> it up against the install (Maintainerr, for example) close the loop entirely
> inside Sportarr.

The saving is not needed either. The same doc says a tool that does resolve an
alias against TheTVDB "gets an empty result rather than a wrong show", and
`TvdbApiService.getSeries(900000278)` already 404s harmlessly.

**Fix: revert the `extractId` change.** If you want to skip the request, decline
inside `getDetails`, which is not what `hasRequiredIds` reads.

Related: the commit message says the provider "sits before TVDB so a league never
fails a TVDB lookup first". `getOrderedProviders()` re-sorts by preference, so
under `TVDB_PRIMARY` the order is `[tvdb, tmdb, sportarr]`. The module array
order is cosmetic.

### 3. The "6 hour cache" is a 20 minute cache

`cache.ts:224` sets `stdTtl: 21600`, but all three reads call `this.get(endpoint)`
with no ttl, and `ExternalApiService.get` writes
`this.cache.set(key, data, ttl ?? DEFAULT_TTL)` with `DEFAULT_TTL = 1200`. An
explicit per-key ttl overrides node-cache's `stdTTL`, measured on the real cache
object:

configured stdTTL: 21600
explicit-ttl entry expires in 1200 s



`TvdbApiService` passes `ttl = 3600` explicitly and `TmdbApiService` passes
`43200` where it matters. **Fix: pass a ttl.** One argument.

### 4. The base URL should not be hardcoded to sportarr.net

`sportarr-hub.service.ts:13` pins `https://sportarr.net/api/metadata`. Two
problems:

- The same endpoints are served by the user's own instance. Every probe above hit
  `http://<my-sportarr>:1867/api/metadata/agents/...`, not sportarr.net.
  Upstream's `agents/README.md`: "All three can point at sportarr.net or at your
  own Sportarr instance by setting the API URL in the agent/plugin settings." The
  agents make it configurable, this does not.
- The `SportarrApi` connection from #3306 already returns most of it. Live from
  `GET /api/leagues`: `posterUrl`, `bannerUrl`, `logoUrl`, `description`,
  `formedYear`. From `GET /api/events`: `thumbUrl`, `images[]`. The getter
  already holds a `SportarrLeague`.

I also confirmed the traffic is ungated: with zero Sportarr connections
configured, one poster lookup for an aliased item produced two requests to
sportarr.net, with no setting to redirect or disable it. (To be fair: zero
requests for non-Sportarr items, normal movies and shows never touch it.)

**Fix: take the base URL from `sportarr_settings.url`, falling back to
sportarr.net.** Better still, serve the artwork from the existing connection and
drop `sportarr-hub-api` entirely.

### 5. No Settings presence, and a cache nothing can clear

`Settings > Metadata` builds one card per provider: title, "primary" radio, API
key, Test Connection, Refresh metadata. Sportarr gets none of it, and:

- `MetadataProviderPreference` has only `TMDB_PRIMARY` / `TVDB_PRIMARY`, so it can
  never be preferred.
- The cache is `persistent: true`, so `CacheManager.flushAll()` skips it at every
  rule-group start.
- `MetadataSettingsService.refreshMetadataCache(provider: MetadataProvider)` is
  typed `'tmdb' | 'tvdb'`, so the Refresh button cannot reach it.

Net: entries can only expire on their own, and a user who fixes artwork upstream
has no way to pick it up.

### Nits

- `sportarr-metadata.provider.ts:137` `findByExternalId` is unreachable. It
  answers only `type === 'tvdb'`, and both `resolveAllIds` and
  `bridgeMissingProviderIds` skip a provider whose `extractId` already resolves,
  which it does from `ids.tvdb`, the value being bridged. Delete it and its spec
  block.
- No `Retry-After` handling. Upstream's `SportarrHttp.cs` exists specifically
  because of "a 429 from the API's per-IP rate limiter" and waits it out up to
  65s. `applyHttpRetry`'s default retries 429 but on exponential backoff (roughly
  1s/2s/4s) and ignores the header. `lib/httpRetry.ts` already exports
  `isRetryableRateLimit` and `rateLimitWaitMs` for this. Pass them as overrides.
- `metadataProviderLogos` (MediaModal) has no `Sportarr` entry, so artwork from
  this provider gets no attribution. Degrades gracefully, so cosmetic, but note
  the trap before adding one: the metadata layer returns `{provider:'Sportarr',
  id: 278}` while `providerIds.sportarr` holds `'lg-000278'`, so a naive entry
  renders `sportarr://278` linking to `/browse/leagues/278`, which is not a
  league id.
- `metadata.service.ts:949` `numericId !== undefined &&` is dead.
  `Number.isFinite(undefined)` is already `false` (unlike global `isFinite`, it
  does no coercion).
- `parseId?` is the only optional member of `IMetadataProvider`, and it forces a
  fork at the single call site plus a conditional spread in the shared test
  factory. TMDB and TVDB can implement it as `Number` in two lines each and the
  fork disappears.
- ARCHITECTURE.md was updated in one place out of three. L104 still lists the api
  submodules without `sportarr-hub-api`, L163 still reads "TMDB and TVDB for
  metadata resolution". AGENTS.md "External Integrations" is not updated either.
- `sportarr-hub.interface.ts` declares 8 fields nothing reads: league `id`,
  `sport`; season `title`, `episode_count`; episode `id`, `season_number`,
  `title`, `air_date`. Four are required, which forces the specs to invent
  `episode_count: 4` and `title: '2001'`.
- "Hub" is invented vocabulary. Nothing in this repo or in Sportarr uses it;
  upstream calls it the metadata agent API. `sportarr-metadata-api` /
  `SportarrMetadataApiService` would read like every other module. Relatedly,
  `cache.ts:224` names the cache `'sportarr.net'` where every sibling names the
  API ("TheTVDB API", "Seerr API", "Emby").
- `sportarr-metadata.provider.ts:160` `leagueBackdrop` takes an inline
  `{fanart_url?, banner_url?}` duplicating two fields of `SportarrHubLeague`,
  which both call sites pass whole. `findSeason` (:167) and `findEpisode` (:172)
  have no declared return type.
- `sportarr-hub.service.spec.ts`: three of four tests assert
  `ExternalApiService`'s pass-through
  (`toHaveBeenCalledWith('/agents/series/lg-000278', undefined)`), not anything
  this service owns. The one service-owned behaviour is the `{error}` envelope.
  It also hand-rolls DI via `(service as any).axios` and mutates the global cache
  singleton in `beforeEach`, unlike the `@suites` `TestBed` used by the sibling
  provider spec.
- `getSeasons` / `getSeasonEpisodes` return `[]` on a failed read, so an outage
  is indistinguishable from "no seasons". The rest of the Sportarr surface from
  #3306 is deliberately fail-closed here. Moot once the year fallback goes, but
  worth keeping in mind if the season paths grow.
- Import insertions break the alphabetical ordering the surrounding blocks keep,
  in `metadata.module.ts` (twice) and `tvdb-metadata.provider.ts`.

### Rebase note

#3604 picked up a review commit that trims `sportarr-external-id.ts` to the two
functions the connection uses and moves the alias-range predicate to
`@maintainerr/contracts` as `isSportarrTvdbAlias`. When you rebase, this branch
needs `sportarrLeagueNumberFromExternalId`, `sportarrLeagueExternalIdFromNumber`
and the league-number-from-alias helper exported again (or, better, use the
contracts predicate in `tvdb-metadata.provider.ts` instead of importing a
`servarr-api` helper into the metadata layer). It will fail to compile rather
than fail silently.

</p>
</details>

enoch85:
Last commit now resolves metadata correctly in my dev box.

image

enoch85:
Also note:

- P3 - shared alias predicate includes an ID the resolver rejects. [sportarr.constants.ts (line 19)](/workspace/packages/contracts/src/sportarr/sportarr.constants.ts:19) treats 900000000 as a Sportarr TVDB alias (>= offset), but the resolver explicitly rejects it as league zero in [sportarr-external-id.ts (line 63)](/workspace/apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts:63). This contradicts the documented exclusive range and the claimed single source of truth. Reproduction: isSportarrTvdbAlias(900000000) === true, while sportarrLeagueExternalIdFromTvdbAlias(900000000) === null. Use > for the lower bound and retain a boundary test.

enoch85:
Would be great if we could avoid migrations! 🙏🏼

Not at the cost of workaround code mess, but if it's possible in any other way, that would be great!

apps/server/src/modules/api/external-api/external-api.service.ts
```diff
@@ -13,6 +13,9 @@ const DEFAULT_ROLLING_BUFFER = 10000;
 interface ExternalAPIOptions {
   nodeCache?: NodeCache;
   headers?: Record<string, unknown>;
+  // Overrides for the standard retry policy, for an upstream that says how
+  // long to wait in a 429 response.
+  retry?: Parameters<typeof applyHttpRetry>[1];
```
enoch85: Why do you need an override here?

apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts
```diff
@@ -46,30 +46,23 @@ export function sportarrLeagueExternalIdFromNumber(n: number): string {
...
 // Before the agents stamped the native id they only carried a numeric alias in
 // the tvdb namespace (e.g. tvdb://900000278). Libraries that were never
 // refreshed since still resolve through it.
-export function sportarrLeagueNumberFromTvdbAlias(
+export function sportarrLeagueExternalIdFromTvdbAlias(
   tvdbAlias: number | undefined | null,
-): number | undefined {
+): string | null {
   if (!isSportarrTvdbAlias(tvdbAlias)) {
-    return undefined;
```
enoch85: Are you aware if the difference between null and undefined in the code base?

apps/server/src/modules/api/sportarr-metadata-api/sportarr-metadata.service.ts
```diff
@@ -0,0 +1,140 @@
...
+@Injectable()
+export class SportarrMetadataApiService extends ExternalApiService {
+  constructor(
+    private readonly settings: SettingsDataService,
+    protected readonly logger: MaintainerrLogger,
+  ) {
+    logger.setContext(SportarrMetadataApiService.name);
+    // Every read passes an absolute URL, so each source caches under its own
+    // host and one never serves another's answer for the same path.
+    super('', {}, logger, {
```
enoch85: Hmm, looks odd. Is this the Maintainerrlogger?

apps/server/src/modules/metadata/interfaces/metadata-provider.interface.ts
```diff
@@ -21,9 +21,9 @@ export interface IMetadataProvider {
 
   /**
    * The provider's numeric id from the string a media server carries for it.
-   * Providers whose ids are plain numbers leave this out.
+   * Undefined when the string is not one of this provider's ids.
    */
-  parseId?(value: string): number | undefined;
+  parseId(value: string): number | undefined;
```
enoch85, line 33: Perfect!

enoch85 (commented):
Quickly scrolled on mobile.

enoch85:
GPT Terra 5.6:

<details>
<summary>Earlier review of #3606 (expand)</summary>

- [P1] Give Sportarr precedence when a Sportarr league ID is present. `MetadataService.getOrderedProviders()` can only prefer TMDB or TVDB, and `resolveImageUrl()` returns the first successful provider. Sportarr is therefore last whenever TMDB is primary. This makes the central feature fail for the normal dual-ID case, despite the UI promising that artwork is read from the user's Sportarr connections first.

  Reproduced on the exact PR server against the real Docker stack: `GET /api/metadata/image/show?sportarrId=lg-000278` returned `provider: Sportarr`; the same Formula 1 request with its actual IDs, `tmdbId=87083&tvdbId=900000278&sportarrId=lg-000278`, returned TMDB's URL with `provider: TMDB`. Both providers had artwork, so the configured Sportarr instance was never queried for the result.

  Make the provider-selection policy put Sportarr first whenever `SportarrMetadataProvider.extractId(ids)` resolves (native ID or legacy TVDB alias), while retaining TMDB/TVDB as fallbacks when Sportarr has no result. Module injection order is insufficient because `getOrderedProviders()` reorders it. Add one focused `MetadataService` regression test covering a TMDB ID plus a Sportarr ID and asserting the Sportarr result.
</details>

Sportarr:
> Tell this to your AI. :)
> 
> Details

Thanks for the review and the commits. I kept yours as they are and put my changes on top. Everything from the list is in, the module is sportarr-metadata-api now, no year from the provider, TVDB extractId back to what it was, ttl on every read, Retry-After on 429s, parseId required, helpers in contracts.

Sportarr:
> Would be great if we could avoid migrations! 🙏🏼
> 
> Not at the cost of workaround code mess, but if it's possible in any other way, that would be great!

Migration is gone. SPORTARR_NET=off in the environment stops the sportarr.net read, same as TELEMETRY=off, and the Settings card just says where the artwork comes from plus a Refresh button.

Sportarr:
> GPT Terra 5.6:
> 
> ACTUAL REVIEW
> * [P1] Give Sportarr precedence when a Sportarr league ID is present. `MetadataService.getOrderedProviders()` can only prefer TMDB or TVDB, and `resolveImageUrl()` returns the first successful provider. Sportarr is therefore last whenever TMDB is primary. This makes the central feature fail for the normal dual-ID case, despite the UI promising that artwork is read from the user's Sportarr connections first.
>   Reproduced on the exact PR server against the real Docker stack: `GET /api/metadata/image/show?sportarrId=lg-000278` returned `provider: Sportarr`; the same Formula 1 request with its actual IDs, `tmdbId=87083&tvdbId=900000278&sportarrId=lg-000278`, returned TMDB's URL with `provider: TMDB`. Both providers had artwork, so the configured Sportarr instance was never queried for the result.
>   Make the provider-selection policy put Sportarr first whenever `SportarrMetadataProvider.extractId(ids)` resolves (native ID or legacy TVDB alias), while retaining TMDB/TVDB as fallbacks when Sportarr has no result. Module injection order is insufficient because `getOrderedProviders()` reorders it. Add one focused `MetadataService` regression test covering a TMDB ID plus a Sportarr ID and asserting the Sportarr result.

Good catch. Providers now have isAuthorityFor(ids). Sportarr claims anything with a Sportarr id, native or alias, TMDB and TVDB never do, and a provider that claims the item goes first with the rest as fallbacks. Your exact query with TMDB primary answers Sportarr now, tmdbId alone still answers TMDB. There's a regression test with a TMDB id plus a Sportarr id.

apps/server/src/modules/api/external-api/external-api.service.ts
Sportarr: applyHttpRetry runs once in ExternalApiService with no overrides, and running it again from the subclass would stack a second retry interceptor on the same axios instance. The option just threads the 429 wait through.

apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts
Sportarr: That string/null is the contract the helper already had before this PR, the getter and action handler check for null. I only removed the second early return. Happy to move it to undefined if you want it consistent.

apps/server/src/modules/api/sportarr-metadata-api/sportarr-metadata.service.ts
Sportarr: Yes, same constructor shape as TmdbApiService. The odd part was the empty base URL, it carries the sportarr.net metadata path now like your version did.

Sportarr:
> Also note:
> 
> * P3 - shared alias predicate includes an ID the resolver rejects. [sportarr.constants.ts (line 19)](/workspace/packages/contracts/src/sportarr/sportarr.constants.ts:19) treats 900000000 as a Sportarr TVDB alias (>= offset), but the resolver explicitly rejects it as league zero in [sportarr-external-id.ts (line 63)](/workspace/apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts:63). This contradicts the documented exclusive range and the claimed single source of truth. Reproduction: isSportarrTvdbAlias(900000000) === true, while sportarrLeagueExternalIdFromTvdbAlias(900000000) === null. Use > for the lower bound and retain a boundary test.

Fixed. The predicate uses > for the lower bound now, both callers lost their own league zero check, and there's a boundary test for the offset itself.

enoch85:
While you're at it:

1. Fix the backdrop link sportar://2737373
2. Fix the Sportarr logo in the backdrop if the metadata is fetched from there. Look at how TVDB and TMDB does it.

enoch85:
Claude Opus 4.8:

<details>
<summary>Re-review of #3606 at 129d1049 (expand)</summary>

Re-reviewed against the real Docker stack: a live Sportarr 4.1.5 instance, dev Jellyfin, Playwright, and the compiled classes rather than reasoning from source alone. Three of the commits on this branch were written by Claude in an earlier session (`a29f3940`, `1e0a839b`, `c5645a36`), and the findings below are hardest on those, not softer.

**Both blockers from the previous review are fixed, and I verified each rather than taking the diff's word for it.**

| Previously raised | Status |
|---|---|
| Year fallback fed a season number into `MetadataDetails.year`, failing `validateDirectIds` for every league | **Fixed.** `getDetails` reports no year. Against the live instance, Formula 1 and English Premier League now resolve ids and return a `posterUrl`, so collection artwork works |
| `TvdbMetadataProvider.extractId` declined the alias range, dropping the Sonarr tvdb candidate | **Fixed.** Reverted to the plain numeric read |
| "6 hour cache" was really 20 minutes | **Fixed.** TTL passed explicitly on every read |
| `sportarrmetadata` cache unclearable (`persistent`, and `MetadataProvider` typed `'tmdb' \| 'tvdb'`) | **Fixed.** `METADATA_CACHE_ID` map plus the Refresh button. Verified live |
| Dead `findByExternalId` branch | **Fixed.** Bare stub |
| Eight unread interface fields | **Fixed.** Interface trimmed to the 11 fields actually read |
| `parseId` optional interface member forcing a call-site fork | **Fixed.** Now required, all three providers implement it |
| ARCHITECTURE.md updated in 1 of 3 places, AGENTS.md untouched | **Fixed.** All four spots updated |
| No Settings presence | **Fixed.** Card, toggle, Refresh, connection test |

**Migration verified end to end** (rule 8). Parsed the generated SQL: `up` creates 48 columns and copies all 47 existing ones, target and select lists identical, new column is the only one not copied. Ran it against a copy of a real install: one column added, **zero** value drift, **zero** column-definition drift, other tables intact, default `1`. Extracted and applied `down()`: schema and row byte-identical to pre-migration. Timestamp orders after `AddTelemetryEnabled`. Migration matrix 9/9.

Full suite green: **2925 server tests** (3 shards), **403 UI tests** (2 shards), catalogs valid, extraction current.

---

## Major

### 1. The 429 override bypasses the very cap it imports

`sportarr-metadata.service.ts:43`

```js
retryCondition: (error) =>
  axiosRetry.isNetworkOrIdempotentRequestError(error) || isRetryableRateLimit(error),
retryDelay: (retryCount, error) =>
  rateLimitWaitMs(error) || axiosRetry.exponentialDelay(retryCount, error),
```

`isNetworkOrIdempotentRequestError` already returns true for **any** 429 on a GET, so the `|| isRetryableRateLimit(error)` clause can never change the outcome, and the `MAX_RATE_LIMIT_WAIT_MS` guard it exists to apply never gates anything. `retryDelay` then returns the raw, uncapped wait. Proven:

```
isNetworkOrIdempotentRequestError(429 GET) = true
rateLimitWaitMs(Retry-After: 600s)         = 600000 ms
isRetryableRateLimit(same)                 = false   <- the 60s cap says "give up"
PR retryCondition                          = true    <- retried anyway
PR retryDelay                              = 600000 ms
```

A 429 carrying `Retry-After: 600` holds a user-facing poster request open for **10 minutes**, up to 3 times. `httpRetry.ts` says why the cap exists: "Past this the send would be held open too long ... so give up instead." Sportarr documents a per-IP rate limiter and its own agent handles 429 with `Retry-After`, so this is an expected response, not a hostile edge.

Fix: the 429 branch must be the only thing matching 429, so the cap gates:
```js
retryCondition: (e) => axiosRetry.isNetworkError(e) || is5xx(e) || isRetryableRateLimit(e),
```

### 2. An unreachable source is never remembered, so the cost repeats forever

`sportarr-metadata.service.ts:122`. `ExternalApiService.get` returns `undefined` on failure and caches nothing, so every call re-walks and re-fails. Measured against a genuinely unreachable host:

```
1st getLeague on an unreachable source: api.get=1, 1505ms
2nd getLeague on an unreachable source: api.get=1, 1486ms
3rd getLeague on an unreachable source: api.get=1, 1557ms
```

No decay. A poster costs two `resolveSource` walks (`getDetails` then `getPosterUrl`), so roughly 3s per card, on every page load, forever. Ten leagues is about 30s of blocked requests per Overview view. That 1.5s is `ECONNREFUSED`; behind a DROP firewall it is the 10s axios timeout times retries, closer to 40s per lookup.

This is a weakness of the source-walk **Claude wrote in `a29f3940`** and the contributor inherited. Fix: remember "this source could not answer" for a short window (a minute or two), separate from the 6 hour positive TTL.

### 3. A source-less Sportarr provider vetoes id resolution for alias-tagged shows

`sportarr-metadata.provider.ts:35`. `isAvailable()` is hardcoded `true`, and `extractId` claims the tvdb alias. So `hasAvailableDirectIds` becomes true, `validateDirectIds` runs, Sportarr cannot answer, and the **whole** id resolution aborts, taking unrelated ids with it. Measured, same show, TVDB unconfigured:

```
no Sportarr provider at all (= development)        -> [{"providerKey":"tvdb","id":900000001}]
Sportarr provider, working source                  -> [{"providerKey":"tvdb","id":900000001}]
Sportarr provider, NO source (fallback off)        -> []
Sportarr provider, only sportarr.net (unreachable) -> []
```

Consumers: the Sonarr getter, the Sonarr action handler, and `ServarrTagService` (the exclusion-tag path from #3578). Reachable whenever the toggle is off with no connection, or sportarr.net is unreachable.

`hasSource()` already exists and answers exactly this question, but only `metadata-settings.service.ts:228` uses it. Wiring it into `isAvailable()` fixes the configuration case. The unreachable case is the sharper question: a Sportarr outage should not veto an unrelated tvdb id.

---

## Minor

### 4. A warn per league, per resolution, that no user can act on

Removing the year fallback (correctly) unmasked `metadata.service.ts:1081`:

```
[WARN] Accepted direct provider IDs for "Formula 1" via Sportarr without a year check
       - Sportarr returned no release year for this entry.
```

Reproduced for every league. That branch's own comment justifies warn with "TMDB/TVDB almost always have one" - a premise Sportarr permanently breaks, as this PR's provider comment states outright. Project rule: warn is for events that are "rare AND the user can act on it", never per-item chatter. Needs a way for a provider to say it has no year concept, so the level drops to debug.

### 5. TVDB-configured installs now fire a doomed lookup per league

Reverting the `extractId` guard (which I recommended, to fix finding 3's predecessor) means TVDB, when configured, extracts `900000278`, calls `/series/900000278/extended`, 404s, and `tvdb.service.ts:231` logs `warn` each time. Both states are wrong in different ways.

Better than either: leave `extractId` honest so candidates and `hasRequiredIds` keep working, and decline the alias range in TVDB's **fetch** path instead. No request, no warn, no lost candidate.

### 6. Two uncached SQLite reads per lookup

`sources()` queries `sportarr_settings` on every call and `resolveSource` re-runs per public method. Measured, fully warm:

```
WARM  one poster lookup : settings reads = 2 | api.get calls = 2
WARM  3 cards           : settings reads = 6 | api.get calls = 6
```

2N queries per Overview page even when every HTTP answer is cached. TMDB/TVDB read settings once at boot and refresh on `Settings_Updated`. The honest complication: the Sportarr connection CRUD emits no event (confirmed - `addSportarrSetting`/update/delete do not emit), which is *why* reading per call is defensible. Emitting there would let this be cached like everything else.

### 7. Four new strings that the sibling cards already ship translated

The sibling provider cards parameterise:
```js
t`${{ providerTitle: config.title }} settings updated`
t`Failed to refresh ${{ providerTitle: config.title }} metadata`
```
`SportarrSection.tsx` hardcodes `Sportarr settings updated`, `Sportarr settings could not be updated`, `Sportarr metadata refresh started`, `Failed to refresh Sportarr metadata`. Four new msgids added untranslated to 13 locales, where reusing the parameterised form with `providerTitle: 'Sportarr'` would ship translated on day one and cost translators nothing. (`Refreshing...` and `Refresh metadata` are reused correctly.)

### 8. The toggle and card shell are copy-pasted from the sibling file

`SportarrSection.tsx:124` reproduces `PrimarySwitch` from `Metadata/index.tsx:380` verbatim - same `h-6 w-11`, `bg-maintainerr-600`, `translate-x-6`/`translate-x-1`, same transitions. The card wrapper and header (`flex h-full flex-col rounded-xl bg-zinc-800 ...` plus the ghost Refresh button) are duplicated too. `PrimarySwitch` is local but two files away in the same folder. Rule 5 asks for shared components.

### 9. One setting, two public names

`settingsUpdateSchema` gains `sportarr_net_fallback` (contracts `settings.ts:55`) while the dedicated endpoint uses `use_sportarr_net`. The dedicated path calls `saveSettings(settings: Settings)` with the entity shape, so the schema entry is not needed for it, and the UI never sends it. It just adds a second writable name on `POST`/`PATCH /api/settings`.

### 10. The toggle renders on while loading, then flips

`setting?.use_sportarr_net ?? true`. An install with it stored `false` shows the switch on, then off. The control is disabled during load so it is visual only, but the direction says placeholders must not change active state.

### 11. The new column is not in the migration matrix's parity block

`migrations.spec.ts` was correctly repointed (the up/down existence test now keys on `sportarr_net_fallback`), but the curated per-column block that every prior migration added to (`AddSportarrSettings:`, `AddTelemetryEnabled:`, `AddKeepCollectionInMaintainerrOnly:`) has no entry asserting `{type:'boolean', notnull:1, dflt_value:'1'}`. That default is the entire upgrade decision - a drift to `0` would silently opt every install out and no test would notice.

---

## Nits

- `metadata.service.ts:948` - `numericId !== undefined &&` is still dead; `Number.isFinite(undefined)` is already false.
- `migrations.spec.ts:207` - the "Grandfathered" comment still describes `telemetryEnabled` as the newest migration's column.
- `metadata-settings.service.ts:253` - the sportarr refresh scopes by `tvdbId` inside the alias window and the comment asserts "the alias is what the collection stores for them, native id or not". True while Sportarr emits the tvdb envelope; `EXTERNAL_IDS.md` allows dropping it per flow, and `collection_media` has no sportarr column, so this quietly misses those rows later.
- `sportarr-metadata.service.ts:39` - `super('', ...)` plus absolute endpoints makes `describeRequestTarget` print a leading-slash-prefixed absolute URL in the debug failure line.

---

## Checked and found clean

Worth stating so nobody re-checks: no em or en dashes anywhere in the added lines or the seven commit messages; no regex for simple string work; no real media titles; catalogs valid and extraction current with **empty** `msgstr` in every non-English locale (correct - Weblate fills them); `sportarr_net_fallback` staleness is properly wired through `init()` and `Settings_Updated`; the toggle persists and takes effect immediately, verified live.

## Two claims I want to retract or refute

- **"Hub is invented vocabulary" from my previous review was wrong.** Sportarr's own source uses it throughout - `HubChangesPollerService`, `HubList`, `HubCastMember`, "hub truth". The rename to `sportarr-metadata-api` reads better on Maintainerr's side either way, but it was not correcting an error of yours.
- **Turning the fallback off does not keep serving sportarr.net answers.** `sources()` is re-evaluated per call and the sportarr.net cache keys are simply never looked up. Live check with the toggle off: an untracked league returned nothing in **51ms**, no outbound call.

## Standing caveat, unchanged

Artwork URLs are absolute `https://sportarr.net/static/...` minted by Sportarr, and an instance 404s those paths (verified). Reading metadata from the local connection removes the third-party **API** dependency, not the **image** one. An offline install still gets grey cards. Not this PR's job to fix, but worth knowing before the toggle is described as making Sportarr artwork self-hosted.

</details>

Sportarr:
> While you're at it:
> 
> 1. Fix the backdrop link sportar://2737373
> 2. Fix the Sportarr logo in the backdrop if the metadata is fetched from there. Look at how TVDB and TMDB does it.

Both done. The modal reads a Sportarr id back as the canonical league id now, so the one the metadata layer answers with as a bare number links to the league instead of a page that doesn't exist, and it stopped adding a second badge next to the one the media server already carries. The backdrop gets a Sportarr logo like TMDB and TVDB do, a square mark rather than stretched into the wordmark box. Screenshots in the PR.

image

Sportarr:
> Claude Opus 4.8:
> 
> Re-review of #3606 at [129d104](https://github.com/Maintainerr/Maintainerr/commit/129d1049d9f65d6e3e79a8fb6c9dcc6149256a17) (expand)

Thanks. I fixed the four that still applied: the 429 branch is the only thing matching 429 now so the cap actually gates, an unreachable source is left alone for a minute instead of costing a connect timeout per card, the provider stands down when it has nowhere to read so it stops vetoing an alias resolution TVDB could have finished, and a provider that dates nothing says so, which drops that accept message to debug. TVDB declines the alias in the fetch path instead of extractId, so no 404 and no warn while the Sonarr candidate stays. Connections are read once per walk, and the card reuses your parameterised strings.

The migration, the toggle, and the second setting name went away a few commits ago, so 9, 10, 11 and the switch half of 8 don't apply anymore. The P1 on ordering was already fixed before that review; providers have isAuthorityFor now.

Fair point on the image URLs. Artwork is minted by sportarr.net either way, so reading metadata locally doesn't make an offline install self-hosted, and I won't describe it that way.

enoch85:
> a square

That looks odd. Stretch it out so that it follows the UI. Consistency is king! 😉

Other than that, great progress here!

apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts
enoch85: I want it to be consistent yes 👍🏼

enoch85:
> Migration is gone. SPORTARR_NET=off

most people won't bother turning it off which means that we will ask sportarr.net all the time.

How does that affect non Sportarr users? Default should be off I think.

enoch85:
GPT Terra 5.6:

<details>
<summary>Review of #3606 at 2d22fc2d (expand)</summary>

Fresh sweep of the current head against the repository rules, implementation history, upstream Sportarr contract, and real dev stack.

## Major: a supported Sportarr alias is discarded when it is not the first TVDB ID

Media-server IDs are arrays, and this repository explicitly accounts for a real TVDB ID and a Sportarr alias coexisting in no guaranteed order: [sportarr-external-id.ts:30-50](https://github.com/Maintainerr/Maintainerr/blob/2d22fc2df2eb9aa4761bf2728ffa4d4764236a90/apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.ts#L30-L50) scans every value, with a regression for `tvdb: ['342040', '900000278']` at [sportarr-external-id.spec.ts:107-115](https://github.com/Maintainerr/Maintainerr/blob/2d22fc2df2eb9aa4761bf2728ffa4d4764236a90/apps/server/src/modules/api/servarr-api/helpers/sportarr-external-id.spec.ts#L107-L115). This also matches Sportarr's [documented compatibility model](https://github.com/Sportarr/Sportarr/blob/6b167b3fe342d278af25efa951fedfc60ea126ad/docs/EXTERNAL_IDS.md): native IDs and TVDB aliases can coexist, with native preferred and aliases retained for old libraries.

The two metadata entry paths ignore that contract:

- [mediaTypeUtils.ts:32-45](https://github.com/Maintainerr/Maintainerr/blob/2d22fc2df2eb9aa4761bf2728ffa4d4764236a90/apps/ui/src/utils/mediaTypeUtils.ts#L32-L45) sends only `values[0]`.
- [metadata.service.ts:940-970](https://github.com/Maintainerr/Maintainerr/blob/2d22fc2df2eb9aa4761bf2728ffa4d4764236a90/apps/server/src/modules/metadata/metadata.service.ts#L940-L970) parses only `values[0]`.

Reproduced with `providerIds: { tvdb: ['342040', '900000278'] }`: the UI builds only `tvdbId=342040`, and `resolveIdsFromMediaItem` produces `{ type: 'tv', tvdb: 342040 }` instead of a Sportarr ID. The Sportarr provider is therefore never authoritative, so these older/mixed library items still miss Sportarr artwork and overview in posters/modals, collection artwork persistence, and metadata-backed rule values.

Scan the complete arrays for a native Sportarr ID or reserved TVDB alias before reducing generic provider IDs to one scalar, and cover this exact ordering in both metadata-service and UI path tests. Prefer one shared resolver so the established connection/action and metadata paths cannot drift again.

## Validation

Real Sportarr league, season, episode, not-found, cache-refresh, native-ID, legacy-alias, mixed-ID precedence, and TMDB control calls were exercised. Maintainerr backend calls passed against the configured Jellyfin, Radarr, Sonarr, Sportarr, and qBittorrent services; direct probes reached Plex, both Jellyfin instances, Emby, Seerr, Tautulli, and Streamystats. Tracearr remains independently unhealthy, and the configured Seerr key is invalid.

Playwright confirmed the shared edit modal enables Refresh metadata and returns the real success toast, while the add-server modal keeps it disabled. Focused suites passed: 127 server tests and 22 UI tests. The full `yarn test` run was intentionally not run per repository review guidance. Current-head type checking, formatting, lint, and catalogs are green; integration and Docker builds are still running.

</details>

enoch85:
Claude Opus 4.8:

<details>
<summary>Review of #3606 at 721ccdd8 (expand)</summary>

Third pass, against the real stack: a live Sportarr 4.1.5 instance, dev Jellyfin, Playwright, and the compiled classes rather than reading source alone. Three commits on this branch were written by Claude in an earlier session, so the source-walk in particular is mine and is judged as harshly as the rest.

**Almost everything from the last review is fixed, and I re-ran each proof rather than trusting the diff.** One new defect is a blocker, and it is in the fix for one of the old ones.

## Blocker: the stand-down latches permanently

`sportarr-metadata.service.ts:129` and `:145`

`isAvailable()` reads `reachableSourceCount`. That counter is written in exactly one place, `sources()`, and `sources()` is reached from exactly two: `onModuleInit`, and `resolveSource` via `getLeague`/`getSeasons`/`getSeasonEpisodes`. Those are only called through the provider, which `getOrderedProviders()` filters out the moment `isAvailable()` is false. So once every source lands in the 60 second cooldown, nothing is left that can ever re-evaluate `unreachableUntil`, and the cooldown can never be observed to expire.

One transient failure disables Sportarr artwork **for the life of the process**. Reproduced against the compiled classes, waiting past the cooldown with the source healthy again:

```
after onModuleInit, source down :  isAvailable = true
lookup 1 (source DOWN)          :  poster=none  isAvailable=false  axiosCalls=1
lookup 2 (source DOWN)          :  poster=none  isAvailable=false  axiosCalls=1
lookup 3 (source DOWN)          :  poster=none  isAvailable=false  axiosCalls=1

--- source is now UP again ---
lookup 1 (source UP)            :  poster=none  isAvailable=false  axiosCalls=1
lookup 2 (source UP)            :  poster=none  isAvailable=false  axiosCalls=1
lookup 3 (source UP)            :  poster=none  isAvailable=false  axiosCalls=1
```

and with a real 65 second wait, past `SOURCE_FAILURE_COOLDOWN_MS`:

```
after the outage        : isAvailable = false | axiosCalls = 1
source back up. waiting 65s, past the 60s SOURCE_FAILURE_COOLDOWN_MS ...
after the cooldown      : isAvailable = false | poster = none | axiosCalls = 1
sportarr.net/connection never retried: true
```

One request in the whole process lifetime, then silence. A Sportarr instance restarting, a brief DNS blip, or Maintainerr booting a few seconds before the Sportarr container is enough.

This is the flip side of the "left alone for a minute" cost fix: the cost problem is solved precisely because the provider never comes back. Both behaviours come from the same latch.

Fix: derive the answer instead of caching a count from the last walk, so it self-heals when the cooldown passes.

```ts
hasReachableSource(): boolean {
  const now = Date.now()
  return (this.configuredSources?.sources ?? []).some(
    (source) => (this.unreachableUntil.get(source) ?? 0) <= now,
  )
}
```

That reads state the class already holds, stays synchronous, and needs no other change. `onModuleInit` still seeds `configuredSources`.

---

## Minor

### The env var is documented nowhere

`SPORTARR_NET` is read at `sportarr-metadata.service.ts:163`. The commit message cites `TELEMETRY=off` as the precedent, but that one is documented at `README.md:82` in the compose example, and `ARCHITECTURE.md:193-196` carries an explicit env-var list (`VERSION_TAG`, `GIT_SHA`, `LOG_LEVEL`, `DEBUG`). `SPORTARR_NET` is in neither, nor the Dockerfile or compose. Grepped the whole tree: the only place a user can learn it exists is a sentence inside a Settings card.

### The card instructs rather than reflects

The Telemetry precedent does more than name the variable. `Telemetry/index.tsx:222` renders "Disabled by TELEMETRY=off in the environment, which overrides this setting" **conditionally**, driven by state the server exposes (`settings.controller.ts:564`, `telemetry.service.ts:98`). The Sportarr card shows one static instruction to every user regardless of whether it is set, so someone who has set it gets no confirmation it took effect. Same shape as telemetry would cost one boolean on the settings response.

### The 60 second cap bounds one wait, not the request

`applyHttpRetry` sets `retries: 3`, so a 429 sitting exactly at the cap is retried three times at 60s each: about **180 seconds** of a held-open poster request. The cap does its job per-attempt; nothing bounds the total. Verified the branch itself is now correct:

```
429 Retry-After 5s   (under the cap)   retry= true  delay= 5000
429 Retry-After 600s (over the cap)    retry= false delay= -
429 no Retry-After                     retry= true  delay= ~209ms
500                                    retry= true  delay= ~211ms
network error                          retry= true  delay= ~224ms
```

### Two copies of the Sportarr id format

`sportarr.constants.ts` gains `sportarrLeagueNumber` (:43) and `sportarrLeagueId` (:65). `sportarr-external-id.ts` still keeps its own `LEAGUE_ID_PREFIX`/`LEAGUE_ID_PAD` (:7-8) and private `leagueNumberFromExternalId` (:13) / `leagueExternalIdFromNumber` (:38) doing the same parse and the same pad. The server helper was not switched over, so the canonical id format now lives in two places that must be kept in step by hand.

### SPORTARR_NET=off does not stop sportarr.net traffic

Worth saying plainly because the card implies otherwise. Artwork URLs are absolute `https://sportarr.net/static/...` links minted by Sportarr, and an instance 404s those paths (verified). Turning the variable off stops the **metadata** read; the browser still fetches every poster and backdrop from the sportarr.net CDN. The card's wording should not leave a user thinking it makes Sportarr artwork self-hosted.

---

## Nits

- `metadata-settings.service.ts:252` - the refresh sweep hardcodes `cm.tvdbId >= :aliasStart` while `isSportarrTvdbAlias` was just changed from `>=` to `>`. Unreachable (there is no league zero) but it is exactly the drift the shared predicate exists to stop.
- `tmdb-metadata.provider.ts` / `tvdb-metadata.provider.ts` - `isAuthorityFor` and `parseId` land as byte-identical stubs in both. Fine as an explicit contract, worth knowing it is duplication rather than behaviour.
- `hasSource()` (configured, async) and `hasReachableSource()` (last walk saw something, sync) are a confusable pair. Each is documented; the names alone do not separate them.

---

## Verified fixed

Every one re-proved, not read off the diff.

| Previously | Now |
|---|---|
| 429 override inert, uncapped wait could hang a request for minutes | Branches on status first, so the cap gates. Table above |
| Unreachable source never remembered, ~1.5s per lookup forever | Remembered. Also the blocker above |
| Source-less provider vetoed id resolution, killing Sonarr tvdb candidates | Fixed. See below |
| Warn per league per resolution for a missing year | `hasReleaseYears` gates it; `false` on Sportarr, `true` on TMDB and TVDB, so the suspicious-record warn is preserved where it matters |
| TVDB 404ing per league | Declines in the fetch path, `extractId` left honest |
| Settings list re-queried per read | 5s TTL |
| Four hardcoded strings duplicating translated sibling messages | Reuses `t\`${{ providerTitle: 'Sportarr' }} ...\`` |
| Hand-rolled toggle duplicating `PrimarySwitch` | Gone with the switch |
| One setting, two public names | Gone with the column |
| Migration risk, entity column, DTO | All removed. Grepped for `sportarr_net_fallback`, `use_sportarr_net`, `sportarrMetadataSetting`, `AddSportarrNetFallback` across the tree: no leftovers |
| Sportarr backdrop had no attribution, and a bare number linked to a page that does not exist | Logo entry plus `displayProviderId`, which also dedupes the badge |

The Sonarr case that regressed twice, with a development baseline so the remaining gap is not misattributed:

```
no Sportarr provider (= development), TVDB off        -> [{"providerKey":"tvdb","id":900000001}]
no Sportarr provider (= development), TVDB configured -> []
Sportarr working source, TVDB off                     -> [{"providerKey":"tvdb","id":900000001}]
Sportarr NO source, TVDB off                          -> [{"providerKey":"tvdb","id":900000001}]
Sportarr NO source, TVDB configured                   -> []
```

The TVDB-configured row is `[]` on development too, so it is pre-existing and not this PR's.

## isAuthorityFor: clean

I went looking for a defect in the new cross-cutting change and did not find one. It reads `ids` through an optional parameter, so `getOrderedProviders()` with no argument behaves exactly as before; `getOrderedProviderKeys()` and the bridging loops are untouched. The partition is stable, `filter` preserving relative order within each half. Three call sites. Live, with TMDB primary:

```
normal movie (control)                        -> TMDB      (unchanged)
normal show  (control)                        -> TMDB      (unchanged)
Formula 1: TMDB id AND Sportarr alias          -> Sportarr  (was TMDB)
English Premier League: alias only             -> Sportarr
```

That middle row is the bug the change exists to fix, and the controls confirm nothing moved for anyone else.

## Also clean

No em or en dashes or smart quotes in any added line or in any of the commit messages. No regex for simple string work. No real media titles. Catalogs valid, extraction current, new `msgstr` empty in all 13 locales, no `#~` resurrections. Server and UI `tsc` clean, both eslint runs clean. Full suite green: **2936 server tests** across 3 shards, **404 UI tests** across 2.

</details>

enoch85:
I just checked UI, and while the metadata connection placement is somewhat correct, I still think that a better option is to scratch that and just add a metadata refresh button here instead

image

Sportarr:
> I just checked UI, and while the metadata connection placement is somewhat correct, I still think that a better option is to scratch that and just add a metadata refresh button here instead
> 
> image

That works for me, and it is less UI than what I built, which I like.

To make sure I put it in the right place. You mean the shared server settings modal, ServarrSettingsModal, in the button row with Docs and Test Connection, shown only when the server is a Sportarr one. That drops SportarrSection and the Metadata card entirely.

Two things that card carries besides the refresh, so tell me where you want them.

The artwork attribution, the line saying league art comes from the connected Sportarr instance or from sportarr.net. The backdrop badge already carries the logo and the link, so I think that covers it and the line can go.

The sportarr.net indicator, which shows whether SPORTARR_NET=on is set. It only tells you what the environment variable already says, so I am happy to drop it unless you want it visible somewhere.

If you are fine with both going, this ends up as one button in the modal and a smaller diff.

Separately, the backdrop logo is sorted. The asset is now the lockup, mark plus wordmark, so it fills the plate the way the TMDB wordmark does instead of sitting square in the middle. It carries a short intro, the ring draws on, a few sports flicker past, the trophy rises and the word spells out, then it holds. Plain CSS inside the SVG, no script, reduced motion respected, and the resting state is a still lockup, so anyone arriving after it plays sees a normal logo. If you would rather nothing moved in your UI, say so and I will drop in the still version instead, same file, same dimensions.

enoch85:
> To make sure I put it in the right place. You mean the shared server settings modal, ServarrSettingsModal, in the button row with Docs and Test Connection, shown only when the server is a Sportarr one. That drops SportarrSection and the Metadata card entirely.

Yes, that's what I mean :+1: 

> The sportarr.net indicator, which shows whether SPORTARR_NET=on is set. It only tells you what the environment variable already says, so I am happy to drop it unless you want it visible somewhere.

Drop it and add to documentation instead IMHO. If you get a lot of requests regarding it, then we can add it, but I like it clean. :) 

> Separately, the backdrop logo is sorted. 

Yup, looks much better now! Cool animation also!

----------------

Claude will post another review comment here soon. Great progress! :tada:

enoch85:
Claude Opus 4.8:

<details>
<summary>Review of #3606 at 817d64f8 (expand)</summary>

Fourth pass, against the real stack: the live Sportarr 4.1.5 instance, dev Jellyfin, and the compiled classes rather than reading source. Every "fixed" below was re-proved by re-running the original repro, not by reading the diff.

**The outage half of the blocker is genuinely fixed. A second instance of the same latch is still there, on the first-run path.**

## Blocker: the latch survives when the install boots with nothing configured

`sportarr-metadata.service.ts:142`

`hasReachableSource()` is now derived rather than a counter, and that fixes the outage case. But it derives from `configuredSources`, which only `sources()` refreshes, and `sources()` is still only reachable through the provider, which `getOrderedProviders()` gates on `isAvailable()`. So the loop closes again whenever the cached source list is empty.

That is the default first run: Maintainerr starts before the user has configured anything.

```
boot, no connections      : isAvailable = false
connection added in Settings...
  immediately                           : isAvailable = false  requests=0
  after 6s (past the 5s source-list TTL): isAvailable = false  requests=0
  direct getLeague still works          : yes  requests=1
  isAvailable after that walk           : true
```

The last two lines are the proof that it is purely the gate: calling the service directly, bypassing `isAvailable()`, works immediately and un-latches it. Through the provider it never gets the chance.

So **a fresh install that adds its first Sportarr connection gets no league artwork until Maintainerr restarts.** The 5 second source-list TTL never helps, because nothing reads it.

There is an escape hatch, but nothing points at it:

```
boot empty                      : isAvailable = false
connection added                : isAvailable = false
after hasSource() (Refresh card): isAvailable = true
```

`hasSource()` calls `readConfiguredSources()` directly, so Settings > Metadata > Refresh metadata revives it. A user has no way to know that.

The same shape of fix works here as last time - stop letting the gate be the only thing that can open the gate. Either have `hasReachableSource()` re-read when the cached list is empty or stale, or seed `configuredSources` from the settings event that already fires when a connection is added rather than only from `onModuleInit`.

---

## Minor

### The Docker docs contradict the PR's own UI

`dockerhub-README.md:104` ends with "Off by default, so nothing is requested from it."

That is not true, and this PR is the thing that establishes it is not true. `SportarrSection.tsx:88` says it correctly: "Artwork files come from there either way, because that is where Sportarr hosts them." Sportarr mints absolute `https://sportarr.net/static/...` URLs and a local install 404s those paths, so every poster and backdrop a user sees is fetched from sportarr.net by the browser no matter what the variable says.

Two docs, opposite claims, and the wrong one is the one an operator reads while deciding whether they need the variable at all. The README and ARCHITECTURE entries are fine; it is only this sentence.

### The card asserts "off" before it knows

`SportarrSection.tsx:27` - `status?.sportarrNetEnabled ?? false`. While the status query is in flight the card renders the "it is off, set `SPORTARR_NET=on`" copy, and it keeps rendering it forever if the query errors. Someone who has already set it on is told to set it on. Same class as the toggle flicker from the last round, moved from the control into the copy.

### The status shape is declared twice instead of once

`settings.controller.ts:549` returns an inline `{ sportarrNetEnabled: boolean }`, and the UI redeclares the same shape for its hook. Every sibling settings shape in this feature is contracts-owned. Rule 7 asks for deliberate, minimal, contract-owned DTOs; this is the one that got away.

### `SPORTARR_NET=on` is inside the translatable string

`SportarrSection.tsx:88` and `:94` put the literal variable name and value in the message. The domain right beside it correctly rides through as `{sportarrDomain}` so a translation cannot change it - the variable deserves the same treatment. A translator can currently rewrite the instruction into something that does not work.

### The attribution badge takes about three seconds to read "Sportarr"

Measured from the stylesheet inside the asset:

| element | starts | fill |
|---|---|---|
| `.markslide` | 2.2s | `both`, so held at `translateX(1523px)` before that |
| `.letter` | 2.52s + 0.05s per letter | `backwards`, so held at `opacity: 0` before that |
| `.shine` | 3.35s, ends ~4.3s | |

Eight letters puts the last one finishing around 3.3s. Until then the mark is displaced and the wordmark is invisible, and it replays on every media-modal open. The TMDB and TVDB badges beside it are static. An attribution mark exists to answer "where did this artwork come from" at a glance, which it cannot do for the first three seconds.

Worth saying clearly: `prefers-reduced-motion` is handled properly here - it sets `animation: none` and leaves the resting lockup, rather than merely shortening. That part is right.

---

## Nits

- `metadata-settings.service.ts:190` - the refresh toast reads "SPORTARR metadata refresh started". `provider.toUpperCase()` is right for two acronyms and wrong for a product name. Pre-existing line, newly reachable because this PR adds a third provider.
- `mediaTypeUtils.ts:107` - `displayProviderId` parses with `Number()` and re-implements the guard, while `sportarrLeagueNumber` sits in the contracts module it already imports from.
- `sportarr-external-id.ts:26` - `sportarrLeagueExternalIdFromTvdbAlias` tests the alias window, then calls a helper that tests it again.
- The logo is 30987 bytes raw, 11607 gzipped, against tmdb 2065/988 and tvdb 2832/1124. It is detailed rather than unoptimised - 83% is path data already at 1-2 decimal places - and it is served once and cached. Worth knowing rather than fixing.

---

## Verified fixed

Each re-proved with the repro that found it.

**The outage latch.** Was: one failure, `isAvailable()` false for the life of the process, one request ever.

```
lookup 1 (DOWN) : poster=none isAvailable=false axiosCalls=1
lookup 2 (DOWN) : poster=none isAvailable=false axiosCalls=1
--- source is UP again. waiting 65s, past the cooldown ---
lookup 1 (UP, post-cooldown) : poster=yes isAvailable=true axiosCalls=2
RECOVERED: true
```

**sportarr.net is opt-in, and the default really is silent.**

```
no connection, SPORTARR_NET unset (default)   requests=0  (none)
no connection, SPORTARR_NET=off               requests=0  (none)
no connection, SPORTARR_NET=on                requests=1  https://sportarr.net
connection, SPORTARR_NET unset (default)      requests=1  http://sportarr.local:1867
connection, SPORTARR_NET=on                   requests=2  connection, then sportarr.net
```

**One retry for a 429, capped.** Done without touching the shared `applyHttpRetry`, by gating the 429 branch on the attempt count:

```
Retry-After 5s      retry=true  delay=5000
Retry-After 59s     retry=true  delay=59000
Retry-After 61s     retry=false delay=-
Retry-After 600s    retry=false delay=-
429 no header       retry=true  delay=223
500 / network       retry=true  (standard policy preserved)

retryCount=0 -> retry = true      retryCount=1 -> retry = false
```

Worst case for a rate-limited lookup is now one 60s wait instead of three.

**Documentation.** `SPORTARR_NET` is in README.md:83 next to the `TELEMETRY=off` line, dockerhub-README.md:104 in the env table, and ARCHITECTURE.md:193 in the env-var list - which is one more place than `TELEMETRY` itself appears.

**One league-id format.** `LEAGUE_ID_PREFIX` and the `padStart` now exist only in `packages/contracts/src/sportarr/sportarr.constants.ts`. The server helper's private copies are gone.

**Live, with a connection configured and `SPORTARR_NET` at its default:**

```
EPL alias only         -> Sportarr poster, no sportarr.net metadata call
F1 tmdb + alias        -> Sportarr (isAuthorityFor still winning over TMDB primary)
normal movie (control) -> TMDB, unchanged
normal show  (control) -> TMDB, unchanged
```

## Also clean

No em or en dashes or smart quotes in any added line or commit message. Catalogs valid, extraction current, new msgstr empty in all 13 locales, nothing left as an obsolete entry, and the card reuses the parameterised sibling messages rather than minting new ones. Server and UI `tsc` clean. Full suite green: **2938 server tests** across 3 shards, **404 UI tests** across 2.

</details>

Sportarr:
> Claude Opus 4.8:
> 
> Review of #3606 at [817d64f](https://github.com/Maintainerr/Maintainerr/commit/817d64f8df6cb3d1566af7badd758052e705c350) (expand)

Done. Refresh Metadata is in the server dialog now, next to Test Connection, and the Metadata card plus its status endpoint are gone.

The availability gate re-reads itself now, with a test for that exact sequence.

Also fixed the Docker overview line. It said nothing is requested from sportarr.net, but Sportarr mints absolute artwork URLs there so the browser loads images either way. Docs PR updated too.

Leaving the animation as it is. I too think it is cool af lol!

enoch85:
Claude Opus 4.8:

<details>
<summary>Review of #3606 at e4c95341 (expand)</summary>

Fifth pass, against the real stack: the compiled service, a fault-injection stand-in for the metadata routes, Playwright, and the running dev server. Every "fixed" below was re-proved by re-running the original repro. All 13 checks are green.

**The review-4 blocker is genuinely fixed. The fix that closed it opened a cost regression on installs that do not use this feature.**

## Major: the gate now reads SQLite on every call, for every install without Sportarr

`sportarr-metadata.service.ts:145`

```ts
if (!cached || cached.sources.length === 0 || now - cached.readAt >= SOURCE_LIST_TTL_MS) {
  void this.refreshConfiguredSources();
}
```

For an install with no Sportarr connection, `readConfiguredSources()` stores `{ sources: [], readAt: now }`, so `cached.sources.length === 0` is permanently true and the 5 second TTL beside it never gets a say. Measured against the compiled service, 50 gate checks inside one TTL window:

```
no Sportarr configured (the default install)  boot=1  +50 gate checks => 50 settings reads
one connection configured                     boot=1  +50 gate checks =>  0 settings reads
```

Zero when the feature is in use, one per call when it is not. That is backwards, and `isAvailable()` is not a cold path: `metadata.service.ts` reaches `getOrderedProviders()`/`isAvailable()` from 12 places, several inside per-item loops (`:295`, `:590`, `:1140`, `:1150`, `:1226`, `:1271`).

Distinguishing "not configured" from "not read yet" fixes it: an empty list that was read 200ms ago is a fresh answer, not a missing one. Only `!cached` and the TTL should trigger the re-read.

## Minor: the fire-and-forget refresh has nothing to catch it

`sportarr-metadata.service.ts:154` calls `void this.refreshConfiguredSources()`, and `settings-data.service.ts:462` does this:

```ts
public async getSportarrSettings() {
  try {
    return this.sportarrSettingsRepo.find();   // not awaited
  } catch (error) { ... }
}
```

`find()` is returned, not awaited, so its rejection never reaches that catch. The service's own comment ("A failed settings read answers a status object rather than throwing") is true for a synchronous throw and false for a locked database. Proven against the compiled service:

```
onModuleInit threw (caught): SQLITE_BUSY: database is locked
UNHANDLED REJECTION -> SQLITE_BUSY: database is locked
  at SportarrMetadataApiService.readConfiguredSources
  at SportarrMetadataApiService.refreshConfiguredSources
  at SportarrMetadataApiService.hasReachableSource
```

It does **not** crash: `main.ts:125` catches it and deliberately does not exit. What the user gets is an ERROR reading "An unhandledRejection has occurred. This is likely a bug, please report this issue on GitHub" for a routine `SQLITE_BUSY` that the repo handles gracefully everywhere else. The missing `await` is pre-existing on development; the un-caught `void` caller is what makes it reachable, and the major above is what makes it reachable on every call.

## Minor: the two strings this PR adds to 13 catalogs can never render

`ServarrSettingsModal.tsx:271`

```ts
message: response?.message ?? (response?.code === 1
  ? t`${{ serviceName }} metadata refresh started`
  : t`Failed to refresh ${{ serviceName }} metadata`)
```

The server always sends a message, so the localised branch is dead. `metadata-settings.service.ts:190` builds it as `${provider.toUpperCase()} metadata refresh started`. Live, in the dialog:

> SPORTARR metadata refresh started

The spec passes anyway, because it mocks a response the server never sends:

```ts
postApiHandler.mockResolvedValue({ code: 1, message: 'Sportarr metadata refresh started' })
...
expect(screen.getByText('Sportarr metadata refresh started')).toBeTruthy()
```

That is the trap `project-notes.instructions.md:343` already warns about in another context: a spec that fabricates a shape production does not produce passes while production takes the other branch. Prefer the localised string for the known codes and keep the server message for the cases that carry a reason (`:161` "already in progress" is shouty and untranslated too).

## Minor: Refresh Metadata is live in the empty Add-server dialog

Open Settings > Sportarr > Add server and the footer offers **Refresh Metadata**, enabled, beside a correctly disabled **Save Changes**. Clicking it on a blank form flushes the whole Sportarr metadata cache and answers "SPORTARR metadata refresh started". `refreshMetadata()` guards only on `!metadataRefreshPath || refreshing`.

The card this replaced had the guards (`Save to refresh`, `Configure to refresh` in `Metadata/index.tsx:112-140`). Moving the control into the dialog dropped them.

## Minor: a second msgid for a label the repo already has

```
msgid "Refresh metadata"     #: src/components/Settings/Metadata/index.tsx
msgid "Refresh Metadata"     #: src/components/Settings/Servarr/ServarrSettingsModal.tsx
```

Same words, different casing, two entries for 13 locales to translate and to drift apart.

## Minor: an orphaned comment landed on the telemetry endpoint

`settings.controller.ts:545` - removing the Sportarr status endpoint left its comment behind, now sitting above `@Get('/telemetry/status')`:

```ts
// Whether the environment lets Sportarr artwork fall back to sportarr.net.
// The card reads this rather than telling every install to set a variable
// it may already have set.
@Get('/telemetry/status')
```

## Nits

- The gate is one call behind by design. Measured: after a connection is added, a synchronous burst of 8 cards all miss; with any await between checks it opens on the next one. Same on outage recovery (1st call false, 2nd true). Documented in the code, worth knowing it means the first lookup after configuring still misses.
- The attribution badge still takes ~3.3s to read: letters at `2.52s + 0.05s` each, gleam at `3.35s`. Unchanged from the last round. `prefers-reduced-motion` is still handled properly.

## Verified fixed

**The first-run latch.** Was: no league artwork until restart.

```
boot, nothing configured        : hasReachableSource = false
connection added, no restart    : gate opened, no restart needed
getLeague after the gate opened : Mock League Alpha | http calls = 1
```

**The retry override is gone and the shared policy took over cleanly.** Measured through the compiled Sportarr client against a 429/503 stand-in:

```
healthy                             10ms  attempts=1  league=yes
429 Retry-After 5                 5009ms  attempts=2   (one wait of exactly what was asked)
429 Retry-After 70 (over cap)        2ms  attempts=1   (over the cap, no retry)
503 Retry-After 30                1503ms  attempts=4   (header ignored, curve kept)
```

Identical to every other client. No leftovers: `sportarrMetadataRetryPolicy`, `retryCountOf` and `ExternalAPIOptions.retry` are gone tree-wide.

**The docs contradiction.** `dockerhub-README.md:104` now ends "Sportarr hosts the artwork files themselves on sportarr.net, so a browser still loads images from there either way." Correct, and it matches what the UI used to claim.

**Three findings went away with the card.** `SportarrSection.tsx`, the `sportarrNetEnabled` endpoint and its duplicated shape, and the env var inside a translatable string are all removed. Grepped tree-wide: no leftovers.

**The alias helper stops double-testing the window.**

## Retraction

My review-4 nit that `displayProviderId` should use `sportarrLeagueNumber` was wrong. That helper parses `lg-000278` into a number; `displayProviderId` takes a bare number and builds the `lg-` id, which is the opposite direction. `Number(providerId)` plus the guard is correct, and `index.spec.tsx:308` shows why it exists.

Also worth correcting from review 3: the Sportarr connection CRUD emitting no event is not a Sportarr shortcut. `settings-operations.service.ts` emits nothing for any servarr connection, so the TTL poll is a reasonable answer to an existing gap rather than a corner cut here.

## Also clean

No em or en dashes, smart quotes, or attribution in any added line or commit message. No underscore-prefixed identifiers. Catalogs valid across 13 files, a fresh extract changes nothing, and every new msgstr is empty. No dead exports in `sportarr.constants.ts` - all seven have real consumers. Server and UI `tsc` clean, lint clean, full suite green.

</details>

Sportarr:
> Claude Opus 4.8:
> 
> Review of #3606 at [e4c9534](https://github.com/Maintainerr/Maintainerr/commit/e4c9534114a07d2af5d0a8b719ca5efdd5ed2e49) (expand)

All five fixed.

enoch85:
Claude Opus 4.8:

<details>
<summary>Review of #3606 at 2d22fc2d (expand)</summary>

Sixth pass, against the real stack: the compiled service, a stand-in for the `/api/metadata/agents` routes and for a 429/503 upstream, Playwright, and the running dev server. Every fix below was re-proved by re-running the repro that found it.

**All six findings from the last round are fixed. Two of the defects this round were mine.**

## What I broke, and why my own gate missed it

My last commit dropped `!metadataRefreshPath` from the refresh guard as redundant. It is redundant to a reader and not to the compiler: it is what narrows the prop for the post two lines below.

```
ServarrSettingsModal.tsx(268,9): error TS2345: Argument of type 'string | undefined'
  is not assignable to parameter of type 'string'.
```

That failed **Type checking** and took **Run integration tests** with it, because `turbo test` depends on `check-types`. Restored, with a comment saying why it cannot be folded into `canRefresh`.

The reason I shipped it: my local gate ran `tsc` for the server workspace only, while CI runs `yarn check-types` across all three. A server-only type check cannot see a UI error. My harness now runs the same command CI does, which reproduces the failure locally and now passes.

## Naming pass

You were right that `refreshPath` was worse than `metadataRefreshPath` - it drops the qualifier that says *which* refresh. Reverted, and I went through every identifier this PR introduces.

One that was worth changing:

| before | after | why |
|---|---|---|
| `hasSource()` | `hasConfiguredSource()` | Its own docstring says "whether anything is configured to read from", which is exactly what separates it from `hasReachableSource()` sitting beside it. The old name read like the general case, and the pair was confusable enough that I flagged it in the third review |

Checked and deliberately left alone: `metadataRefreshPath`, `canRefresh`, `LeagueAnswer` (matches the file's "answers" voice), `seriesUrl` (Sportarr's own agent route calls a league a series), `sources()`, `configuredSources`, `unreachableUntil`, and the contracts helpers (`sportarrLeagueId`, `sportarrLeagueNumber`, `sportarrLeagueNumberFromTvdbAlias`, `isSportarrTvdbAlias`) - all say what they hold.

## One stale comment, now that #3614 has landed

`readConfiguredSources` said the settings read "answers a status object, except a locked database, which rejects". #3614 is in development and merged into this branch, so it no longer rejects. The comment now says what the catch is actually there for: the gate calls this without awaiting, so an error has no caller to land in.

I left the catch itself. It is belt-and-braces today, and I am not going to remove a second thing that "cannot fire" in the same review.

## Verified fixed, each re-measured

| finding | before | after |
|---|---|---|
| gate read SQLite per call without Sportarr | 50 checks, **50 reads** | 50 checks, **0 reads** |
| locked database raised an unhandled rejection | `exit=1`, trace through `hasReachableSource` | `exit=0`, `survived` |
| Refresh live in the Add-server dialog | enabled, fired a real cache flush | `[disabled]`, confirmed in Playwright |
| refresh message | `SPORTARR metadata refresh started` | `Sportarr metadata refresh started` |
| duplicate msgid | `Refresh metadata` + `Refresh Metadata` | one entry |
| orphaned comment on `/telemetry/status` | present | removed |

## Behaviour the cost fix changed, and why it is right

**First run now takes one TTL instead of one call.** Dropping the empty-list trigger means a new connection is noticed on the next TTL rather than the next gate check.

```
boot, nothing configured : gate = false
connection added         : gate opened = true after 5076ms
lookup                   : Mock League Alpha
```

Five seconds for the person who just clicked Save, nothing at all for the majority who never configure Sportarr. No latch, which was the original blocker.

**A locked database fails closed for one TTL and then recovers.**

```
boot while the db is locked : gate = false
db recovered, immediately   : gate = false
after the TTL expires       : gate = true
and a real lookup works     : Mock League Alpha
```

**The retry policy is unchanged by the merge.** Through the compiled Sportarr client against a 429/503 stand-in:

```
healthy                          10ms  attempts=1
429 Retry-After 5              5008ms  attempts=2   (one wait of exactly what was asked)
429 Retry-After 70 (over cap)     2ms  attempts=1   (over the cap, no retry)
503 Retry-After 30             1612ms  attempts=4   (header ignored, curve kept)
```

## Also clean

No em or en dashes, smart quotes, or attribution in any added line or commit message. Catalogs valid across 13 files, a fresh extract changes nothing, new msgstr entries empty. No dead exports in `sportarr.constants.ts`. Full suite: **2941 server tests** across 3 shards, **406 UI** across 2, `yarn check-types` and lint clean across all three workspaces.

</details>

enoch85:
@Sportarr I'm about to merge this. Just doing some final checks.

Thanks for your hard work, and being so nice to work with!

Sportarr:
Thank you for your support, fixes, and patience! Excited for the users to test it out.

PR #3636, opened 2026-09-01
fix(collections): stop recording an unanswered media server write as a refusal

enoch85:
Reported on Discord: every item in one large Plex TV library was added to its collection as **MANUAL**, so rules could never remove it, while the same items had already gone from the Plex collection.

## Cause

A collection mutation that times out is treated as a refusal. Plex commits a write it has begun processing and can answer later than the 30s client timeout, so the change is applied while Maintainerr records that it did not happen.

| Path | On "failed" | Result |
|---|---|---|
| Add | `continue` before the membership insert | no row; Plex holds a child nothing local accounts for |
| Manual child import | child has no row | adopts it as a hand-added member |
| Rules | manual-only rows are skipped by `removalScope: 'rule'` | never removable |
| Worker | due query filters on `addDate` only | deleted from disk after `deleteAfterDays` |

Measured against a real Plex 1.43.3: once it has begun processing a collection write it commits ~100% of the time (60 of 76 aborted trials; 0/16 only when the abort landed within 1ms). "No answer" is the one outcome that must never be read as "nothing changed".

This is the same shape as #3228/#3229, which fixed it for Sonarr and Radarr. The read side has been hardened against an absent answer four times (#3211, #3248, #3344, #3413); the write side had not been.

## Change

| Area | Change |
|---|---|
| `media-server.interface.ts` | batch mutators return `{refused, unknown}` instead of one failed list |
| all three adapters | classify: 4xx = refused, 5xx or no response = unknown |
| `collections.service.ts` | an unanswered add records a marker with the new `direction: 'add'` |
| reconciler | holds an unconfirmed add out of adoption; never self-heal-removes it |
| Plex adapter | no per-item retry when nothing answered (2 items cost 1 timeout, not 3) |
| empty-collection heal | keys on genuine refusals, so pure timeouts cannot arm it |
| Emby adapter | singular add/remove threw away the batch result and reported failures as success |
| bulk messages | a timeout is no longer reported as "refused by the media server" |

Migration adds one column defaulting to `'remove'`, so existing markers keep their meaning. `collection_media_rule_removal`'s reconciler, cascade, chunking and trustworthiness check are reused rather than duplicated.

`AGENTS.md` gains a fifth guard and the corrected count: this bug class has now been fixed twelve times, and it regenerates because the adoption decision defaults to adopting and subtracts cases computed from `collection_media` rows, which cannot cover a bug whose symptom is a *missing* row.

## Verified

Against the real dev Plex, with a proxy that forwards the mutation and holds the response past the client timeout.

Before: `Importing 2 item(s) ... as manual member(s)`, both rows `isManual=1`, three timeouts.

After:

```
Plex:        4 children  [4, 1, 3, 2]      the adds committed
Maintainerr: 2 rule-owned rows [1, 4]      no rows for 3 and 2
markers:     [('2','add'), ('3','add')]
MANUAL rows: none
```

Then a normal run converges: all four rows `includedByRule=1`, markers cleared. Also checked the case where the rule stops matching while a marker is outstanding, which is where the previous in-run guard (#2440) failed and had to be replaced by persistence (#3298): no adoption, marker held.


enoch85:
All three P2s reproduced and fixed in ad88fb7a. Thanks - two of them were the same class this PR exists to fix, which is exactly the wrong way for it to be wrong.

**1. Manual intent (collections.service.ts:4316).** Confirmed: the condition tested `collectionIds.manualCollection` (the collection's kind) and ignored `manual` (the membership intent). A manual add to an automatic collection that timed out wrote a marker, the next run held the child out of adoption, and since no rule matches it nothing would ever re-add it - the user's add doing nothing, permanently. Adoption is precisely what fulfils a manual add, so markers are now written only for rule-driven ones.

**2. Stale direction (collections.service.ts:500).** Reproduced your result before changing anything:

```
insert or ignore (1,'x','remove') then (1,'x','add')
  -> [ { collectionId: 1, mediaServerId: 'x', direction: 'remove' } ]
```

Now an upsert (`.orUpdate(['direction'], ['collectionId','mediaServerId'])`). Added `collection-media-rule-removal.integration.spec.ts` - a real in-memory SQLite DataSource through TypeORM, following the EntitySchema pattern from the exclusion-scoping integration test - because a mocked repository cannot see what a conflict does. I checked it fails with `.orIgnore()` and passes with the upsert:

```
- "direction": "add"
+ "direction": "remove"
```

**3. Single-item endpoint (collections.service.ts:3387).** Confirmed: `MediaCollectionActionWithContext` merged `serverRejectedIds` and `unpersistedIds` and dropped `serverUnconfirmedIds`, so `POST /api/collections/media/add` answered 201. Now propagated, and the controller answers 502 with wording that does not claim a refusal, since the server never answered.

Three regression tests, each verified failing against the previous commit. Full server suite green (131 suites, 2931 tests).

On the E2E note: that was me - I moved the workspace to `development` to branch the follow-up PRs, which recompiled the running server out from under you. Sorry. The mutation harness is `.dev-repro/` (a proxy that forwards a Plex write, lets it commit, then holds the response past the 30s client timeout); `verify.sh` and `case-b.sh` re-run the add/adopt and rule-stops-matching scenarios. I have not re-run them against these three changes - the marker upsert is covered by the real-DB spec, but the manual-add path is worth a live pass before merge.

enoch85:
Both items done.

## The ineffective test (334cda09)

You were right, and it is the exact failure a regression test exists to rule out. The spec copied the query chain, so it proved the chain rather than the code — and my earlier "verification" edited that copy, which proved nothing about production.

It now calls `CollectionsService.markRuleRemoved` against a real SQLite repository. Checked by reverting production to `.orIgnore()`:

```
-     "direction": "add",
+     "direction": "remove",
Tests: 1 failed, 3 passed
```

Getting there meant dropping the `.into(CollectionMediaRuleRemoval)` from `markRuleRemoved`. Naming the entity class made TypeORM resolve it against the DataSource, which would have meant registering the whole relation graph in the test. It was redundant anyway — the builder comes from the marker repository — and the emitted SQL is unchanged:

```
ON CONFLICT ( "collectionId", "mediaServerId" ) DO UPDATE SET "direction" = EXCLUDED."direction"
```

## Live pass, on the PR head

Rebuilt the workspace from `334cda09` first — you were right that the running server was still `development`. Fresh database, real dev-plex behind the proxy that forwards the write, lets Plex commit, then holds the response past the 30s client timeout. 4 timeouts recorded.

**1. Rule-driven timeout and convergence.** I also had to fix the harness: its `idle()` polled `/rules/execute/status`, which does not reflect a queued run, so phases were overlapping and the earlier slow phase recorded 0 timeouts. It now counts the executor's own "done" lines.

| phase | rows | markers | MANUAL |
|---|---|---|---|
| 1 fast | `[1, 4]` rule-owned | none | none |
| 2 slow | `[1, 4]` unchanged | `[(2,add), (3,add)]` | none |
| 3 fast | `[1, 2, 3, 4]` rule-owned | cleared | none |

**2. Delayed manual add, single-item API.** `POST /api/collections/media/add {mediaId: 2693}`, an item the rule does not match:

```
HTTP 502  "The media server did not answer for 1 of 1 item(s), so it is unclear whether they were added"   (30s)
markers   none                      <- the manual add is not suppressed
plex      committed (2693 present)
after sync  2693  isManual=1  includedByRule=0  source=local
```

That is finding 1 end to end: no marker, so adoption fulfils the request rather than holding it out forever.

**3. Same flow through the shared modal (Playwright, bulk endpoint).** Selected the card, Add/Exclude selected, "Add to collection", Submit:

```
POST /api/collections/media/bulk -> 201
{"mediaId":"2693","code":0,
 "message":"Failed - the media server did not answer in time, so it is unclear whether this was added"}
```

Not "refused by the media server". The card stayed selected, no marker was written, Plex had committed, and the next sync adopted it: `Importing 1 item(s) ... as manual member(s): 'Sintel' (2693)`.

Evidence in `.dev-repro/live-pass-334cda09.log` (git-excluded). Full server suite green: 131 suites, 2932 tests.

Agreed that approval should come from an independent reviewer — I wrote this, so an approval from me carries no signal.

PR #3637, opened 2026-09-01
fix(rules): stop the stale-link resync marking a manual collection's members as manual

enoch85:

## Cause

`collection.manualCollection` is a property of the **collection** - the rule group points at a collection the user made, rather than Maintainerr creating one. It says nothing about how any individual item became a member.

The stale-link resync passed it as the per-**membership** "added by hand" flag:

```ts
collection = await this.collectionService.addToCollection(
  collection.id,
  collMediaData.map((m) => ({ mediaServerId: m.mediaServerId })),
  collection.manualCollection,   // <- membership-level parameter
);
```

Everything the executor passes is already a `collection_media` row, so it can only reach `updateExistingCollectionMediaForAdd`, which writes `manualMembershipSource` on every existing row when `manual` is true. On a manual collection whose link went missing, **every rule-owned row is rewritten as a manual member**. A manual member is force-kept by every later rule run yet still ages into `deleteAfterDays`, so the item can never leave the collection and is eventually deleted from disk.

There is a **mirror case on the same line**: for an *automatic* collection the resync passes `manual = false`, which forces `includedByRule: true` onto a user's manual-only member, silently promoting it to rule-owned. Both are fixed by the change below, because it stops that update running at all.

Both arguments are `boolean`, so the compiler could not catch it. Introduced in `65f551d4` (2023-09-18) and copied verbatim through every later rewrite of the membership model - `174a5cb2` only renamed `plexId` to `mediaServerId`. The only spec covering that branch pinned `manualCollection: false`.

## Reachable when

A manual collection sitting at `mediaServerId = null` with members still in `collection_media`. `checkAutomaticMediaServerLink` cannot repair it - its whole body is gated on `!collection.manualCollection` - and two ordinary UI paths write that state, neither guarded by `manualCollection`:

- opening the collection in the UI: a failed `getCollectionChildren` plus a confirming probe nulls the link
- saving the rule group or collection settings: `updateCollection` clears the link when the probe reports the server collection missing

So: point a rule group at an existing collection, let it fill, delete or rename that collection on the media server, then touch the collection in the UI. Every rule run from then on re-stamps every member, and writes an "Added" log record per item per run.

Once stamped there is no self-heal: force-keep puts the item back in `desiredMediaServerIds` every run so it is never in `mediaToRemove`, and the manual-child import only ever adopts.

## Change

Pass an empty member list and no flag.

`addToCollectionInternal` re-reads the collection's members from the database and syncs *those* to the freshly created server collection, so the list passed in was redundant. Its only effect was to run every existing row back through the membership update - which re-marked them, and wrote the per-item log record, a cost the comment above the call already noted.

| | before | after |
|---|---|---|
| rule-owned row on a manual collection | rewritten as manual, force-kept, deleted at `deleteAfterDays` | untouched |
| manual-only row on an automatic collection | promoted to rule-owned | untouched |
| "Added" log record per member per run | written | not written |
| server collection recreated and repopulated | yes | yes |

A follow-up commit documents at the signature that an empty list is load-bearing rather than a no-op, so a later cleanup does not remove it and restore the bug.

Regression test added; verified it fails against the unfixed code and passes with it.

### DIFF
--- a/apps/server/src/modules/collections/collections.service.ts
+++ b/apps/server/src/modules/collections/collections.service.ts
@@ -3019,6 +3019,17 @@ export class CollectionsService {
     };
   }
 
+  /**
+   * Add media to a collection, creating or relinking its media server
+   * collection first if that is missing.
+   *
+   * An EMPTY `media` list is meaningful and load-bearing, not a no-op: the
+   * find-or-create still runs, and the existing members are read from the
+   * database and pushed to the collection it creates. That is how the rule
+   * executor rebuilds a link that went missing, without running every current
+   * member back through the membership update - which would re-mark provenance
+   * and write an "Added" log record per item, per run.
+   */
   async addToCollection(
     collectionDbId: number,
     media: CollectionMediaChange[],
--- a/apps/server/src/modules/rules/tasks/rule-executor.service.spec.ts
+++ b/apps/server/src/modules/rules/tasks/rule-executor.service.spec.ts
@@ -1294,11 +1294,9 @@ describe('RuleExecutorService', () => {
     expect(
       collectionService.addToCollectionWithResolvedLink,
     ).not.toHaveBeenCalled();
-    expect(collectionService.addToCollection).toHaveBeenCalledWith(
-      1,
-      [{ mediaServerId: 'm1' }],
-      false,
-    );
+    // No member list and no membership flag: addToCollection re-reads the
+    // collection's members itself, and the resync must not re-mark them.
+    expect(collectionService.addToCollection).toHaveBeenCalledWith(1, []);
     expect(
       collectionService.removeFromCollectionWithResolvedLink,
     ).toHaveBeenCalledWith(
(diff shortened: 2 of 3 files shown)


enoch85:

I traced the stale-link resync through `addToCollection` / `addToCollectionInternal` and the membership-source fields, then compared the blame and regression history around the shared-collection and manual-membership fixes. Passing an empty input here is the minimal correct change: `addToCollection` deliberately reloads the existing `collection_media` rows, while passing those rows as new input was what restamped their provenance.

Verification: focused rule-executor suite passed 51/51, latest CI is green, and the real Plex, Jellyfin, and Emby APIs were exercised for collection reads and mutations.

enoch85:
Re-review complete at final merged head `c6993d1`: no findings.

The final diff still uses the existing resync path to reload persisted memberships without restamping their provenance. I rechecked it against the current collection ownership helpers and the shared-link history; the merged behavior remains consistent with the manual-membership guardrails. Current CI is green, and the repeated real media-server collection checks did not expose a regression.

PR #3638, opened 2026-09-01
fix(collections): report adds that had no media server collection to add to

enoch85:

## Cause

`addToCollectionInternal` gates the whole membership block on a link existing:

```ts
if (newMedia.length > 0 && (collection.mediaServerId || keepInMaintainerrOnly)) {
```

When a collection has no `mediaServerId` and is not kept in Maintainerr only, execution falls straight past it. Nothing is written to the media server, and **no `collection_media` row is written either** - but `rejectedByServer` and `unpersistedIds` keep their empty initialisers, which is exactly the shape of a total success.

Both interactive callers then report success for an add that did nothing:

| Caller | Reports |
|---|---|
| `POST /collections/media/add` | 201 with the collection, so the modal closes as though it worked |
| `POST /collections/media/bulk` | every id defaults to `{ mediaId, code: 1 }` |

## Reachable when

The find-or-create above could not produce a link:

- a manual collection whose `manualCollectionName` no longer resolves on the media server (the code warns and moves on)
- a library search that threw, which deliberately skips the create for that run

## Change

Report it as its own outcome rather than folding it into the refusals - the server was never asked, so it is neither a refusal nor a local persistence failure, and #3636 keeps `serverUnconfirmedIds` separate for the same reason.

- `CollectionAddResult` gains `unlinkedIds`
- the skipped branch records those ids and warns, naming the collection
- `POST /media/add` answers **400** naming the cause, rather than 201
- the bulk endpoint fails those ids with "this collection has no media server collection to add to" instead of silently succeeding

No media-server interface change, no `supportsFeature` branching, no migration.

Regression test added; verified it fails against the unfixed code (`Expected: ["item-1"] Received: undefined`) and passes with it.

### DIFF
--- a/apps/server/src/modules/collections/collections.controller.ts
+++ b/apps/server/src/modules/collections/collections.controller.ts
@@ -436,6 +436,14 @@ export class CollectionsController {
       );
     }
 
+    // Nothing was written locally or remotely, so answering 201 would close the
+    // modal on a no-op. Not a media server refusal: it was never asked.
+    if (result.unlinkedIds?.length) {
+      throw new BadRequestException(
+        'This collection has no media server collection to add to. Check that its collection still exists on the media server.',
+      );
+    }
+
     if (result.serverRejectedIds.length > 0) {
       throw new BadGatewayException(
         `The media server refused ${result.serverRejectedIds.length} of ${result.resolvedCount} item(s)`,
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -1867,6 +1867,37 @@ describe('CollectionsService', () => {
     expect(mediaServer.deleteCollection).not.toHaveBeenCalled();
   });
 
+  it('reports ids it could not add because the collection has no media server link', async () => {
+    // The find-or-create above could not produce a link (an unresolvable manual
+    // collection name, or a library search that threw). Nothing is written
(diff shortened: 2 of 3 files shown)


enoch85:
Correction to the PR body: I described the overlap with #3636 as minor. Checking `development...origin/fix/unconfirmed-collection-writes` properly, five of the six edit sites here sit inside hunks #3636 also rewrites - `CollectionAddResult`, the accumulator declarations, the membership-block gate, the result construction and the bulk reporting branch.

Nothing is wrong with either change, but they will conflict textually. Whichever merges second wants a rebase and a re-run of the collections suite rather than a blind resolve.

enoch85:

The new `unlinkedIds` outcome remains distinct from a media-server refusal, an unanswered write, and a local persistence failure, and it is handled consistently by both the singular and bulk API surfaces. I also checked the overlap with the now-merged unconfirmed-write work after the rebase.

Verification: focused collections service suite passed 171/171 on the change, and the latest-head integration, lint, type, build, and CodeQL checks are green.

enoch85:
Re-review complete at final merged head `b4ce6f0`: no findings.

The final result still distinguishes an unlinked no-op from refusal, unknown write outcome, and persistence failure, and both singular and bulk callers report it as failure. I rechecked the merged code against the now-current collection result shape and UI outcome handling. Current CI and the real backend/API smoke pass are green.

PR #3639, opened 2026-09-01
fix(collections): treat an unset deletion window as never, and stop clearing it

enoch85:
Found while auditing the deletion paths behind #3636. Independent of it; branches off `development`.

## Cause

`getCollectionDangerDate` coerces a null `deleteAfterDays` to `0`:

```ts
new Date(Date.now() - +(deleteAfterDays ?? 0) * 86400000)
```

so the danger date is *now*, and the worker's due query (`addDate <= dangerDate`) matches **every member of the collection at once**. The configured action then runs against all of it, and the default action is DELETE.

Every other consumer reads the same null as "never":

| Consumer | Reads null as |
|---|---|
| `getCollectionDeleteDate` (contracts) | no delete date |
| collection card | "Never" |
| overlay processor | no countdown drawn |
| **collection worker** | **0 - everything due now** |

The worker is the outlier, and the only one that deletes files.

## Reachable when

**Not** through the rule form - a `superRefine` requires the field unless the action is Do Nothing or Change Quality Profile, and submit only omits it for those two. (An earlier draft of this report said otherwise; that was wrong.)

It is reachable through the API. `POST /api/rules` is unvalidated (`@Body() body: RuleGroupDto`, no Zod pipe), and `updateRules` reset a saved window to null whenever an update omitted the `collection` block:

```ts
visibleOnRecommended: params.collection?.visibleOnRecommended ?? dbCollection?.visibleOnRecommended,
visibleOnHome:        params.collection?.visibleOnHome        ?? dbCollection?.visibleOnHome,
deleteAfterDays:      params.collection?.deleteAfterDays      ?? null,   // <-
```

The two fields above it fall back to the saved value, under a comment saying an omitted block must keep them. The one field that decides whether files are deleted did not.

## Change

Both sides:

- the worker skips a collection with no window, alongside the existing Do Nothing skip, and says so in the handler summary
- `updateRules` preserves the saved `deleteAfterDays` like its neighbours (the create path still defaults to null, which is correct)

Two regression tests, both verified failing against the unfixed code. The second one's failure output shows the exact hazard: `deleteAfterDays: null` alongside `arrAction: 0`.

### DIFF
--- a/apps/server/src/modules/collections/collection-worker.server.spec.ts
+++ b/apps/server/src/modules/collections/collection-worker.server.spec.ts
@@ -122,6 +122,66 @@ describe('CollectionWorkerService', () => {
     expect(collectionHandler.handleMedia).not.toHaveBeenCalled();
   });
 
+  it('should not handle media for a collection with no deletion window set', async () => {
+    // A null window is "never" to the contracts helper, the collection card and
+    // the overlay processor. The due query read it as 0, which made every member
+    // due at once and handed the whole collection to a DELETE action.
+    const collection = createCollection({
+      arrAction: ServarrAction.DELETE,
+      deleteAfterDays: null,
+    });
+    const collectionMedia = createCollectionMedia(collection);
+
+    collectionRepository.find.mockResolvedValue([collection]);
+    collectionMediaRepository.find.mockResolvedValue([collectionMedia]);
+
+    await collectionWorkerService.execute();
+
+    expect(collectionHandler.handleMedia).not.toHaveBeenCalled();
+  });
+
+  it('should still handle a quality profile collection that has no window', async () => {
+    // The rule form hides the timer for CHANGE_QUALITY_PROFILE and saves no
+    // window, so skipping every null-window collection would strand every such
+    // rule forever and no profile change would ever reach the *arr.
+    const collection = createCollection({
+      arrAction: ServarrAction.CHANGE_QUALITY_PROFILE,
+      deleteAfterDays: null,
+    });
+    const collectionMedia = createCollectionMedia(collection);
+
+    collectionRepository.find.mockResolvedValue([collection]);
+    collectionMediaRepository.find.mockResolvedValue([collectionMedia]);
+    collectionHandler.handleMedia.mockResolvedValue('handled');
+
+    await collectionWorkerService.execute();
+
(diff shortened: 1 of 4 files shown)


enoch85:
Worth stating explicitly, since it is a behaviour change beyond the reported bug: the skip covers **every** action, not just DELETE. A collection with no window set and an action of UNMONITOR, UNMONITOR_SHOW_IF_EMPTY, DELETE_SHOW_IF_EMPTY or CHANGE_QUALITY_PROFILE also stops being handled.

That is the intended reading - "no deadline" means nothing is ever due, which is what the collection card, the contracts helper and the overlay processor already tell the user - but it is a wider change than the title suggests and should be reviewed as such.

enoch85:
**[P1] Do not skip quality-profile actions that intentionally have no window**

The new guard in `collection-worker.service.ts:215` returns before due-item selection for every collection whose `deleteAfterDays` is null. The current rule form intentionally hides the timer for `CHANGE_QUALITY_PROFILE` and sends `deleteAfterDays: undefined` (`AddModal/index.tsx:1074-1078`). Existing handler coverage at `collection-handler.spec.ts:531` also establishes that this action is supposed to call the Servarr handler and then remove the item from the collection.

As a result, every newly created Change Quality Profile rule reaches this worker with a null window and is skipped forever, so no Radarr, Sonarr, or Sportarr profile change occurs. A rule converted from another action can behave differently because the other change in this PR preserves its previous window, making this history-dependent.

Reproduction: create a rule with Change Quality Profile, run the rule so it owns media, then run collection handling. The worker logs the new no-window skip and never calls `collectionHandler.handleMedia`.

Please exempt `ServarrAction.CHANGE_QUALITY_PROFILE` from this guard, or give that action an explicit immediate representation, and add a worker regression covering the null-window quality-profile case.

enoch85:
Re-review complete at `c9391d8`: the prior P1 is resolved and I found no new findings.

`effectiveDeletionWindow` now gives `CHANGE_QUALITY_PROFILE` an explicit immediate window of 0 while preserving null as never for every other action. It also ignores a vestigial window after converting an older rule, so new and converted rules behave consistently.

Focused latest-head verification passed 43/43 tests, including null-window quality-profile handling and the converted-rule case. Current CI is green.

PR #3640, opened 2026-09-01
fix(plex): drop the cached collection children after a collection mutation

enoch85:

## Cause

`plexApi.getCollectionChildren(collectionId, useCache = true)` reads through the 5-minute `plexguid` cache. The adapter never passes `false`, and `collections.service.ts` never imports `cacheManager` at all - so the only thing that ever cleared it was the `cacheManager.flushAll()` at the top of each rule group's run.

Plex is alone in this:

| Server | Caches the children read | Invalidated by a mutation |
|---|---|---|
| Plex | yes, 5 min | **no** |
| Jellyfin | yes | yes, after every collection mutation |
| Emby | no | n/a |

The rule executor's membership sync compares that child list against a **freshly read** database snapshot, so a stale list is taken as the server's current truth. This is not hypothetical: a cached children read already produced phantom manual members once, in #1446 ("When syncing manual added/removed, we were not bypassing the cache when getting the items in the Plex collection. If cached items were returned then that could cause incorrect add & removes").

## Change

- `PlexApi.invalidateCachedUri(fragment)` drops every cached entry whose request uri contains the fragment. `queryAll` caches one entry per page and the key is the serialized request options, so a paginated read leaves several keys that all carry the uri; matching on the uri clears them together.
- `PlexApiService` calls it after `addChildToCollection`, `addChildrenToCollection` and `deleteChildFromCollection` - mirroring where the Jellyfin adapter does it, so every caller benefits.

Two details worth noting:

- Including the trailing `/children` in the fragment keeps collection `24` from matching collection `241`.
- Invalidated on the **failure** path too. A write that timed out may still have been applied - the premise of #3636 - so the cached list is no more trustworthy after a failure than after a success.

Regression tests added, verified failing against the unfixed code (`Number of calls: 0`). One existing test stubbed `plexClient` without the new method; the stub was incomplete rather than the code being wrong, so the stub is completed.

### DIFF
--- a/apps/server/src/modules/api/lib/plexApi.ts
+++ b/apps/server/src/modules/api/lib/plexApi.ts
@@ -222,6 +222,22 @@ class PlexApi {
     }
   }
 
+  /**
+   * Drop every cached response whose request uri contains `fragment`.
+   *
+   * queryAll caches one entry per page, and the cache key is the serialized
+   * request options, so a paginated read leaves several keys that all carry the
+   * uri. Matching on the uri is the only way to clear them together.
+   */
+  invalidateCachedUri(fragment: string): void {
+    const stale = this.cache.data
+      .keys()
+      .filter((key) => key.includes(fragment));
+    if (stale.length > 0) {
+      this.cache.data.del(stale);
+    }
+  }
+
   private serializeCacheKey(params: Record<string, unknown>) {
     try {
       return `${JSON.stringify(params)}`;
--- a/apps/server/src/modules/api/plex-api/plex-api.service.spec.ts
+++ b/apps/server/src/modules/api/plex-api/plex-api.service.spec.ts
@@ -142,6 +142,63 @@ describe('PlexApiService.getMetadata', () => {
     );
   });
 
+  // A membership decision made from a stale child list has already produced
+  // phantom manual members once (#1446), and nothing invalidated this cache.
+  it('drops the cached child pages after a collection add, and after one that failed', async () => {
+    const invalidateCachedUri = jest.fn();
+    (service as any).plexClient = {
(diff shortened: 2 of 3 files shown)


enoch85:
**[P1] Invalidate the children cache when the batch add has an unknown outcome**

The singular add invalidates in both the success and catch paths, but the production batch path `addChildrenToCollection` only invalidates after a successful `putQuery`. Its catch at `plex-api.service.ts:1554` returns without clearing the cached child listing.

I reproduced this against the real development PMS 1.43.3 through the timeout proxy:

- cached a newly created collection with 0 children
- sent the batch PUT; Plex committed it and answered 200 upstream
- held the response beyond the client timeout
- cached read still returned 0 children, while an uncached read returned 1

That is the exact uncertain-write case described by this PR, and the batch method is the path used by collection synchronization. The stale snapshot can immediately drive the same wrong membership/adoption decisions this fix is meant to prevent.

Please call `invalidateCollectionChildrenCache(collectionId)` in the batch catch path as well, and cover a rejected batch request where the subsequent read must miss cache. The temporary live collection was removed and the proxy was returned to normal mode.

enoch85:
Re-review complete at `43884b6`: the prior P1 is resolved and I found no new findings.

The batch catch now invalidates the same child-cache URI as the success and singular paths, with direct regression coverage.

I repeated the real PMS 1.43.3 timeout reproduction on this exact head. Plex committed the delayed batch add and returned 200 upstream; the client reported the expected NOK/unknown outcome; the next cached read returned 1 child instead of the former stale 0. The focused suite passed 106/106. The temporary collection was removed, its 404 was verified, and the proxy is back in normal mode.

PR #3641, opened 2026-09-01
fix(collections): count a custom collection as a sibling when guarding adoption

enoch85:

## Cause

`getSiblingMedia` hardcoded `manualCollection: false`, so "sibling" meant "sibling **automatic** collection" regardless of the subject's kind. Both adoption guards built on it returned nothing for a custom collection sharing the same media-server collection:

- `getSiblingRuleOwnedMediaServerIds`
- `getSiblingMemberMediaServerIds`

An automatic collection that title-links onto a user's custom collection therefore adopts that collection's members as its own **manual** members, which the worker ages into *this* rule's `deleteAfterDays` and hands to *this* rule's action. `skipManualChildImport` does not prevent it - it only suppresses the run that establishes the link, as its own comment says. The same blind spot lets the rule-removal self-heal pull an item out of the shared collection that the custom rule group still holds.

## Why this is a gap and not a decision

The filter came from `98b96e8f` (#2766). Its report, #2765, was explicitly two same-titled **automatic** rule groups - "When two different rules target the same (non-manual) collection" - and the PR body says it applied the approach "to the automatic-collection same-title case it didn't cover". It scoped the fix to the case it was fixing; it did not decide that custom siblings should be ignored.

Kind is irrelevant to this question: a custom collection's member is as foreign to this rule as an automatic sibling's is.

**Not** the same question as `isMediaServerCollectionShared`, which compares kinds on purpose (with a comment saying so) to decide delete-versus-leave on teardown. That one is left alone.

## Change

One line: drop the `manualCollection: false` predicate, plus the doc comment that described siblings as automatic-only, and a note on the query saying why kind is irrelevant here and pointing at the function that legitimately does compare kinds.

The existing spec asserted the old scope (`excludes manual sibling collections`); it now asserts the corrected contract, and I verified it fails against the unfixed code.

### DIFF
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -615,6 +615,84 @@ describe('CollectionsService', () => {
     expect(Array.from(result)).toEqual(['rule-owned']);
   });
 
+  // Now that a custom collection's members are no longer adopted, an automatic
+  // collection can genuinely drain to zero while a custom one still points at
+  // the same media server collection - and that one is the user's own.
+  it('does not delete a media server collection a custom collection still points at', async () => {
+    const collection = createCollection({
+      id: 12,
+      mediaServerId: 'remote-collection',
+      manualCollection: false,
+    });
+
+    collectionRepo.findOne.mockResolvedValue(collection);
+    collectionMediaRepo.find
+      .mockResolvedValueOnce([
+        createCollectionMedia(collection, { mediaServerId: 'item-1' }),
+      ])
+      .mockResolvedValue([]);
+    collectionRepo.save.mockImplementation(
+      async (value) => value as Collection,
+    );
+    // The only other collection on this media server collection is a custom
+    // one, so the same-kind question answers "not shared".
+    collectionRepo.count.mockResolvedValue(1);
+    jest
+      .spyOn(service, 'isMediaServerCollectionShared')
+      .mockResolvedValue(false);
+    jest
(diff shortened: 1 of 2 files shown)


enoch85:
**[P2] Keep the link when the sibling lookup is unknown**

`isMediaServerCollectionLinkedElsewhere` returns `true` when `collectionRepo.count` throws (`collections.service.ts:410-416`). The caller treats that value as a confirmed sibling and immediately saves `mediaServerId: null` at lines 3704-3708.

So a transient failed lookup avoids the remote delete but still drops the only local link. That orphans the media-server collection and revokes the shared-`mediaServerId` sibling guard the collection logic relies on. This is reproducible with `collectionRepo.count.mockRejectedValue(...)` and a successful `collectionRepo.save`: the remote collection remains while this collection is unlinked.

An unanswered lookup needs a third state. Please throw/propagate it, or return an explicit unknown result that makes the caller leave the collection unchanged and retry later. Add that failure-path regression alongside the confirmed shared/non-shared cases.

enoch85:
Re-review complete at `11423da`: the prior P2 is resolved and I found no new findings.

The sibling lookup now returns an explicit unknown state on repository failure, and the empty-collection caller only unlinks or deletes on confirmed true/false. The unknown branch leaves the link intact for retry. The shared ownership lookup also consistently includes custom and automatic collections where kind is irrelevant.

Focused latest-head verification passed 173/173 tests, including the failed-count lookup. Current CI is green.

PR #3642, opened 2026-09-01
refactor(rules): ask a capability, not the server type, about empty child lists

enoch85:

## Cause

The membership sync branched on the configured server to decide two things:

```ts
const isJellyfin = this.settings.media_server_type === MediaServerType.JELLYFIN;
const isEmby     = this.settings.media_server_type === MediaServerType.EMBY;
...
reconcileRuleRemovedOrphans(collection, children ?? [], new Set(), !(isJellyfin || isEmby));
const shouldCheckRemovals = isJellyfin || isEmby ? children && children.length > 0 : true;
```

Both are asking one question: **is an empty child list a trustworthy "this collection is empty", or possibly a sync-delay artefact?** That is a property of the media server, and the guardrails in `ARCHITECTURE.md` and implementation rule 6 say the shared layer must use `supportsFeature()` rather than branch on server type.

It also fails open in the dangerous direction: the ternary's default arm is the Plex one, so a fourth media server would silently inherit "act on an empty list" - the branch that flags every member as removed by hand.

## Change

Add `MediaServerFeature.TRUSTWORTHY_EMPTY_COLLECTION`, held by Plex only. The .NET BoxSet API can answer with no children for a brief window after a collection is modified, which is the entire reason the workaround exists - so it belongs in the capability matrix beside the other server properties, documented where the other Jellyfin/Emby exclusions are documented.

The sweep condition is also stated the way it actually reads now:

```ts
const shouldCheckRemovals =
  emptyChildListIsTrustworthy || (children && children.length > 0);
```

**Behaviour is unchanged for all three supported servers.** The existing Jellyfin and Plex tests pass untouched, because the spec's `supportsFeature` mock already delegates to the real matrix - so they now exercise the flag.

One test added, asserting the decision follows the capability rather than the configured type (a Plex service with the flag forced off must not trust an empty read). Verified failing against the unfixed code.

### DIFF
--- a/apps/server/src/modules/rules/tasks/rule-executor.service.spec.ts
+++ b/apps/server/src/modules/rules/tasks/rule-executor.service.spec.ts
@@ -279,6 +279,40 @@ describe('RuleExecutorService', () => {
     );
   });
 
+  // The decision follows the capability, not the configured server. Branching on
+  // the type in the shared layer is what the architecture guardrails forbid, and
+  // it would silently hand a fourth media server Plex's answer.
+  it('trusts an empty child list only when the server declares the capability', async () => {
+    const { service, mediaServer, collectionService } = createService(
+      MediaServerType.PLEX,
+    );
+    mediaServer.getCollectionChildren.mockResolvedValue([]);
+    collectionService.getCollectionMedia.mockResolvedValue([]);
+    mediaServer.supportsFeature.mockImplementation(
+      (feature: MediaServerFeature) =>
+        feature !== MediaServerFeature.TRUSTWORTHY_EMPTY_COLLECTION,
+    );
+
+    await (
+      service as unknown as {
+        syncManualMediaServerToCollectionDB: (
+          ruleGroup: { id: number; collectionId: number },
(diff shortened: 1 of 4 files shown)


enoch85:

The capability is defined in the shared contracts map, exposed through the media-server abstraction, and used by the rule executor without reintroducing server-type branching. Plex is the only advertised implementation, which matches the guard being replaced.

Verification included the focused rule-executor suite (51/51), latest green CI, and disposable real-server collection cycles on PMS 1.43.3, Jellyfin 10.11.11, and Emby 4.9.5.0. The Jellyfin and Emby collections were created, populated, emptied, read again, and removed successfully.

enoch85:
Re-review complete at final merged head `3ff67f7`: no findings.

The merged capability remains defined centrally, advertised only by Plex, and consumed through the media-server abstraction without type branching. I repeated real empty/populated/empty collection cycles on Plex, Jellyfin 10, Jellyfin 12, and Emby; all temporary collections were removed and verified absent afterward. Current CI is green.

PR #3643, opened 2026-09-01
fix(ui): say why a bulk action failed instead of one catch-all

enoch85:

## Cause

`postInChunks` caught the transport error and never looked at it:

```ts
} catch {
  for (const remaining of batches.slice(index)) {
    for (const mediaId of remaining) {
      results.push({ mediaId, code: 0, message: t`request error` })
    }
  }
  break
}
```

So a read timeout, a 409 lock conflict (the rules controller documents one: "A collection or rule run held the execution lock for too long"), a 500 and a dropped connection all reach the toast identically.

The shared `getApiErrorMessage` helper is not correct for this path. It ends in `normalizeConnectionErrorMessage`, whose vocabulary is written for media-server and *arr connection tests, and rewrites timeout-shaped failures to "Connection timed out after 5 seconds." This browser-to-Maintainerr request sets no Axios timeout, so that would invent a duration. The local classifier instead preserves the backend message, then the HTTP status, and falls back to `request error` when the response supplies neither.

The old code also stamped `batches.slice(index)` - the failed batch **and every batch after it** - with the same message, though the later ones were never sent.

## Change

- classify the failed chunk locally: use the backend's message when present, otherwise report its HTTP status, and keep `request error` as the fallback when no response detail exists
- report the batches that were never sent as `not attempted`, so a retry is not chasing a fault those items never hit

None of the local messages carries a `Failed - ` prefix: that prefix is the server's and the toast strips it by matching the English text, so a translated one would survive the strip and read twice (the existing comment, still honoured).

i18n: the file already uses the non-reactive `t` from `@lingui/core/macro`, which is correct inside an async callback per `i18n.instructions.md`. Catalogs re-extracted for the two new strings; `yarn i18n:check` passes.

Two new tests plus expanded chunk-failure coverage, all verified against the unfixed code. Full UI suite green (74 files, 407 tests).

Not in this PR: `handleBulkOutcome` is duplicated across `CollectionMediaPage`, `Exclusions` and `Overview` with byte-identical heads and tails, and has already diverged three ways. That is a refactor rather than this bug, so it is worth its own change.

### DIFF
--- a/apps/ui/src/api/bulkMediaAction.spec.ts
+++ b/apps/ui/src/api/bulkMediaAction.spec.ts
@@ -59,11 +59,61 @@ describe('postBulkExclusions', () => {
     ).toBe(true)
     expect(
       response.results
-        .slice(25)
+        .slice(25, 50)
         .every(
           (result) => result.code === 0 && result.message === 'request error',
         ),
     ).toBe(true)
+    // The last chunk was never sent, so it did not fail for that reason.
+    expect(
+      response.results
+        .slice(50)
+        .every(
+          (result) => result.code === 0 && result.message === 'not attempted',
+        ),
+    ).toBe(true)
+  })
+
+  it('reports the reason the server gave instead of a single catch-all', async () => {
+    // A lock conflict and a dropped connection both used to read "request
+    // error". The server's own message is the honest answer for the one that
+    // has it.
+    postMock.mockRejectedValueOnce(
+      Object.assign(new Error('Request failed with status code 409'), {
+        isAxiosError: true,
+        toJSON: () => ({}),
+        response: {
+          status: 409,
+          data: {
+            message:
+              'A collection or rule run held the execution lock too long',
+          },
+        },
+      }),
+    )
+
+    const response = await postBulkExclusions({ mediaIds: ids(25), action: 0 })
+
+    expect(response.results).toHaveLength(25)
+    expect(response.results[0].code).toBe(0)
+    expect(response.results[0].message).toBe(
+      'A collection or rule run held the execution lock too long',
+    )
+  })
+
+  it('does not invent a timeout budget the request never had', async () => {
+    // The connection-test vocabulary says "timed out after 5 seconds"; this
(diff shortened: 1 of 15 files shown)


enoch85:
The behavior itself passed review: on the exact PR head, Playwright forced the shared bulk modal request to return HTTP 409 with `Collection is locked`; the toast reported that reason and the failed item stayed selected. The focused API suite also passed 6/6.

There is one current merge blocker: the required **Formatting** check is failing. Its log identifies both changed files:

- `apps/ui/src/api/bulkMediaAction.ts`
- `apps/ui/src/api/bulkMediaAction.spec.ts`

Please run the repository formatter on those files and push the result. I found no additional functional defect once that required check is restored.

enoch85:
Re-review at `bf6c2f3`: the prior formatting blocker is resolved. Prettier passes on both changed source files, the focused suite passes 6/6, and exact-head Playwright again showed the server-provided 409 reason while keeping the failed item selected.

One non-blocking cleanup remains under the trim-comments rule: `bulkMediaAction.ts` has two consecutive JSDoc blocks before `failureReason`. The first describes `postInChunks`, but no longer sits above that function, so it is now an orphaned/misplaced comment. The catch comment also says "the shared classifier" even though this PR deliberately uses the local `failureReason` helper instead. Please move the first block back above `postInChunks` and change that wording to "the helper" or remove it. No runtime defect found.

enoch85:
Re-review result at d7d86d0: no current-code defects found. The prior JSDoc placement and classifier wording cleanup is resolved. Exact-head verification passed: 6/6 focused API tests, UI type-checking, focused lint, formatting, translation-catalog validation, and diff checks. Playwright forced the real shared-modal request to return HTTP 409 with Collection is locked; the toast surfaced that reason and retained the failed selection. The real success path also completed against the backend/Plex stack. All current required CI checks are green.

Resolved at 9fab0f39: the PR description now accurately documents why failureReason is local, its message/status/fallback order, the two new catalog strings, and the expanded regression coverage. No code change was required.

PR #3644, opened 2026-09-01
fix(collections): drop rule-removal markers when a collection stops syncing

enoch85:
Found while auditing the collection-membership code behind #3636. Branches off `development`.

## Cause

`stopMediaServerSync` tears down the media server collection and nulls the link, but left the collection's `collection_media_rule_removal` markers behind.

`reconcileRuleRemovedOrphans` returns early on a missing `mediaServerId`, so from that point nothing could ever reconcile or clear them, and the FK cascade only fires when the collection row itself is deleted.

The shared-leave path (`leaveSharedMediaServerCollection`) already clears its markers. Only the non-shared unlink path did not.

## Why it is not just dead rows

A marker means "the media server may still be holding this", and the teardown just established that it is not.

If the user later turns the keep-in-Maintainerr-only opt-in back off, a new media server collection is created and the stale markers are reconciled against **that** one. An item that is present there but is not a member - one the user added by hand - matches `present && !memberOrSibling` and is read as a lingering orphan, so the self-heal takes it back out of the user's collection.

## Change

Wipe the collection's markers after a successful teardown - the same wipe `deactivateCollection` already does after its own, with the same reasoning (the FK cascade cannot fire while the collection row survives).

Kept when the teardown **fails**: the link is kept so it retries, and the markers are the record of what still has to come out of that collection.

Two tests, the first verified failing against the unfixed code.

## Overlap with #3636

#3636 adds a `direction` column to this table and changes `reconcileRuleRemovedOrphans`. It does not touch `stopMediaServerSync`, so the two should not conflict textually - but they are the two PRs closest together, so whichever merges second is the one to re-run the collections suite on.

### DIFF
--- a/apps/server/src/modules/collections/collections.service.spec.ts
+++ b/apps/server/src/modules/collections/collections.service.spec.ts
@@ -71,6 +71,13 @@ describe('CollectionsService', () => {
     ruleRemovalRepo = unitRef.get(
       getRepositoryToken(CollectionMediaRuleRemoval) as string,
     );
+    // stopMediaServerSync unlinks inside a transaction. Run the callback and
+    // hand back the same repository mocks, so these cases still assert on the
+    // writes; the rollback itself needs a real database and is covered in
+    // stop-media-server-sync.integration.spec.ts.
+    dataSource.transaction.mockImplementation((runInTransaction: any) =>
+      runInTransaction({ withRepository: (repo: unknown) => repo }),
+    );
     metadataService = unitRef.get(MetadataService);
     mediaItemEnrichmentService = unitRef.get(MediaItemEnrichmentService);
     mediaItemEnrichmentService.enrichItems.mockImplementation(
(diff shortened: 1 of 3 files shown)


enoch85:
**[P2] Make unlinking and marker deletion atomic**

The new sequence saves `mediaServerId: null` first and then deletes the rule-removal markers (`collections.service.ts:4142-4162`). If that delete rejects after the save succeeds, the method reports failure but the database is already unlinked and the markers remain. Every later call returns at line 4133 because there is no `mediaServerId`, so the stranded-marker defect targeted by this PR remains.

The added tests cover teardown failure and save failure, but not this inverse partial failure. A regression with a successful `collectionRepo.save` and rejected `ruleRemovalRepo.delete` exposes it directly.

Please put the local unlink and marker deletion in one database transaction so either both commit or neither does. The focused collections service spec passes 172/172; this atomicity gap is outside its current cases.

enoch85:
Re-review complete at `972c0cb`: the prior P2 is resolved and I found no new findings.

The local unlink and marker deletion now run in one TypeORM transaction, and the update event is emitted only after commit. The new real-SQLite test proves that a marker-delete failure rolls the link update back, covering the partial-failure case the original version missed.

Exact-head focused verification passed 175/175 tests across the service and rollback integration specs. Current CI is green.

PR #3654, opened 2026-09-01
refactor(ui): report a bulk outcome from the outcome itself

enoch85:
`reportBulkOutcome` took counts, so every call site restated fields off the `MediaActionOutcome` it already held. It now takes the outcome.

| | Before | After |
|---|---|---|
| Signature | `{ action, collectionId, succeeded: number, failed: number, failureReasons }` | `MediaActionOutcome` |
| Call sites | 7 lines each, 3 sites | `reportBulkOutcome(outcome)` |
| Net | | -40 lines |

Overview passed a de-duplicated `succeededIds.size` where the other two passed `succeededIds.length`. Not reachable today (one server result per selected id, and the input is a `Set`), so this is not a bug fix, but it is the drift that restating invites.

No behaviour change. Toast wording, the `Failed - ` prefix strip, and the failed-items-stay-selected contract are untouched.

## Verification

| Check | Result |
|---|---|
| `CollectionMediaPage` / `Overview` / `Exclusions` specs | 29/29, unchanged before and after |
| Full UI suite, both shards, `--force` | 74 files, 407 tests, 0 cached |
| `check-types`, `lint`, `format:check` | pass |
| `i18n:check` | catalog current |

Real stack, live Plex 1.43.3 through the UI:

| Path | Toast |
|---|---|
| Add 2 items, real write | `2 items added.` (success variant, selection closed) |
| 3 items forced to fail with a reason | `0 items added. 3 items could not be added (Collection is locked); the failed items stay selected.` (error variant, selection kept open) |

The success case was confirmed server-side: collection membership and the Plex collection both went 3 to 4, and the added item was removed again afterwards.

### DIFF
--- a/apps/ui/src/components/Collection/CollectionDetail/Exclusions/index.tsx
+++ b/apps/ui/src/components/Collection/CollectionDetail/Exclusions/index.tsx
@@ -121,13 +121,8 @@ const CollectionExcludions = (props: ICollectionExclusions) => {
     })
   }
 
-  const handleBulkOutcome = ({
-    action,
-    collectionId,
-    succeededIds,
-    failedIds,
-    failureReasons,
-  }: MediaActionOutcome) => {
+  const handleBulkOutcome = (outcome: MediaActionOutcome) => {
+    const { action, collectionId, succeededIds, failedIds } = outcome
     applyBulkOutcome(new Set(failedIds))
 
     for (const mediaServerId of succeededIds) {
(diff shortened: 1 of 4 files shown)


enoch85:
Re-review result at 8b70e509: no current-code defects found. I traced MediaActionOutcome from the shared modal through all three consumers and through the server contract: selected IDs are unique, each endpoint returns one result per unique ID, and chunk failures also preserve one result per selected ID, so the new array-length counts match the API invariant. The refactor also keeps the shared modal and reporter as the single source of outcome semantics established by #3411 and #3453.

Exact-head verification passed: 29/29 affected component tests, UI type-checking, focused lint, formatting, and diff checks. Playwright against the real backend/Plex stack showed the real success toast (1 item removed), and the item was restored and read back from both SQLite and Plex. A forced per-item refusal reported: 0 items removed. 1 item could not be removed (Collection is locked); the failed items stay selected. The item remained selected. All current required CI checks are green. Ready from the review side.

PR #3676, opened 2026-09-05
fix(sonarr): read the episode list fresh before unmonitoring or deleting files

enoch85:
Reported on Discord: "Unmonitor and delete season" left the Sonarr season unmonitored while its episodes stayed monitored. That part does not reproduce. Sonarr's `SeriesService.UpdateSeries` propagates a changed season flag to that season's episodes on the series PUT in v2, v3 and v4, and both a raw replay and the real collection handler against dev Sonarr 4.0.19 unmonitored the season and all of its episodes. What did reproduce, in the same function, is a stale read in front of the file deletes.

## Cause

`unmonitorSeasons` and `UnmonitorDeleteEpisodes` took the episode list from the cached `get`, the entry rule evaluation fills and keeps for 20 minutes. The delete loop keys off that list's `episodeFileId`s.

| List state at action time | Before |
|---|---|
| file added since the cached read | skipped; `Removed season N` logged; item left the collection |
| file replaced since the cached read | old id deleted, 404, action fails until the entry expires |
| listing failed | season PUT already sent, then the delete loop threw |

## Change

| Path | Change |
|---|---|
| `getEpisodes` | `{ fresh: true }` option: `getWithoutCache` with the slow-instance timeout, the shape of `getRootFolders({ fresh })`; the rule getters keep the cached read |
| `unmonitorSeasons`, `UnmonitorDeleteEpisodes` | read fresh; a failed listing returns before any PUT or delete |
| `existing` scope | unmonitors each episode from that one read instead of a per-episode `UnmonitorDeleteEpisodes` round trip |
| season success log | carries the deleted file count, or says there were none, as the Radarr log does |
| Radarr | already read `moviefile` uncached before deleting; a regression test pins it |

Sportarr reads uncached before every delete, the Plex, Jellyfin and Emby deletes take the item id, and Seerr and the download client read live, so nothing changes there.

## Verified

Dev Sonarr 4.0.19 through the real handler (`POST /api/collections/media/handle`, Plex season to Sonarr series), files hardlink-imported so the fixtures survive:

| Sequence | Before | After |
|---|---|---|
| season action with no files, import 2 files, re-run within the TTL | 0 deletes, Sonarr log silent, `Removed season 1` logged, item removed | 2 deletes, both in Sonarr's `MediaFileDeletionService` log |
| season and episode monitored state after the action | season and all 13 episodes unmonitored | same |

### DIFF
--- a/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.spec.ts
+++ b/apps/server/src/modules/api/servarr-api/helpers/radarr.helper.spec.ts
@@ -266,6 +266,31 @@ describe('RadarrApi', () => {
       expect(getWithoutCacheSpy).toHaveBeenCalledTimes(1);
     });
 
+    it('deletes current file ids without using the rule evaluation cache', async () => {
+      const cached = jest
+        .spyOn(api, 'get')
+        .mockResolvedValue([createRadarrMovieFile({ id: 800 })]);
+      const fresh = jest
+        .spyOn(api, 'getWithoutCache')
+        .mockResolvedValueOnce(movie)
+        .mockResolvedValueOnce([createRadarrMovieFile({ id: 900 })]);
+      jest.spyOn(api as any, 'runPut').mockResolvedValue(true);
+      const runDeleteSpy = jest
+        .spyOn(api as any, 'runDelete')
+        .mockResolvedValue(true);
+
+      await expect(
+        api.updateMovie(5, { monitored: false, deleteFiles: true }),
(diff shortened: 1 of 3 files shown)


enoch85:
@whitestrake Astra ;)

whitestrake:
@enoch85 Astra found no blockers at all and recommended a merge.

> Three independent passes covered correctness, caller integration, failure handling, test coverage, and simplification. The fresh reads preserve rule caching and the unmonitor-before-delete safeguards.

Nice!

PR #3693, opened 2026-09-08
feat: add Transmission download client cleanup

MartyPortatoes:
### Description & Design

Maintainerr can currently clean up qBittorrent jobs after Radarr or Sonarr media removal, but installations using Transmission have no equivalent integration. This leaves completed jobs in the Arr queue when their downloads can no longer be imported.

This change adds Transmission as a second download-client type while preserving qBittorrent as the default for existing installations. It:

- adds settings for the full Transmission RPC endpoint and optional credentials;
- handles Transmission's 409 session-id handshake and Basic authentication through the existing retried HTTP client;
- looks up torrents by exact info hash and removes only matching jobs;
- waits for Transmission's ratio or idle seeding goal, with the existing fallback ratio when no client goal is active;
- retains downloaded data by default and protects shared cross-seed paths from deletion;
- adds the generated settings migration, UI selection and fields, translations, and focused tests.

The RPC URL is intentionally stored as a complete endpoint, normally `http://host:9091/transmission/rpc`, because deployments commonly vary in proxy path and port. Cleanup remains event-driven: it runs when Maintainerr processes the existing media-removal action. This PR does not add a background sweep for jobs left by removals that happened before Transmission was configured.

### Related issue

No existing Transmission issue or pull request was found in this repository.

### AI-Assisted Development

Codex was used to research the existing download-client abstraction, implement the Transmission integration, generate and validate the TypeORM migration, add tests, and run the repository checks. The resulting diff was reviewed against the qBittorrent behavior, existing HTTP retry conventions, settings flow, migration workflow, and Transmission RPC semantics. This remains a draft so the account owner can complete their own final review before marking it ready.

### Checklist

- [ ] I understand the code I am submitting and can explain how it works (owner review pending while this PR is a draft)

### How to test

1. Run `yarn install --immutable`, `yarn lint`, `yarn format:check`, `yarn check-types`, `yarn test`, `yarn i18n:validate`, and `yarn build` on Node 26.
2. In Settings > Download Client, select Transmission, enter the full RPC URL and optional credentials, save, and use the connection test.
3. Exercise cleanup with exact-hash fixtures for unfinished seeding, finished seeding, no configured goal, missing torrents, authenticated RPC, and a shared content path; verify the tests retain data unless deletion is explicitly allowed.

### Additional context

Validated locally on Node 26.8.1. All 3,480 existing and new tests passed, along with lint, formatting, type-checking, locale validation, and the production build. The branch is rebased on the current `development` head.


enoch85:
Thanks for the PR, but if you just vibe coded and can't even explain it, then I won't accept it. 

Vibe-engineered is different story though. I have no issues with using AI as a _tool_, but not as a total replacement.

enoch85:
Took this over: rebased onto `development`, folded into one commit with you as co-author, and changed the following.

| Change | Why |
| --- | --- |
| Transmission 401/403/421 mapped in the helper, naming `rpc-whitelist` and `rpc-host-whitelist` | A 403 surfaced as "Invalid API key" and a 421 as a bare status line. Transmission has no API key, and both whitelists are on by default in the daemon |
| qBittorrent's 403 text moved into `QbittorrentApi`; the service no longer branches on client type | Each client owns its own failure mapping |
| Ids limited to 40-hex, content path via the shared `stripTrailingSlashes`, version trimmed to `4.1.3` | A 64-hex id never resolves in Transmission; less code; no nested parentheses in the alert |
| Stored-settings branch of `testDownloadClient` removed | No route reaches it |
| Redundant `?? qbittorrent` fallbacks and the schema default dropped | The column is `NOT NULL DEFAULT 'qbittorrent'` |
| URL help text for both clients, example URL as a placeholder | Rows no longer shift on switch; example values stay out of translations |
| Specs trimmed | Duplicate CSRF test, factory spec folded, 403 test moved next to the code it covers |

Known limitation kept: Transmission does not persist its idle-finished flag, so a torrent its idle limit stopped reads as unfinished again after a daemon restart. Docs note follows separately.

Verified with the full suite, the real qBittorrent 5.2.3, a spec-modelled Transmission stand-in over real HTTP, the live API and Playwright.

enoch85 (approved):
OK, after my own edits.

PR #3717, opened 2026-09-12
fix(plex): scan library after file deletion

sp71:
## Description & Design

Maintainerr deletes media files through Sonarr, Radarr, or Plex, but Plex can retain the deleted item until its next library scan. This leaves orphaned shows visible in Plex clients.

The change adds a Plex library-scan operation and invokes it once after a successful batch of delete-style actions. Unmonitor-only and quality-profile actions do not scan. Scan failures are logged without retrying a completed deletion.

## Related request

https://features.maintainerr.info/posts/106/update-library-command-on-delete-similar-to-radarr

## AI-Assisted Development

AI assistance was used to inspect the codebase, trace the deletion flow, draft the implementation and tests, and review the diff. I reviewed the design and changes, kept the scan at the collection-handling boundary, and confirmed the Plex endpoint against a live server.

## Checklist

- [x] I have read the CONTRIBUTING.md document.

## How to test

1. Configure a Plex server and a delete-style Maintainerr collection action.
2. Handle a disposable media item through the normal worker or manual collection endpoint.
3. Confirm Maintainerr starts a scan for the affected Plex library and the deleted item disappears after the scan.
4. Confirm unmonitor-only actions do not start a scan.

### DIFF
--- a/apps/server/src/modules/api/media-server/media-server.interface.ts
+++ b/apps/server/src/modules/api/media-server/media-server.interface.ts
@@ -449,6 +449,12 @@ export interface IMediaServerService {
   deleteFromDisk(itemId: string): Promise<void>;
 
   /**
+   * Ask the media server to scan a library for filesystem changes.
+   * Currently implemented by Plex only.
+   */
+  scanLibrary?(libraryId: string): Promise<void>;
+
+  /**
    * Get all media server IDs for a context action (add/remove from collection).
    * Handles show→season→episode traversal based on collection type.
    *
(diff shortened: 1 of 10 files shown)


enoch85:
We already have left over cleanup. Did you try that?

https://docs.maintainerr.info/collections/#leftover-folder-cleanup

enoch85:
Hmm, had a look at this again and I was wrong, we don't cover it. 

I'm not sure this is what we want though. Re-scanning the media server can be a heavy operation - and should IMHO be set on a schedule on the media server itself.

I'm leaning towards a documentation change, but will experiment with some other options as well.

enoch85:
This scans too much, and only covers plex. 

Closing.

PR #3720, opened 2026-09-14
feat(collections): rescan the item's folder after an *arr delete

enoch85:
After Radarr, Sonarr or Sportarr deletes an item's files, Maintainerr now asks the media server to rescan just that item's folder, so the item drops out of Plex, Jellyfin or Emby without waiting for the next library scan ([feature request](https://features.maintainerr.info/posts/106/update-library-command-on-delete-similar-to-radarr)). Supersedes #3717 with a targeted scan on all three servers.

| Media server | Request | Server behaviour |
| --- | --- | --- |
| Plex | `GET /library/sections/{id}/refresh?path=<folder>` | partial scan of that folder; the item is trashed and removed with "Empty trash automatically after every scan" (default on) |
| Jellyfin | `POST /Library/Media/Updated` with the folder | re-validates the nearest existing folder after its library monitor delay (60 s default) |
| Emby | `POST /Library/Media/Updated` with the folder | same, after its monitor delay (90 s default) |

### Where the folder comes from

The media server's own item metadata (`MediaItem.path`: the file for movies and episodes, the folder for shows and seasons; a Plex season hops to its show), read before the action. No media mount in the Maintainerr container and no path mapping are needed.

### When nothing is sent

Unmonitor and quality-profile actions, and deletes the media server performed itself. A failed rescan request logs one warning; the delete stands.

Verified on the dev stack: real Radarr deletes through Maintainerr reached Plex (request with the exact folder in its server log), Emby (204 logged) and Jellyfin; the servers could not drop the items there because none of the dev containers can read the media volume at the moment.


enoch85:
@sp71 Please test that it behaves as expected by using the `development` tag.

Please don't change the tag back until next release to avoid any inconveniences.

sp71:
@enoch85  confirmed, works well. Thank you!

PR #3764, opened 2026-09-22
feat: support Jellyfin backdrop overlays

parac:
## Summary

Adds configurable backdrop overlays for Jellyfin, including the landscape artwork variants used by Infuse and Apple TV clients.

## Changes

- Add `backdrop` overlay templates and collection configuration.
- Synchronize Jellyfin Backdrop, Thumb, and Banner artwork for landscape clients.
- Fall back from Backdrop to Thumb/Primary when source artwork is missing.
- Track poster and backdrop processing state independently.
- Back up each original landscape artwork type separately and restore it exactly on revert.
- Keep titles and metadata intact while overlays are reverted or reset.

## Validation

- Server TypeScript check passes.
- Overlay processor tests: 39 passed.
- Jellyfin overlay provider tests: 9 passed.
- Tested with Jellyfin and Infuse artwork paths.


enoch85:
Thanks for the PR!

How is this different from the overlays we currently support?

Do you have screenshots?

parac:
> Thanks for the PR!
> 
> How is this different from the overlays we currently support?
> 
> Do you have screenshots?

We mainly use Infuse Pro (instead of Jellyfin directly) and Infuse uses different pictures as source for their "backdrop".

Screenshots coming up.

enoch85:
> Infuse Pro

No need to provide screenshots.

I will actually decline this. The better PR would be to support Infuse Pro in that case. Reason? I don't like stuff that can't be used by all media servers, and this is super specific and can only be utilized by a few people + adds code we need to maintain just for that specific use case. So, better to support Infuse instead _in that case_.

enoch85:
One correction to my earlier comment: Infuse is a client, not a server, so there is nothing to integrate with there either.

parac:
> I will actually decline this. The better PR would be to support Infuse Pro in that case. Reason? I don't like stuff that can't be used by all media servers, and this is super specific and can only be utilized by a few people + adds code we need to maintain just for that specific use case. So, better to support Infuse instead _in that case_.

I would argue against this - but slightly. I would assume, that currently, and also in the future, potential clients and servers would use the additional pictures for the overlay. Currently, only one picture is taken as a base for the overlay - this PR extends it to all available artefacts.

enoch85:
Fair point on where clients draw artwork from, and I don't doubt the use case. Two things though: as submitted this only implements the backdrop target for Jellyfin (Plex and Emby would not get it), and more importantly this is a scope decision rather than a technical one. The overlay feature is going to stay poster and title card only; I'm not adding further artwork targets on any server, because each one is another slot to back up, keep in sync and restore, and the maintenance cost outweighs the number of people it reaches. Thanks for the work regardless.

